PR #1, opened 2021-11-29
Loosen lowerbound python requirements

NatPRoach:
Loosening python requirements from `>=py3.6.10` -> `>=3.6.0`, which should also be supported.

Still passes tests on 3.6.0 based on testing locally, and it gives more flexibility for use in other projects.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,4 +1,12 @@
 [[package]]
+name = "appdirs"
+version = "1.4.4"
+description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
+category = "dev"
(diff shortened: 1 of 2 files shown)


nh13 (changes requested):
Please updated github actions to test both `3.6.0` and `3.6.10`: https://github.com/fulcrumgenomics/fgpyo/blob/9e78e1b4feeeebcc631102e0eec8c19ac154517a/.github/workflows/pythonpackage.yml#L12.  I think it is testing the "latest" version of `3.6`, so we don't know if this works via CI

nh13 (approved):
LGTM once tests pass

PR #2, opened 2021-12-19
Add docs for readthedocs

nh13:
For a live version of this branch, see: https://fgpyo.readthedocs.io/en/readthedocs/index.html

Can update the site in the fgypo "about" homepage.


docs/Makefile
```diff
@@ -0,0 +1,88 @@
+# Makefile for Sphinx documentation
```
nh13, line 1: this is very similar to the makefile from `pysam` and `snakeparse` and others

docs/api.rst
```diff
@@ -0,0 +1,34 @@
+=====
+FGYPO
```
NatPRoach, line 2:
```suggestion
FGPYO
```

NatPRoach (approved):
LGTM, just one small typo we should probably fix before merging.

PR #4, opened 2022-02-04
Adding custom error type such that errors can be handled better

NatPRoach:
More robust error handling that should allow for catching syntax errors etc that were previously being suppressed.


nh13 (approved):
Thank-you for improving this!

PR #14, opened 2022-11-18
A bunch of re-organization followed by adding a bunch of code

tfenne:
The commits are stepwise, and it's probably easiest to review commit by commit.


README.md
```diff
@@ -53,12 +53,6 @@ conda activate fgpyo
 poetry install
 ```
 
-If, during `poetry install` on Mac OS X errors are encountered running gcc/clang to build `pybedtools` or other packages with native code, try setting the following and re-running `poetry install`:
-```bash
-export CFLAGS="-stdlib=libc++"
-``` 
-
-
```
nh13, line 61: why is this hint no longer needed?

fgpyo/collections/__init__.py
```diff
@@ -0,0 +1,84 @@
+from typing import Any
+from typing import Callable
+from typing import Generic
+from typing import Iterable
+from typing import Iterator
+from typing import List
+from typing import TypeVar
+from typing import Union
+
+
```
nh13, line 10: 1. I think we should give attribution to where this came from
2. I think the long docs with examples are super useful to new folks on our team to show how to use this like you did for the other modules.  Can you copy those over too?

fgpyo/collections/tests/test_peekable_iterator.py
```diff
@@ -0,0 +1,53 @@
+"""Tests for :py:mod:`~samwell.itertools`"""
```
nh13, line 1:
```suggestion
"""Tests for :py:mod:`~fgpyo.collections.PeekableIterator`"""
```

fgpyo/sam/__init__.py
```diff
@@ -0,0 +1,586 @@
...
+    """
+
+    elements: Tuple[CigarElement, ...] = attr.ib(default=())
+
+    @classmethod
+    def from_cigartuples(cls, cigartuples: Optional[List[Tuple[int, int]]]) -> "Cigar":
+        """Returns a Cigar from a list of tuples returned by pysam.
+
+        Each tuple denotes the operation and length.  See
+        :class:`~samwell.sam.CigarOp` for more information on the
```
nh13, line 417:
```suggestion
        :class:`~fgpyo.sam.CigarOp` for more information on the
```

nh13 (approved):
Minor stuff. I do think we should add attribution to where these came from.  Thank-you for doing this!!!

README.md
tfenne: Because pybedtools is no longer a [transitive] dependency!

fgpyo/collections/__init__.py
tfenne: I thought I had copied those module docs - have now.
Re: attribution I added a section to the LICENSE doc.  I think that's sufficient to merge but if you feel like we should add more elsewhere let me know.

PR #15, opened 2022-11-18
New Template class and TemplateIterator.

tfenne:


fgpyo/sam/__init__.py
```diff
@@ -584,3 +586,110 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+                else:
+                    r2_supplementals.append(rec)
+            if rec.is_secondary:
+                if is_r1:
+                    r1_secondaries.append(rec)
+                else:
+                    r2_secondaries.append(rec)
+
+        return Template(
+            name=next(iter(recs)).query_name,
```
nh13, line 652: Wait, how will this work if `recs` is an iterator?  You've exhausted `recs` in the for loop.  Also consider a better error message if `recs` is empty.

fgpyo/sam/__init__.py
```diff
@@ -584,3 +586,110 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+    or insert.
+
+    Attributes:
+        name: the name of the template/query
+        r1: Primary alignment for read 1, or None if there is none
+        r2: Primary alignment for read 2, or None if there is none
+        r1_supplementals: Supplementary alignments for read 1
+        r2_supplementals: Supplementary alignments for read 2
+        r1_secondaries: Secondary (non-primary) alignments for read 1
+        r2_secondaries: Secondary (non-primary) alignments for read 2
```
nh13, line 603: do you want to do any validation on these?  E.g. must have either r1 or r2?  That all reads have the same query name as each other and `name`?  That `r1`, `r2` are primary, that `*_supplementals` are really supplementals, etc.?  You would be fine if you use the `build` method, but I could see bugs occur if someone uses `attr.evolve` and there's not validation.

nh13: `name` could be a property

```python
@property
def name(self) -> str:
   return r1.query_name if r1 is not None else r2.query_name
```

fgpyo/sam/__init__.py
```diff
@@ -584,3 +586,110 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+        )
+
+    def primary_recs(self) -> List[AlignedSegment]:
+        """Returns a list with all the primary records for the template."""
+        recs = []
+        if self.r1 is not None:
+            recs.append(self.r1)
+        if self.r2 is not None:
+            recs.append(self.r2)
+        return recs
```
nh13, line 668:
```suggestion
        return [r in (self.r1, self.r2) if r is not None]
```

fgpyo/sam/__init__.py
nh13: Could you add a test for `build` where an `iter` is passed?

nh13 (changes requested):

fgpyo/sam/__init__.py
```diff
@@ -156,7 +156,7 @@
 import enum
 import io
 from pathlib import Path
-from typing import Any
+from typing import Any, Iterable, Iterator
```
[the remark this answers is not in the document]
tfenne: Is there a way to have flake8 or black enforce this?  I thought it both of those were passing I was good to go.

fgpyo/sam/__init__.py
```diff
@@ -584,3 +586,110 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+        r2_secondaries: Secondary (non-primary) alignments for read 2
+    """
+
+    name: str
+    r1: Optional[AlignedSegment] = None
+    r2: Optional[AlignedSegment] = None
+    r1_supplementals: List[AlignedSegment] = []
+    r2_supplementals: List[AlignedSegment] = []
+    r1_secondaries: List[AlignedSegment] = []
+    r2_secondaries: List[AlignedSegment] = []
```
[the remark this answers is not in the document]
tfenne: Some days I hate Python.  I was hoping it was smarter than this.

fgpyo/sam/__init__.py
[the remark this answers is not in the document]
tfenne: See #16

fgpyo/sam/__init__.py
```diff
@@ -584,3 +586,110 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
     insert_size = isize(r1, r2)
     r1.template_length = insert_size
     r2.template_length = -insert_size
+
+
+@attr.s(auto_attribs=True, frozen=True)
```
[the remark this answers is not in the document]
tfenne: Maybe we do that separately if you think we should?  I don't want to mess with the attrs version in case it screws up `Metric`.

fgpyo/sam/__init__.py
tfenne: I've changed it.  But f' python.  `Iterator` should not implement `Iterable`.  Sigh.

fgpyo/sam/__init__.py
tfenne: I'm 50/50 on the validation.  I don't necessarily want to incurr that overhead on every construction, especially if the TemplateIterator is doing it.  I don't think r1 or r2 has to be present ... the validations would boil down to validating flags on everything that _is_ present, _and_ validating that the query_name is the same across all records.  Thoughts?

`name` could be a property, but it require searching all reads/lists to find a record, so I think it's nicer to compute it when we're iterating all the reads at construction.  Do you disagree now that I've fixed the way it is collected?

fgpyo/sam/__init__.py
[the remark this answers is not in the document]
nh13: Thank-you!

fgpyo/sam/__init__.py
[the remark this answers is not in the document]
nh13: sgtm

fgpyo/sam/__init__.py
nh13: I see the value of not validating upon every construction.  I see two paths forward:

1. have a private attributed `_validated` that be set by the `TemplateIterator` to `true` to skip validation, and if not set, the validation methods are no longer no-ops?
2. make it clear in the docstring that validation does not occur?

I'd lean to (2) for now, but (1) is cute.

nh13 (approved):
One last decision for you on validation (I gave you an out :)) and then ship it!

fgpyo/sam/__init__.py
tfenne: I split the difference and:

1. Implemented a `validate()` function that asserts all kinds of things
2. Made the `build()` function call that, but with a parameter to disable it
3. Added class level docs to encourage use of `Template.build()` over `Template()`

PR #16, opened 2022-11-19
Add isort as a dev dependency and use it to keep imports tidy.

tfenne:
Just for you @nh13 !

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ mypy = ">=0.770"
 flake8 = ">=3.8.1"
 black = ">=19.10b0"
 pytest-cov = ">=2.8.1"
-isort = "^5.10.1"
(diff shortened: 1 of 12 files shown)


nh13 (approved):
To write python, you must first invent the universe.  Ship it!

pyproject.toml
```diff
@@ -41,6 +41,7 @@ mypy = ">=0.770"
 flake8 = ">=3.8.1"
 black = ">=19.10b0"
 pytest-cov = ">=2.8.1"
+isort = "^5.10.1"
```
nh13, line 44: why fix the version, and not make it at least this version?

tfenne: Apparently that's just what happens when I `poetry add isort`.  Fixed and relocked.

PR #17, opened 2022-11-25
Added a function to the sam module to calculate mms, nm, etc.

tfenne:


fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
     r2.template_length = -insert_size
 
 
+@attr.s(auto_attribs=True, frozen=True)
+class ReadEditInfo:
+    """Counts various stats about how a read compares to a reference sequence."""
+
+    mismatches: int
+    insertions: int
```
nh13, line 595: consider stating somewhere the # of insertions are relative to the reference (i.e. extra bases in the read that are not in the reference).  I am going to come back to this in 6mo and not remember and have to look at the code.

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+    inserted_bases: int
+    deletions: int
+    deleted_bases: int
+    nm: int
```
nh13, line 599: Perhaps a longer doc-string would clarify if `nm` is the `NM` SAM tag verbatim or re-computed, or meant to be something else.

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+
+
+def calc_edit_info(
+    r: AlignedSegment, reference_sequence: str, reference_offset: Optional[int] = None
+) -> ReadEditInfo:
```
nh13, line 604: I'd argue for using longer names for parameters and try to avoid abbreviations for function names, where possible.  I think the function name here is less important, but I think `r` needs to be changed at least.


```suggestion
def build_edit_info(
    recrod: AlignedSegment, reference_sequence: str, reference_offset: Optional[int] = None
) -> ReadEditInfo:
```

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+    """
+    Calculates the number of mismatches and the NM tag value for a read.  The read must be aligned.
```
nh13, line 606: Calculates more than mismatches & NM now.

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+        r: the read/record for which to calculate values
+        reference_sequence: the reference sequence (or fragment thereof) that the read is
+          aligned to
+        reference_offset: if provided, assume that reference_sequence[reference_offset] is the
+          first base aligned to in reference_sequence, otherwise use r.reference_start
+
+    Returns:
+        a ReadEditInfo with information about how the read differs from the reference
+    """
+    query_offset = 0
```
nh13, line 618: probably not expensive just to assert it's aligned even though the docs say it must
```suggestion
    assert r.is_mapped, f"Read must be mapped: {r.query_name}"
    query_offset = 0
```

fgpyo/sam/__init__.py
nh13: Consider describing how NM is computed.

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+        elif op == CigarOp.D:
+            deletions += 1
+            del_bases += elem.length
+        elif op == CigarOp.M or op == CigarOp.X or op == CigarOp.EQ:
+            for i in range(0, elem.length):
+                q = r.query_sequence[query_offset + i].upper()
+                t = reference_sequence[target_offset + i].upper()
+                if q != t or q not in ok_bases:
+                    mms += 1
+
```
nh13, line 640: do you want to worry about skip bases, or assert there are none?

fgpyo/sam/__init__.py
```diff
@@ -590,6 +587,70 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+    Returns:
+        a ReadEditInfo with information about how the read differs from the reference
+    """
+    query_offset = 0
+    target_offset = reference_offset if reference_offset is not None else r.reference_start
+    cigar = Cigar.from_cigartuples(r.cigartuples)
+
+    mms, insertions, ins_bases, deletions, del_bases = 0, 0, 0, 0, 0
+    ok_bases = ["A", "C", "G", "T"]
```
nh13, line 623: make into a set, so we don't have to do a linear scan for containment below
```suggestion
    ok_bases = {"A", "C", "G", "T"}
```

nh13 (approved):
A number of small things that I think need updating.  Let me know if you don't agree and then we should discuss, otherwise LGTM.

fgpyo/sam/__init__.py
tfenne: Lol, you've been spending too much time in gene editing space ;)  I'll add docs.

fgpyo/sam/__init__.py
tfenne: Skip as in `N` operators?  I think those are handled by having the query and target offsets incremented outside the if/elif/else.

tfenne:
I think I addressed all your feedback @nh13, will merge once CI passes.

PR #18, opened 2022-11-26
Fixed the str() method for SupplementaryAlignment to safely round-trip.

tfenne:


fgpyo/sam/__init__.py
```diff
@@ -519,7 +519,17 @@ class SupplementaryAlignment:
...
+            str(item)
+            for item in (
+                self.reference_name,
+                self.start + 1,
+                "+" if self.is_forward else "-",
+                self.cigar,
+                self.mapq,
+                self.nm,
+            )
+        )
```
nh13, line 532: it was too cute, too cute.  Thanks for fixing this!

nh13 (approved):

PR #21, opened 2022-11-27
Add ability to set mapq when adding pairs and single reads to SamBuilder

### LINKED ISSUE #19: SamBuilder should allow setting mapqs in add_pair() and add_single()
Default value should be the value used when constructing the SamBuilder but should be overridable when calling `add_*()`.

tfenne:
Fixes #19

### DIFF
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -147,7 +147,12 @@ class SamBuilder:
         return "".join(self._random.choices("ACGT", k=length))  # type: ignore
 
     def _new_rec(
-        self, name: str, chrom: str, start: int, attrs: Optional[Dict[str, Any]]
+        self,
+        name: str,
+        chrom: str,
+        start: int,
+        mapq: Optional[int],
+        attrs: Optional[Dict[str, Any]],
     ) -> AlignedSegment:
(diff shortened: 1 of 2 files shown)


fgpyo/sam/builder.py
```diff
@@ -169,10 +175,11 @@ def _new_rec(
         rec.query_name = name
         rec.reference_name = chrom
         rec.reference_start = start
-        rec.mapping_quality = self.mapping_quality
+        rec.mapping_quality = mapq if mapq is not None else self.mapping_quality
```
nh13, line 178: I've seen you do this pattern a few times (` _ if _ is not None else _`), why not `_ if _ is None else _` which is more concise?
```suggestion
        rec.mapping_quality = self.mapping_quality if mapq is None else mapq
```

nh13 (approved):
One question about style, but LGTM

fgpyo/sam/builder.py
tfenne: I tend to write it that way as that's the way I say it in English ... "give me _that thing" if it's available, else this other thing".  I.e. the thing that takes preference if it's there is first.  It's closer to scala's style `mapq.getOrElse(self.mapping_quality)`

PR #23, opened 2022-12-02
Let the pysam requirement float.

tfenne:
Specifying a precise required version is tough for a library.


nh13 (approved):
Shall we remove support for 3.6 as well?  We'll need to update the github actions in this PR to make it pass.

tfenne:
Probably makes sense to go 3.7+ given 3.6 stopped receiving security updates a year ago!

Issue #20, opened 2022-11-27
SamBuilder should support writing out to a path _without_ sorting.


tfenne:
I need to emit query-grouped BAMs, so I'm doing this outside SamBuilder for now.  But would be nice if the `to_path()` method had a `sort: bool = True` parameter that could be turned off.  The header would then need to be edited to have `SO:unsorted` upon writing.

clintval:
Closed with https://github.com/fulcrumgenomics/fgpyo/pull/25

PR #25, opened 2022-12-27
Adding the option to specify the sort order of an output SAM file produced by SamBuilder.

NatPRoach:
The support for ``TemplateCoordinate`` sorted records is coming in the first wave of wrapped `pysam` `samtools` dispatch calls.

This should address https://github.com/fulcrumgenomics/fgpyo/issues/20.

This is also more or less identical to an addition made to samwell in a PR merged presumably after the sambuilder module code was pulled over from samwell to here.

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -136,6 +136,7 @@ The module contains the following public classes:
     - :class:`~fgpyo.sam.SupplementaryAlignment` -- Stores a supplementary alignment record
         produced by BWA and stored in the SA SAM tag.
     - :class:`~fgpyo.sam.SamFileType` -- Enumeration of valid SAM/BAM/CRAM file types.
+    - :class:`~fgpyo.sam.SamOrder` -- Enumeration of possible SAM/BAM/CRAM sort orders.
     - :class:`~fgpyo.sam.CigarOp` -- Enumeration of operators that can appear in a Cigar string.
     - :class:`~fgpyo.sam.CigarElement` -- Class representing an element in a Cigar string.
     - :class:`~fgpyo.sam.CigarParsingException` -- The exception raised specific to parsing a
@@ -899,3 +900,14 @@ class TemplateIterator(Iterator[Template]):
         name = self._iter.peek().query_name
         recs = self._iter.takewhile(lambda r: r.query_name == name)
         return Template.build(recs, validate=False)
+
+
+class SamOrder(enum.Enum):
+    """
+    Enumerations of possible sort orders for a SAM file.
+    """
+
+    Unsorted = "unsorted"  #: the SAM / BAM / CRAM is unsorted
+    Coordinate = "coordinate"  #: coordinate sorted
+    QueryName = "queryname"  #: queryname sorted
+    Unknown = "unknown"  # Unknown SAM / BAM / CRAM sort order
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -14,6 +14,7 @@ from array import array
 from pathlib import Path
 from random import Random
 from tempfile import NamedTemporaryFile
+from typing import IO
 from typing import Any
 from typing import Callable
 from typing import Dict
@@ -26,6 +27,7 @@ from pysam import AlignedSegment
 from pysam import AlignmentHeader
 
 from fgpyo import sam
+from fgpyo.sam import SamOrder
 
 
 class SamBuilder:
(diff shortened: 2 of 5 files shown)


fgpyo/sam/builder.py
```diff
@@ -116,14 +119,27 @@ def __init__(
             extra_header: a dictionary of extra values to add to the header, None otherwise.  See
                           `::class::~pysam.AlignmentHeader` for more details.
             seed: a seed value for random number/string generation
+            sort_order: optional sort order, if `None` reads will be output in the same order as
+                they were appended. If `SamOrder.Coordinate`, reads will be ordered by reference
+                index and coordinate order. If `SamOrder.QueryName`, reads will be ordered by
+                query name. `SamOrder.TemplateCoordinate` is not currently supported.
```
nh13, line 125: 1. lets remove `TemplateCoordinate` it until it is supported
2. I don't think you need to explain the options here, you have the type, and folks can look that up

fgpyo/sam/builder.py
```diff
@@ -116,14 +119,27 @@ def __init__(
...
         self.r2_len: int = r2_len if r2_len is not None else self.DEFAULT_R2_LENGTH
         self.base_quality: int = base_quality
         self.mapping_quality: int = mapping_quality
+
+        sort_order = SamOrder.Unsorted if sort_order is None else sort_order
+        assert sort_order in [
+            SamOrder.Coordinate,
+            SamOrder.QueryName,
+            SamOrder.Unsorted,
+        ], "`sort_order for `SamBuilder` must be one of `Coordinate` `QueryName` or `Unsorted`"
```
nh13, line 138:
```suggestion
        assert sort_order in [
            SamOrder.Coordinate,
            SamOrder.QueryName,
            SamOrder.Unsorted,
        ], f"`sort_order for `SamBuilder` must be one of `Coordinate` `QueryName` or `Unsorted`.  Found {sort_oder}"
```

nh13 (changes requested):
minor comments

fgpyo/sam/builder.py
```diff
@@ -116,14 +119,24 @@ def __init__(
...
         self.r2_len: int = r2_len if r2_len is not None else self.DEFAULT_R2_LENGTH
         self.base_quality: int = base_quality
         self.mapping_quality: int = mapping_quality
+
+        sort_order = SamOrder.Unsorted if sort_order is None else sort_order
+        assert sort_order in [SamOrder.Coordinate, SamOrder.QueryName, SamOrder.Unsorted], (
+            "`sort_order` for `SamBuilder` must be one of `Coordinate` `QueryName` or `Unsorted`."
+            + f" Found {sort_order}"
+        )
+        self.sort_order: SamOrder = sort_order
```
nh13, line 136: Can you propagate this as private so that folks don't mutate this and expect the header to change?
```suggestion
        self._sort_order: SamOrder = sort_order
```

TedBrookings:
@nh13 In addition to the change you requested, I made three other notable changes

1. I updated the poetry environment to succeed in more recent python environments.
2. I removed the `None` option for `sort_order`. `Unsorted` and `Unknown` are both options and it feels messy to have `None` kicking around in there too.
3. I allowed writing with `Unknown` sort order, since it's in the spec. It's the same as `Unsorted` in its effect. This feels marginal to me, and I think you might feel strongly about removing it, so I wanted to draw attention.

fgpyo/sam/__init__.py
```diff
@@ -899,3 +900,14 @@ def __next__(self) -> Template:
         name = self._iter.peek().query_name
         recs = self._iter.takewhile(lambda r: r.query_name == name)
         return Template.build(recs, validate=False)
+
+
+class SamOrder(enum.Enum):
```
nh13, line 905: Having looked at this many months later, I wonder if this is too simplistic (see our Scala equivalent: https://github.com/fulcrumgenomics/fgbio/blob/ff1ca67772c34a8defeb835d692a53ad6e536f74/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala#L172).  It doesn't allow custom sort orders, including setting group order, sub-sort, etc.  I am fine with this for now, but changing this later will require breaking changes, so lets require @tfenne to weigh in too.

nh13 (approved):

fgpyo/sam/__init__.py
tfenne: I would stop where we are - there's nothing here that would prevent us from doing something more sophisticated later.

Issue #22, opened 2022-11-30
Create methods and type stubs on pysam htslib C extension calls


NatPRoach:
It is often helpful to use `pysam` bindings to htslib functions a la
```
import pysam
pysam.sort(myfile, "-o", my_output)
```
etc.

`mypy` will complain when attempting to use these functions (presumably because they originate from a C extension module) with an error "AttributeError: module pysam has no attribute 'sort'"

I think it would behoove us to make wrapper functions around the pysam htslib calls with named parameters and type checking etc.
This will require generating stubs for the functions, e.g.
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    def sort(*args: Any) -> None:
        pass
else:
    from pysam import sort
```

nh13:
Why not solve it in pysam so everyone can benefit?  https://github.com/pysam-developers/pysam/issues/1153

tfenne:
Well, @nh13 part of the goal of this would be to provide a _better_ interface that has the commonly used parameters to each of `samtools`' command line modules exposed as actual python parameters, so they are easier to use.  E.g. instead of `sort(*args: Any)`, we might want:

```python
def sort(in_bam: Path, out_bam: Path, compress: int = 5, by_name: bool = False, threads: int = 1)
```

nh13:
That makes sense. My only hesitancy is that the interface is tied to a specific samtools version, so we’d have to pin the version and revisit the available options each time we update the samtools dependency. And then why not have it in pysam?

NatPRoach:
@nh13 @tfenne I have a draft PR for the first of the wrappers here, more to come. If we want to get them reviewed here and then use that to make a PR into `pysam` I am also fine with that. Would just need the heads up if that's what we want to do.

https://github.com/fulcrumgenomics/fgpyo/pull/26

PR #26, opened 2022-12-27
Feature/ samtools typehinted wrappers

NatPRoach:
Closes: https://github.com/fulcrumgenomics/fgpyo/issues/22

EDIT: This is the first 4 type hinted wrappers around pysam dispatch functions. ~~A lot more to do but I wanted to start with the functions most important to us.~~(Per conversation with Tim these 4 functions are the only ones we'll be wrapping for now.) ~~Opening it up for review for now, but will be working on another PR shortly with the most important bcftools functions.~~

~~This (at the moment) is the first of the pysam dispatch wrappers for fgpyo. More to do here and I want to do a batch of the ones most important to us so this is a draft for now.~~

### DIFF
--- a/fgpyo/samtools.py
+++ b/fgpyo/samtools.py
@@ -11,9 +11,9 @@ Examples
    >>> from fgpyo import samtools
    >>> Path("example1.sorted.bam.bai").exists()
    False
-   >>> samtools.sort(input=Path("example1.bam"), output=Path("example1.sorted.bam"))
+   >>> samtools.sort(alignment_file=Path("example1.bam"), output=Path("example1.sorted.bam"))
    ''
-   >>> samtools.index(input=Path("example1.sorted.bam"))
+   >>> samtools.index(alignment_file=Path("example1.sorted.bam"))
    ''
    >>> Path("example1.sorted.bam.bai").exists()
    True
@@ -35,7 +35,9 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 from typing import List
 from typing import Optional
+from typing import Sequence
 from typing import Tuple
+from typing import Union
 
 from fgpyo.sam import SamFileType
 from fgpyo.sam import SamOrder
@@ -61,7 +63,7 @@ else:
 
 
 def dict(
-    input: Path,
+    fasta_file: Path,
     output: Path,
     no_header: bool = False,
     alias: bool = False,
@@ -72,6 +74,7 @@ def dict(
 ) -> str:
     """
     Calls the `samtools` function `dict` using the pysam samtools dispatcher.
+    Create a dictionary file from an input FASTA file.
 
     Arguments will be formatted into the appropriate command-line call which will be invoked
     using the pysam dispatcher.
@@ -79,7 +82,7 @@ def dict(
     Returns the stdout of the resulting call to pysam.dict()
 
     Args:
-        input: Path to the FASTA files to generate a sequence dictionary for.
+        fasta_file: Path to the FASTA files to generate a sequence dictionary for.
         output: Path to the file to write output to.
         no_header: If true will not print the @HD header line.
         assembly: The assembly for the AS tag.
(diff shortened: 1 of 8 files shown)


fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
...
+    if compression_level is not None:
+        args.extend(["-I", str(compression_level)])
+
+    if sort_order == SamOrder.QueryName:
+        args.append("-n")
+    elif sort_order == SamOrder.TemplateCoordinate:
+        args.append("--template-coordinate")
+    else:
+        assert (
+            sort_order == SamOrder.Coordinate
```
clintval, line 361: I think the check should be for "not unknown" and "not unsorted" in case other orders are added.

fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
...
+            complementing where appropriate. This has the effect of collating some similar data
+            together, improving the compressibility of the unmapped sequence. The minimiser kmer
+            size is adjusted using the ``kmer_size`` option. Note data compressed in this manner
+            may need to be name collated prior to conversion back to fastq.
+
+            Mapped sequences are sorted by chromosome and position.
+        kmer_size: the kmer-size to be used if sorting unmapped reads.
+        compression_level: The compression level to be used in the final output file.
+        memory_per_thread: Approximately the maximum required memory per thread, specified either
+            in bytes or with a K, M, or G suffix.
```
clintval, line 317: Suffixes include `M` as valid yet the CLI signature default has `MB`?

fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+def sort(
+    alignment_file: Path,
+    output: Path,
+    index_output: bool = True,
+    sort_unmapped_reads: bool = False,
+    kmer_size: int = 20,
+    compression_level: Optional[int] = None,
+    memory_per_thread: str = "768MB",
```
clintval, line 287: Where did this default come from?

fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
...
+    threads: int = 1,
+    no_pg: bool = False,
+) -> str:
+    """
+    Calls the `samtools` function sort using the pysam samtools dispatcher.
+
+    Arguments will be formatted into the appropriate command-line call which will be invoked
+    using the pysam dispatcher.
+
+    Returns the stdout of the resulting call to pysam.sort()
```
clintval, line 301: For Google style docstrings, use the `Returns:` block like you would an `Args:` or `Attributes`.

Only making this comment here, but it applies to all other docstrings too.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
...
+
+    args = [] if len(alignment_file) == 1 else ["-M"]
+
+    if index_type != SamIndexType.BAI:
+        args.append(index_type.value)
+
+    if index_type == SamIndexType.CSI:
+        args.extend(["-m", str(csi_index_size)])
+    args.extend(["-@", str(threads)])
+    args.extend(f"{input_path}" for input_path in alignment_file)
```
clintval, line 273: The formatting of when you add a newline whitespace between code vs choosing not to is changing throughout the PR. I'd recommend keeping the style consistent.

fgpyo/samtools.py
```diff
@@ -0,0 +1,375 @@
+"""
```
clintval, line 1: Does this file automatically show up in generated docs?

I know we'll eventually need to fixup the VCF work so it is indexed: https://github.com/fulcrumgenomics/fgpyo/issues/72

clintval (changes requested):
Thank you for reviving this PR @TedBrookings!

I haven't looked at the tests yet, but I do see many new code paths/branches which are untested. Could you add additional tests where it makes sense to ensure we have coverage over important codepaths?

Most of my comments are about the API and documentation.

Could you also update the PR title to be more descriptive so it will be easier to understand what this commit does once squashed to main?

I'll do another end-to-end review (including tests) after you've taken another pass at the PR!

Issue #9, opened 2022-11-12
Metric should handle Optional fields better


tfenne:
Formatting of None should result in "", and formatting of Some(x) should result in formatting x.
And on the parsing side, need to set None if the value is empty, else parse the inner value.

nh13:
@tfenne should we instead use the same "empty" token like we do in fgbio (`:none:`)?  See: https://github.com/fulcrumgenomics/fgbio/blob/f72a0c3271e4a147a4b0b66b83eeebf1671ab0f2/src/main/scala/com/fulcrumgenomics/util/Metric.scala#L134.  This delineates difference between a non-empty string, an empty-string, and no string.

samfulcrum:
@nh13 just created the draft PR with unit tests we discussed https://github.com/fulcrumgenomics/fgpyo/pull/27

samfulcrum:
Proposed solution in https://github.com/fulcrumgenomics/fgpyo/pull/29

samfulcrum:
@tfenne @nh13 can we close this issue?

nh13:
Not until #29 is approved by @tfenne and merged.

samfulcrum:
@nh13 ah yes my bad I meant that comment for issue #10

PR #27, opened 2023-01-03
Created unit tests designed to fail pertaining to issue #9

samfulcrum:
The first commit to branch `issue_9` contains unit tests which are designed to fail per my conversation with @nh13.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -175,6 +175,7 @@ class Metric(ABC, Generic[MetricType]):
         parsers = cls._parsers()
         header = cls.header()
         assert len(fields) == len(header)
+        # Replaced fields with map(cls.parser_formatting, fields) SW
         return inspect.attr_from(
             cls=cls, kwargs=dict(zip(header, map(cls.parser_formatting, fields))), parsers=parsers
         )
@@ -183,7 +184,9 @@ class Metric(ABC, Generic[MetricType]):
     def parser_formatting(self, field: Optional[str]) -> Any:
         """Returns None if a field value is an empty string"""
         if not field:
-            return None
+            ex = None
+            print(type(ex))
+            return ex
         else:
             return field
 
@@ -231,9 +234,6 @@ class Metric(ABC, Generic[MetricType]):
         Args:
             value: the value to format.
         """
-        if value is None:
-            return str("")
-
         if issubclass(type(value), Enum):
             return str(cls.format_value(value.value))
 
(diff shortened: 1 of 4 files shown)


fgpyo/util/metric.py
```diff
@@ -175,7 +175,17 @@ def parse(cls, fields: List[str]) -> Any:
         parsers = cls._parsers()
         header = cls.header()
         assert len(fields) == len(header)
-        return inspect.attr_from(cls=cls, kwargs=dict(zip(header, fields)), parsers=parsers)
+        return inspect.attr_from(
+            cls=cls, kwargs=dict(zip(header, map(cls.parser_formatting, fields))), parsers=parsers
```
nh13, line 179: prefer a list comprehension since it is a lot clearer what it's doing, and there's almost never a performance degredation.
```suggestion
            cls=cls, kwargs=dict(zip(header, [cls.parser_formatting(field) for field in fields])), parsers=parsers
```

fgpyo/util/metric.py
```diff
@@ -175,7 +175,17 @@ def parse(cls, fields: List[str]) -> Any:
...
+        )
+
+    @classmethod
+    def parser_formatting(self, field: Optional[str]) -> Any:
```
nh13, line 183: 1. when is `field` ever a `None`?  In it's usage in `parse`, the `fields passed in are of type `str`.
2. do we want to expose this function as part of the public API (i.e. a public method in `Metric`).  I'd argue to keep it private for now.

fgpyo/util/metric.py
```diff
@@ -175,7 +175,17 @@ def parse(cls, fields: List[str]) -> Any:
...
+        """Returns None if a field value is an empty string"""
+        if field == "":
+            return None
+        else:
+            return field
```
nh13, line 188: One liner?
```suggestion
        return None if field == "" else field
```

fgpyo/util/metric.py
```diff
@@ -217,12 +227,9 @@ def format_value(cls, value: Any) -> str:
         Args:
             value: the value to format.
         """
-        if value == None:
+        if value is None:
```
nh13, line 230: Yes!

fgpyo/util/tests/test_metric.py
```diff
@@ -133,6 +133,7 @@ def test_metric_roundtrip(tmpdir: TmpDir, metric: DummyMetric) -> None:
     metrics: List[DummyMetric] = list(DummyMetric.read(path=path))
 
     assert len(metrics) == 1
+    print(metrics[0])
```
nh13, line 136: bingo!
```suggestion
```

fgpyo/util/tests/test_metric.py
```diff
@@ -171,8 +172,8 @@ def test_metric_parse() -> None:
     assert Person.parse(fields=["name", "42"]) == Person(name="name", age=42)
 
 
-# def test_metric_parse_with_None() -> None:
-#    assert Person.parse(fields=[None, "40"]) == Person(name='', age=40)
+def test_metric_parse_with_None() -> None:
+    assert Person.parse(fields=["", "40"]) == Person(name="None", age=40)
```
nh13, line 176: Wait, don't we want name to be `None` not `"None"`.  I.e. the value `None`, not the string with contents `"None"`?  I am not sure how tests are passing since I _think_ your code in `metric.py` is correct.
```suggestion
    assert Person.parse(fields=["", "40"]) == Person(name=None, age=40)
```

nh13: Add a unit test for testing `format_value` too

nh13 (changes requested):
I am befuddled why you're unit tests pass, can you check it out along with my other comments?

PR #29, opened 2023-01-10
Proposed solution for Issue #9

samfulcrum:
Altered `util/metric.py`, and `util/types.py` to implement requested changes in fgpyo Issue #9. 
changes in `util/metric.py` = added line to format `None` to empty string `""`.
changes in `tutil/types.py` = changed line in `none_parser()` to return `None` if value is `""` vs `str(None)`.

I also added tests to the bottom of `tests_metric.py` to test for the correct return of `""` or `None` for variables with type `Optional` in both lists and as singular values.


fgpyo/util/metric.py
```diff
@@ -233,6 +233,10 @@ def format_value(cls, value: Any) -> str:
                 return ""
             else:
                 return "{" + ",".join(cls.format_value(v) for v in value) + "}"
+
+        if value is None:
+            return ""
+
```
nh13, line 239:
```suggestion
```

nh13 (approved):
@tfenne I think we should bump the fgpyo major version due to a break change.   Thoughts?

fgpyo/util/metric.py
samfulcrum: @nils removing this makes it so that `None` will no longer be formatted as `""`. Is that the intention and I have massively misunderstood something. Or is it in the incorrect place?

samfulcrum: @nils am I correct in assuming this is meant to tell me those lines are not placed correctly in the function (ie move them lower)?

samfulcrum: I moved those lines lower in commit [6fad7c9](https://github.com/fulcrumgenomics/fgpyo/commit/6fad7c9bfe4b3332f8de3c340c7bba6e2d62ee58)

Issue #11, opened 2022-11-12
Create an Io class/module that provides some functions similar to fgbio's Io class


tfenne:
Functions to:
1. Assert one or more Paths exist and are readable
2. Assert that one or more Paths exist and are directories
3. Assert that one or more Paths are writeable (parent exists and is writable, and file exists and is writable, or does not exist).
4. `reader` function that opens a Path for reading, and based on whether it ends with a common gzip extension (.gz, .bgz, etc.) will use gzip.open() instead
5. `writer` function akin to `reader`
6. `read_lines(p: Path): List[str]` that takes a path and reads it into a list of strings, removing line terminators along the way.  Maybe also optionally strip()s lines too?
7. `write_lines(p: Path, Iterable[Any])` that will write a file with one line per item, after `str()`'ing the items

nh13:
See: https://github.com/fulcrumgenomics/commons/blob/main/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala

samfulcrum:
@nils @tfenne I apologize for not breaking this up into multiple PR's, due to the interconnected-ness I have made one branch `issue_11`.  I will hopefully be making a PR or draft PR soon, I am trying to clean it as much as possible to prevent >2 rounds of review.

nh13:
See #30

clintval:
> * Assert one or more Paths exist and are readable

Solved with [`fgpyo.io.assert_path_is_readable`](https://fgpyo.readthedocs.io/en/latest/api.html#fgpyo.io.assert_path_is_readable)

> * Assert that one or more Paths exist and are directories

Solved with [`fgpyo.io.assert_directory_exists`](https://fgpyo.readthedocs.io/en/latest/api.html#fgpyo.io.assert_directory_exists)

> * Assert that one or more Paths are writeable (parent exists and is writable, and file exists and is writable, or does not exist).

Solved with [`fgpyo.io.assert_path_is_writeable`](https://fgpyo.readthedocs.io/en/latest/api.html#fgpyo.io.assert_path_is_writeable)

> * `reader` function that opens a Path for reading, and based on whether it ends with a common gzip extension (.gz, .bgz, etc.) will use gzip.open() instead

Solved with [`fgpyo.io.to_reader`](https://fgpyo.readthedocs.io/en/latest/api.html#fgpyo.io.to_reader)

> * `writer` function akin to `reader`

[...]

PR #30, opened 2023-01-19
Proposed solution for issue #11

sam-white04:
This branch contains fgpyo/io/io.py, a module to help in reading and writing files based on the issue opened by @tfenne. The main issue I am currently facing is that 'to_reader()' and 'to_writer()' currently return Any, since the parameter 'mode' changes the object type. Error messages are based on the scala version.


fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+        assert (
+            Path(parent).is_dir() is True
+        ), f"Cannot write file because parent exists and is not a directory: {path}"
+        assert (
+            os.access(parent, os.W_OK) is True
+        ), f"Cannot write file because parent directory is not writeable: {path}"
+
+
+def to_reader(path: Path, mode: str = "rt") -> Any:
+    # TODO find compatible return type
```
nh13, line 98: probably a union type.  Something like: https://github.com/fulcrumgenomics/fgpyo/blob/2714d092734e3240ace2b06422b4b76b63071299/fgpyo/sam/__init__.py#L187

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+from pathlib import Path
+from typing import Any
+from typing import Iterable
+from typing import Set
+
+COMPRESSED: Set[str] = {".gz", ".bgz"}
+
+
+def paths_are_readable(path: Path) -> None:
+    """Checks that file exists and returns True, else raises FileNotFOundError
```
NatPRoach, line 47:
```suggestion
    """Checks that file exists and returns True, else raises FileNotFoundError
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def paths_are_writeable(path: Path, parent_must_exist: bool = True) -> None:
+    """Checks that path is writable and returns True, else raises an Exception
+    Args:
+         path: Path to be investigated
+    Example:
+    _writeable(path = Path("example.txt"))
+    """
+    assert path.exists() is True, f"Cannot write file because path is non-existantant: {path}"
```
NatPRoach, line 80:
```suggestion
    assert path.exists() is True, f"Cannot write file because path is non-existent: {path}"
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+def paths_are_readable(path: Path) -> None:
+    """Checks that file exists and returns True, else raises FileNotFOundError
+
+    Args:
+        paths: a list of one or more Paths to be investigated
+
+    Example:
+    _file_exists(path = Path("some_file.csv"))
+    """
+    assert path.exists() is True, f"Cannot read non-existant path: {path}"
```
NatPRoach, line 55:
```suggestion
    assert path.exists() is True, f"Cannot read non-existent path: {path}"
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+IO
+-------
+Module for reading and writing files.
+
+The functions in this module make is easy to:
+    check if a file exists and is writeable
+    check if a file and its parent directories exist and are writeable
+    check if a file exists and is readable
+    check if a path exists and is a directory
+    open an appropriate reader or writer based on the file extenison
```
NatPRoach, line 11:
```suggestion
    open an appropriate reader or writer based on the file extension
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+~~~~~~~~
+Examples:
+.. code-block:: python
+   >>> import fgpyo.io.io as IO
+   >>> from pathlib import Path
+   Assert that paths exist and are readable
+   >>> path_flath: Path = Path("example.txt")
+   >>> path_compressed: Path = Path("example.txt.gz")
+   >>> IO.paths_are_readable(path_flat)
+   AssertionError: Cannot read non-existant path: example.txt
```
NatPRoach, line 23:
```suggestion
   AssertionError: Cannot read non-existent path: example.txt
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,170 @@
...
+   >>> IO.paths_are_readable(compressed_file)
+   AssertionError: Cannot read non-existant path: example.txt.gz
```
NatPRoach, line 25:
```suggestion
   AssertionError: Cannot read non-existent path: example.txt.gz
```

fgpyo/io/tests/test_io.py
```diff
@@ -0,0 +1,116 @@
...
+def test_paths_are_readable_pass() -> None:
+    """Returns none when no assertions are violated"""
+    with NamedTemp(suffix=".txt", mode="w", delete=True) as write_file:
+        path = Path(write_file.name)
+        IO.paths_are_readable(path=path)
+
+
+def test_directory_exists_error() -> None:
+    """Ensure OSError when directory does not exist"""
+    path = Path("/non/existant/dir/")
```
NatPRoach, line 41:
```suggestion
    path = Path("/non/existent/dir/")
```

fgpyo/io/tests/test_io.py
```diff
@@ -0,0 +1,116 @@
+"""Basic tests for io module"""
+
+import io
```
NatPRoach, line 3: run isort on the imports here please

sam-white04: @NatPRoach I ran `isort` directly on [fgpyo/io/tests/test_io.py](https://github.com/fulcrumgenomics/fgpyo/pull/30/files#diff-1d74bdaa98107b4715a13b022dd0b47df7fca024ff9f83861aab4718e072befa) and the format is changed back after running `./ci/check.sh`

sam-white04: @NatPRoach I ran `isort` directly and the format was changed back after running `./ci/check.sh`

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+Module for reading and writing files.
+
+The functions in this module make is easy to:
+    check if a file exists and is writeable
+    check if a file and its parent directories exist and are writeable
+    check if a file exists and is readable
+    check if a path exists and is a directory
+    open an appropriate reader or writer based on the file extension
+    writitng items to a file
+    reading lines from a file
```
nh13, line 13: Please change to an unordered list in restructured text: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+from pathlib import Path
+from typing import IO
+from typing import Any
+from typing import Iterable
+from typing import Set
+from typing import TextIO
+from typing import Union
+from typing import cast
+
+COMPRESSED: Set[str] = {".gz", ".bgz"}
```
nh13, line 48:
```suggestion
COMPRESSED_FILE_EXTENSIONS: Set[str] = {".gz", ".bgz"}
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+def paths_are_readable(path: Path) -> None:
+    """Checks that file exists and returns True, else raises FileNotFoundError
+
+    Args:
+        paths: a list of one or more Paths to be investigated
+
+    Example:
+    _file_exists(path = Path("some_file.csv"))
+    """
+    assert path.exists() is True, f"Cannot read non-existent path: {path}"
```
nh13, line 60: Apply here and below since we know the return types are boolean (and not list or something else "truthy")
1. Can get rid of `is True` everywhere
2. Can change `foo is False` to `not foo`

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import cast
+
+COMPRESSED: Set[str] = {".gz", ".bgz"}
+
+
+def paths_are_readable(path: Path) -> None:
+    """Checks that file exists and returns True, else raises FileNotFoundError
+
+    Args:
+        paths: a list of one or more Paths to be investigated
```
nh13, line 55: Do you want the arg to be a list of paths?  If so, change the behavior of the function.  If not, change the docstring.

nh13 (changes requested):
Some comments that can be applied/fixed before I do a full re-review

fgpyo/io/io.py
sam-white04: Can do

fgpyo/io/io.py
sam-white04: Fixed

fgpyo/io/io.py
sam-white04: Fixed

fgpyo/io/io.py
sam-white04: Thank you for catching that

fgpyo/io/io.py
sam-white04: Fixed

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+        paths: list of one or more Paths to be investigated
+
+    Example:
+    _directory_exists(path = Path("/example/directory/"))
+    """
+    assert path.exists(), f"Path does not exist: {path}"
+    assert path.is_dir(), f"Path exists but is not a directory: {path}"
+
+
+def paths_are_writeable(path: Path, parent_must_exist: bool = True) -> None:
```
nh13, line 78: paths is plural, the argument is not

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import Iterable
+from typing import Set
+from typing import TextIO
+from typing import Union
+from typing import cast
+
+COMPRESSED_FILE_EXTENSIONS: Set[str] = {".gz", ".bgz"}
+
+
+def paths_are_readable(path: Path) -> None:
```
nh13, line 51: paths is plural, the argument is not

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    """Checks that file exists and returns True, else raises FileNotFoundError
+
+    Args:
+        path: a Path to be investigated
```
nh13, line 55: to be investigated -> to check

here and throughout

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    _directory_exists(path = Path("/example/directory/"))
+    """
+    assert path.exists(), f"Path does not exist: {path}"
+    assert path.is_dir(), f"Path exists but is not a directory: {path}"
+
+
+def paths_are_writeable(path: Path, parent_must_exist: bool = True) -> None:
+    """Checks that path is writable and returns True, else raises an Exception
+    Args:
+         path: Path to be investigated
```
nh13, line 81: missing `parent_must_exist` in docstring

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Example:
+    _writeable(path = Path("example.txt"))
+    """
+    assert path.exists(), f"Cannot write file because path is non-existent: {path}"
+    assert path.is_file(), f"Cannot write file because path is a directory: {path}"
+    assert os.access(path, os.W_OK), f"File exists but is not writebale: {path}"
+
+    if parent_must_exist:
```
nh13, line 89: To avoid long indent blocks, consider returning early, avoiding the indents after:

```python
if not parent_must_exist:
    return None
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+        ).is_dir(), f"Cannot write file because parent exists and is not a directory: {path}"
+        assert os.access(
+            parent, os.W_OK
+        ), f"Cannot write file because parent directory is not writeable: {path}"
+
+
+def to_reader(
+    path: Path, mode: str = "rt"
+) -> Union[gzip.GzipFile, io.TextIOWrapper, TextIO, IO[Any]]:
+    """Opens a Path for reading with a specifeied mode or default 'rt'.
```
nh13, line 105: typo: specifeied

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    >>> reader.close()
+    """
+    if path.suffix in COMPRESSED_FILE_EXTENSIONS:
+        return gzip.open(path, mode=mode)
+    else:
+        return path.open(mode=mode)
+
+
+def to_writer(path: Path, mode: str = "wt") -> Union[gzip.GzipFile, IO[Any], io.TextIOWrapper]:
+    # TODO find compatible return type
```
nh13, line 123: is this still a todo?

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    if path.suffix in COMPRESSED_FILE_EXTENSIONS:
+        return gzip.open(path, mode=mode)
+    else:
+        return path.open(mode=mode)
+
+
+def read_lines(path: Path, strip: bool = True) -> Any:
+    """Takes a path and reads it into a list of strings, removing line terminators
+    along the way.
+    # TODO Make line strip optional
```
nh13, line 144: can this be removed?

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    >>> writer.write(f"{something}\n")
+    >>> writer.close()
+    """
+    if path.suffix in COMPRESSED_FILE_EXTENSIONS:
+        return gzip.open(path, mode=mode)
+    else:
+        return path.open(mode=mode)
+
+
+def read_lines(path: Path, strip: bool = True) -> Any:
```
nh13, line 141: Fix return type.  I think it's currently `List[str]`, but I think we want  to return either`Iterator[str]` or `Generator[str, None, None]`.  So you'll also have to change the  code below

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Args:
+        path: Path to read from
+        strip: Boolean to strip lines or not
+
+    Example:
+    >>> read_back: List[str] = io.read_lines(path)
+    """
+    reader = to_reader(path=path)
+    list_of_lines = reader.readlines()
+    reader.close()
```
nh13, line 155: 1. why not use a `with`?
2. why not return a generator so you don't have to read in all the lines into memory?  The scala version returns an iterator for this reason
3. use f-strings rather than `str` (it's faster)
4. use `.rstrip("\r\n")` for stripping trailing line terminators (supports windows)

```python
with to_reader(path=path) as reader:
  if strip:
    for line in reader:
       yield f"{line}".strip()
  else:
    for line in reader:
       yield f"{line}".rstrip("\r\n")
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    Args:
+        path: Path to write to
+        lines_to_write: items to write to file
+
+    Example:
+    >>> lines: List[Any] = ["things to write", 100]
+    >>> path: Path = Path("file_to_write_to.txt")
+    >>> io.write_lines(path = path, lines_to_write = lines)
+    """
+    writer = to_writer(path=path)
```
nh13, line 174: again, use `with to_writer(path=path) as writer`...

fgpyo/io/io.py
```diff
@@ -0,0 +1,178 @@
...
+    # cast is needed to pass mypy
+    lines = cast(list, [f"{item}\n" for item in lines_to_write])
```
nh13, line 176: Do we really need cast
```suggestion
    lines: List[str] = [f"{item}\n" for item in lines_to_write]
```

Also, do you really need to make a list of all the lines?  That could be memory intensive.  You have an iterator as input, so just write them one by one

nh13 (changes requested):

fgpyo/io/io.py
sam-white04: Fixed here and in test

fgpyo/io/io.py
sam-white04: Fixed

fgpyo/io/io.py
sam-white04: Fixed

fgpyo/io/io.py
sam-white04: No, my bad!

fgpyo/io/io.py
sam-white04: Removed

fgpyo/io/io.py
sam-white04: From what I have found online, `cast` is needed when mypy complains about type errors (that really are not type errors) for example I had to use `cast` in `to_writer() & to_reader()` to wrap gzip in `TextIOWrapper`.

sam-white04 (commented):
@nh13 Thank you for your comments, I believe I have fixed typos, removed all TODOs, and updated `to_reader()` , `to_writer()`, `read_lines()`, and `write_lines()` according to our discussion. Thanks!

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
+"""
```
tfenne, line 1: I don't like having to `import fgpyo.io.io` - can these functions move into the module `__init_.py` please?

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+-------
+Module for reading and writing files.
+
+The functions in this module make is easy to:
+* check if a file exists and is writeable
+* check if a file and its parent directories exist and are writeable
+* check if a file exists and is readable
+* check if a path exists and is a directory
+* open an appropriate reader or writer based on the file extension
+* writitng items to a file
```
tfenne, line 12:
```suggestion
* write lines to a file
```

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+from typing import Iterator
+from typing import Set
+from typing import TextIO
+from typing import Union
+from typing import cast
+
+COMPRESSED_FILE_EXTENSIONS: Set[str] = {".gz", ".bgz"}
+
+
+def path_is_readable(path: Path) -> None:
```
tfenne, line 53: For any functions that assert like this please name them `assert_<expected-state>`.

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+    """Checks that file exists and returns True, else raises FileNotFoundError
```
tfenne, line 54: That's not true - it raises AssertionError right?

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+    parent: str = f"{path.parent.absolute()}"
+    assert Path(
+        parent
+    ).exists(), f"Cannot write file because parent diretory does not exist: {path}"
+    assert Path(
+        parent
+    ).is_dir(), f"Cannot write file because parent exists and is not a directory: {path}"
+    assert os.access(
+        parent, os.W_OK
+    ), f"Cannot write file because parent directory is not writeable: {path}"
```
tfenne, line 104: I think this is fairly far from what's desired.  The expected behavior here is:

1. _If_ file exists, it must be writeable
2. _If_ file doesn't exist and `parents_must_exist` is True, then check that `path.absolute().parent` exists, is a directory and is writeable
3. Else if file doesn't exist and `parents_must_exist` is False, recursively call parent until you find the first extent path, and check that it is a directory and is writeable.

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def to_reader(path: Path) -> Union[io.TextIOWrapper, TextIO, IO[Any]]:
+    """Opens a Path for reading with a specified mode or default 'rt'.
```
tfenne, line 108: I don't see a `mode` parameter that this implies.  This should either take a mode parameter (which is open to abuse - what if a user passes `w` as the mode?) or should be called `def open_for_text_reading(path: Path)`.

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+    >>> reader.readlines()
+    >>> reader.close()
+    """
+    if path.suffix in COMPRESSED_FILE_EXTENSIONS:
+        return io.TextIOWrapper(cast(IO[bytes], gzip.open(path, mode="rb")), encoding="utf-8")
+    else:
+        return path.open(mode="r")
+
+
+def to_writer(path: Path) -> Union[IO[Any], io.TextIOWrapper]:
```
tfenne, line 124: Same comment here about mode.
Also, perhaps this extends to both `open` functions - should they return a class that implements `ContextManager`?  Or do all types in the union(s) work as context managers, such that you could write `with io.open_for_text_writing(path) as fg`?

fgpyo/io/io.py
```diff
@@ -0,0 +1,175 @@
...
+        return path.open(mode="w")
+
+
+def read_lines(path: Path, strip: bool = True) -> Union[Iterator[str], Generator[str, None, None]]:
+    """Takes a path and reads each line into a generator, removing line terminators
+    along the way.
+
+    Args:
+        path: Path to read from
+        strip: Boolean to strip lines or not
```
tfenne, line 147: This needs more explanation - since it always rstrips() the line endings.

fgpyo/io/tests/test_io.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+from pathlib import Path
+from tempfile import NamedTemporaryFile as NamedTemp
+from typing import Any
+from typing import List
+
+import pytest
+from pytest import raises
+
+import fgpyo.io.io as IO
```
tfenne, line 13: Why alias?

tfenne (changes requested):

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Example:
+    >>> assert_path_is_writeable(path = Path("example.txt"))
+    // Assuming example.txt exists and is writeable
+    >>> example.txt
+    """
+    # If file exists, it must be writeable
+    if path.is_file():
+        assert os.access(path, os.W_OK), f"File exists but is not writebale: {path}"
+        return path
```
tfenne, line 102: This needs to be something like:
```python
if path.exists():
    assert path.is_file() and os.access(path, os.W_OK), f"..."
```

I.e. what if the path exists, but is a directory so can't be written to?

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+    elif path.is_dir() & parent_must_exist is False:
+        for directory in list(path.parents):
+            try:
+                assert (
+                    directory.is_dir() & directory.exists() & os.access(directory, os.W_OK)
+                ), f"File does not have a writeable parent directory: {path}"
+                return directory
+            except AssertionError:
+                pass
+        raise AssertionError(f"No parent directories of the path provided are writeable: {path}")
```
tfenne, line 125: This is super confusing, and I still don't think it does the right thing.  What happens if you have `/A/B/C/D.txt` as your `path`, and `C` exists and is non-writable, but `B` also exists and is writable?

Instead this should be something more like:

```python
for p in path.parents:
    if p.exists():
        assert p.is_dir() and os.access(directory, os.W_OK), f"..."
        return
raise AssertionError(f"No parent directories exist for: {path}")
```

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+    >>> writer.write(f'{something}\n')
+    >>> writer.close()
+    """
+    if path.suffix in COMPRESSED_FILE_EXTENSIONS:
+        return io.TextIOWrapper(cast(IO[bytes], gzip.open(path, mode="wb")), encoding="utf-8")
+    else:
+        return path.open(mode="w")
+
+
+def read_lines(path: Path, strip: bool = True) -> Union[Iterator[str], Generator[str, None, None]]:
```
tfenne, line 162: I think `strip` should be `False` by default as it will be less surprising.

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+    """Takes a path and reads each line into a generator, removing line terminators
+    along the way. Trailing and leading characters are stripped by default
```
tfenne, line 164:
```suggestion
    along the way. Trailing and leading whitespace is stripped by default
```

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+    however, there is an option to only strip trailing characters using the argument
+    `strip = False`.
+
+    Args:
+        path: Path to read from
+        strip: Boolean to strip lines or not
```
tfenne, line 170:
```suggestion
        strip: True to strip lines of all leading and trailing whitespace, False to only remove trailing CR/LF characters.
```

fgpyo/io/__init__.py
tfenne: If you set the return type to just `Iterator[str]` does mypy complain?  Typing docs say that should work.

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+    Example:
+    >>> read_back = io.read_lines(path)
+    """
+    with to_reader(path=path) as reader:
+        if strip:
+            for line in reader:
+                yield (f"{line!s}").strip()
+        else:
+            for line in reader:
+                yield f"{line!s}".rstrip("\r\n")
```
tfenne, line 181: Why are these not just `line.strip()` and `line.rstrip("\r\n")`?

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        lines_to_write: items to write to file
+
+    Example:
+    >>> lines: List[Any] = ["things to write", 100]
+    >>> path_to_write_to: Path = Path("file_to_write_to.txt")
+    >>> io.write_lines(path = path_to_write_to, lines_to_write = lines)
+    """
+    with to_writer(path=path) as writer:
+        for line in lines_to_write:
+            writer.write(f"{line}" + "\n")
```
tfenne, line 198:
```suggestion
        for line in lines_to_write:
            writer.write(str(line))
            writer.write("\n")
```
Using an f-string and string concatenation are unnecessary.

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
+"""
+fio
+-------
+Module for reading and writing files.
```
tfenne, line 4:
```suggestion
Module for reading and writing files
------------------------------------
```

Title should be helpful text; sphinx will usually fail if the `---` isn't the same length as the title (sigh).

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Example:
+    >>> assert_path_is_writeable(path = Path("example.txt"))
+    """
+    # If file exists, it must be writeable
+    if path.exists():
+        assert path.is_file() and os.access(
+            path, os.W_OK
+        ), f"File exists but is not writebale: {path}"
+        # return None
```
tfenne, line 104: Can you just delete the commented out return lines please.

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # Else if file doesnt exist and parent_must_exist is True then check
+    # that path.absolute().parent exists, is a directory and is writeable
+    elif parent_must_exist:
+        assert (
+            path.absolute().parent.exists()
+            & path.absolute().parent.is_dir()
+            & os.access(path.absolute().parent, os.W_OK)
+        ), f"Path does not exist and parent isn't extent/writable: {path}"
```
tfenne, line 113:
```suggestion
        parent = path.absolute().parent
        assert (
            parent.exists()
            & parent.is_dir()
            & os.access(parent, os.W_OK)
        ), f"Path does not exist and parent isn't extent/writable: {path}"
```

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        # return None
+
+    # Else if file doesn't exist and parent_must_exist is False, test parent until
+    # you find the first extent path, and check that it is a directory and is writeable.
+    else:
+        for parent in path.parents:
+            if parent.exists():
+                assert parent.exists() & os.access(
+                    parent, os.W_OK
+                ), f"File does not have a writeable parent directory: {path}"
```
tfenne, line 123:
```suggestion
                assert os.access(parent, os.W_OK), \
                    f"File does not have a writeable parent directory: {path}"
```

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+            return
```
tfenne, line 124: The return needs to be indented one more level, so it's only triggered under `if parent.exists()` after the assert.

fgpyo/io/__init__.py
```diff
@@ -0,0 +1,200 @@
...
+        return io.TextIOWrapper(cast(IO[bytes], gzip.open(path, mode="wb")), encoding="utf-8")
+    else:
+        return path.open(mode="w")
+
+
+def read_lines(path: Path, strip: bool = False) -> Iterator[str]:
+    """Takes a path and reads each line into a generator, removing line terminators
+    along the way. Trailing characters are stripped by default
+    however, there is an option to strip both leading and trailing characters using the argument
+    `strip = True`.
```
tfenne, line 166:
```suggestion
    """Takes a path and reads each line into a generator, removing line terminators
    along the way. By default only line terminators (CR/LF) are stripped.  The `strip`
    parameter may be used to strip both leading and trailing whitespace from each line.
```

tfenne (changes requested):
Getting closer.

Can you please also add a test that will fail given where the `return` is on line 124, then fix it and ensure the test passes?

Lastly, can you please try to figure out how to generate the docs and make sure they generate ok?

tfenne (approved):

nh13 (approved):

Issue #31, opened 2023-01-27
ProgressLogger should have a `record(rec: AlignedSegment)` function


tfenne:
Because progress logger wants 1-based coordinates and AlignedSegments use zero-based coordinates, this makes correct use clunky, e.g.:

```python
if rec.reference_start is not None:
    progress.record(rec.reference_name, rec.reference_start + 1)
else:
    progress.record()
```

If you don't do this, and you happen to hit a logging threshold when the last reference start is `0` the progress logger will raise an assertion error!

sam-white04:
@tfenne I am just getting into this, but for clarification are you suggesting added functionality to ProgressLogger.record() or a new function entirely?

nh13:
Since you can't have multiple functions with the same name, you have two options:

1. Create two differently named methods.  You would keep the `def record` as is, but have a separate `def record_alignment` or the like that itself calls the `record` method.
2. Have all the arguments be optional to the single `def record` method.  In this case, add a fourth parameter (e.g. `record: Optional[AlignedSegment] = None`).  In this case, you should verify (assert) that either reference_name/position are given, or `record` is given, but not both.  Then you can use the one that is defined (i.e. not `None`)

Perhaps the next step for @sam-white04 is to brainstorm the tradeoffs in a detailed manner from the perspective of the API user and implementation (we can do so over Slack), then pick one to implement?

sam-white04:
@nh13 I think I prefer option 2 listed above, but I will outline the tradeoffs as you suggested and share that with you over slack.

sam-white04:
@nils I have slacked you and @tim regarding the options outlined above

sam-white04:
@nils here is an example for solution 1 - [record_alignment()](https://github.com/fulcrumgenomics/fgpyo/blob/issue_31_v2/fgpyo/util/logging.py)
here is an example for solution 2 - [optional arguments](https://github.com/fulcrumgenomics/fgpyo/blob/issue_31/fgpyo/util/logging.py)

nh13:
Lets just go with `record_alignment` for now.

sam-white04:
@nh13 great, I will make a PR. Thanks!

clintval:
Closed by: https://github.com/fulcrumgenomics/fgpyo/pull/33

PR #33, opened 2023-02-28
Proposed solution for Issue #31

sam-white04:
Hello,
Per the discussion in Issue #31 I have added the method `record_alignment()` to handle logging of zero-based coordinates. 
Thanks, 
Sam

### DIFF
--- a/fgpyo/util/logging.py
+++ b/fgpyo/util/logging.py
@@ -163,7 +163,10 @@ class ProgressLogger(AbstractContextManager):
         Returns:
             true if a message was logged, false otherwise
         """
-        position_from_rec = rec.reference_start + 1
+        if rec.reference_start is None:
+            position_from_rec = 0
+        else:
+            position_from_rec = rec.reference_start + 1
         reference_name_from_rec = rec.reference_name
         logged = self.record(position=position_from_rec, reference_name=reference_name_from_rec)
         return logged
--- a/fgpyo/util/logging.py
+++ b/fgpyo/util/logging.py
@@ -132,25 +132,14 @@ class ProgressLogger(AbstractContextManager):
         self,
         reference_name: Optional[str] = None,
         position: Optional[int] = None,
-        rec: Optional[AlignedSegment] = None,
     ) -> bool:
         """Record an item at a given genomic coordinate.
-
         Args:
             reference_name: the reference name of the item
             position: the 1-based start position of the item
-
         Returns:
             true if a message was logged, false otherwise
         """
-        try:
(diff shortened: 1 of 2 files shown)


fgpyo/util/logging.py
```diff
@@ -151,6 +151,22 @@ def record(
         else:
             return False
 
+    def record_alignment(
+        self,
+        rec: AlignedSegment,
+    ) -> bool:
+        """Cerrectly record pysam.AlignedSegments (zero-based coordinates).
```
tfenne, line 158:
```suggestion
        """Correctly record pysam.AlignedSegments (zero-based coordinates).
```

fgpyo/util/logging.py
```diff
@@ -151,6 +151,22 @@ def record(
...
+
+        Args:
+            rec: pysam.AlignedSegment object
+
+        Returns:
+            true if a message was logged, false otherwise
+        """
+        position_from_rec = rec.reference_start + 1
+        reference_name_from_rec = rec.reference_name
+        self.record(position=position_from_rec, reference_name=reference_name_from_rec)
```
tfenne, line 168: In the issue it implies that `rec.reference_start` can return None - which would cause an error in this function.  Can you please:

1. Add tests for this function that uses `SamBuilder()` to feed it three kinds of records: a) a mapped record, b) an unmapped record with reference_name/start set, c) an unmapped read with no reference_name or reference_start set
2. Fix up this function to ensure those tests pass

tfenne: The function doesn't appear to actually return anything - there's no `return` on this last line.

tfenne (changes requested):
Please also always run `./ci/check.sh` before pushing.  This would have caught the missing return.

fgpyo/util/logging.py
```diff
@@ -151,6 +151,26 @@ def record(
...
+        Returns:
+            true if a message was logged, false otherwise
+        """
+        if rec.reference_start is None:
+            position_from_rec = 0
+        else:
+            position_from_rec = rec.reference_start + 1
+        reference_name_from_rec = rec.reference_name
+        logged = self.record(position=position_from_rec, reference_name=reference_name_from_rec)
+        return logged
```
tfenne, line 172: I think this would likely be clearer as:

```suggestion
        if rec.reference_start is None:
            return self.record(None, None)
        else:
            return self.record(rec.reference_name, rec.reference_start + 1)
```

tfenne (approved):
Approved so long as suggested change is merged and passes lints/tests.

Issue #32, opened 2023-02-13
sam.reader should support reading unmapped BAMs


nh13:
Since unmapped BAMs do not have a sequence dictionary in their header, opening with `pysam` will fail.  So one has to instead use:

```python
        with AlignmentFile(str(in_bam), mode="rb", check_sq=False) as reader:
```

Can we add an `unmapped: bool = false` option to `sam.reader` and add it to the `kwargs` passed to `sam._pysam_open`?

clintval:
Closed by: https://github.com/fulcrumgenomics/fgpyo/pull/37

PR #37, opened 2023-04-13
add support for unmapped BAM file reading

danielmarrama:
For issue #32 

`sam.reader` takes unmapped bool and passes to `_pysam_open`

```python
    return _pysam_open(
        path=path, open_for_reading=True, file_type=file_type, unmapped=unmapped
    )
```

Then `kwags["check_sq"] = False` added if unmapped and passed to `pysam.AlignmentFile`


fgpyo/sam/__init__.py
```diff
@@ -219,8 +219,8 @@ def from_path(cls, path: Union[Path, str]) -> "SamFileType":
 
 
 def _pysam_open(
-    path: SamPath, open_for_reading: bool, file_type: Optional[SamFileType] = None, **kwargs: Any
-) -> SamFile:
+    path: SamPath, open_for_reading: bool, file_type: Optional[SamFileType] = None, 
+    unmapped: bool = False, **kwargs: Any) -> SamFile:
```
nh13, line 223: don't forget to add `unmapped` to the docstring.

nh13 (changes requested):
Could you add a quick unit test to test this?

fgpyo/sam/__init__.py
danielmarrama: Yes, I forgot that. I'll add a test as well.

danielmarrama:
Added `unmapped.sam` file and unit test `test_unmapped_sam_file_open_reading_with_reader` in [test/test_sam.py](https://github.com/fulcrumgenomics/fgpyo/blob/main/fgpyo/sam/tests/test_sam.py).  Behavior should be the same with both SAM and BAM files, right? I can add another test checking BAM file reading if necessary.

nh13 (approved):

nh13:
Thanks @danielmarrama !!!

PR #41, opened 2023-05-15
Added `Alignment` class

jacarey:
Replicated the `Alignment` class from `fgbio`

### DIFF
--- a/fgpyo/alignment/__init__.py
+++ b/fgpyo/alignment/__init__.py
@@ -1,0 +1,232 @@
+from typing import Callable
+from typing import List
+
+from fgpyo.sam import Cigar
+from fgpyo.sam import CigarElement
+from fgpyo.sam import CigarOp
+
+
+class Alignment:
+    """
+    A general class to describe the alignment between two sequences or partial ranges thereof
+    """
+
+    def __init__(
+        self,
+        query: str,
+        target: str,
+        query_start: int,
+        target_start: int,
+        cigar: Cigar,
+        score: int,
+    ):
+        """
+        Args:
+            query: the query sequence
+            target: the target sequence
+            query_start: the 1-based position in the query sequence where the alignment begins
+            target_start: the 1-based position in the target sequence where the alignment begins
+            cigar: a Cigar object describing the alignment of the two sequences
+            score: the alignment score
+        """
+        self.query: str = query
+        self.target: str = target
+        self.query_start: int = query_start
+        self.target_start: int = target_start
+        self.cigar: Cigar = cigar
+        self.score: int = score
+
+    @property
+    def query_end(self) -> int:
+        """
+        Returns:
+            One based closed coordinate of the end of the alignment on the query sequence.
+        """
+        return self.query_start + self.cigar.length_on_query() - 1
+
+    @property
+    def target_end(self) -> int:
+        """
+        Returns:
+            One based closed coordinate of the end of the alignment on the target sequence.
+        """
+        return self.target_start + self.cigar.length_on_target() - 1
+
+    def padded_string(
+        self,
+        match_char: str = "|",
+        mismatch_char: str = ".",
+        gap_char: str = " ",
+        pad_char: str = "-",
+    ) -> List[str]:
+        """Generates a padded text representation of the alignment for visualization.
+        The returned sequence will consist of three lines as follows (minus the labels on the left):
+
+        query : ACGTGAACTGACT-ACTGTATGCG
+        align : ||||| |||||| ||||||||.|
+        target: ACGTG--CTGACTGACTGTATGGG
+
+        Args:
+            match_char: The character to use in the alignment line when the bases match.
+            mismatch_char: The character to use in the alignment line when the bases mismatch.
+            gap_char: The character to use in the alignment line when one of the
+                sequences is gapped.
+            pad_char: The character to use in the query or target sequence lines when
+                padding is required.
+
+        Returns:
+            Three lines representing the alignment.
+        """
+        query_buffer = []
+        align_buffer = []
+        target_buffer = []
+
+        q_offset = self.query_start - 1
+        t_offset = self.target_start - 1
+
+        for cig in self.cigar.elements:
+            length = cig.length
+
+            if cig.operator == CigarOp.I:
+                for _ in range(length):
+                    query_buffer.append(self.query[q_offset])
+                    align_buffer.append(gap_char)
+                    target_buffer.append(pad_char)
+                    q_offset += 1
+            elif cig.operator == CigarOp.D:
+                for _ in range(length):
+                    query_buffer.append(pad_char)
+                    align_buffer.append(gap_char)
+                    target_buffer.append(self.target[t_offset])
+                    t_offset += 1
+            else:
+                for _ in range(length):
+                    q = self.query[q_offset]
+                    t = self.target[t_offset]
+                    query_buffer.append(q)
+                    target_buffer.append(t)
+
+                    if cig.operator == CigarOp.EQ:
+                        align_buffer.append(match_char)
+                    elif cig.operator == CigarOp.X:
+                        align_buffer.append(mismatch_char)
+                    elif cig.operator == CigarOp.M:
+                        align_buffer.append(match_char if q == t else mismatch_char)
+                    else:
+                        raise ValueError(f"Unsupported cigar operator: {cig.operator}")
+
+                    q_offset += 1
+                    t_offset += 1
+
+        return ["".join(query_buffer), "".join(align_buffer), "".join(target_buffer)]
+
+    def sub_by_query(self, start: int, end: int) -> "Alignment":
+        """Returns a subset of the Alignment representing the region defined by start
+        and end on the query sequence. The returned alignment will contain the entire
+        query and target sequences, but will have adjusted query_start and target_start
+        positions and an updated cigar. The score is set to 0.
+
+        Args:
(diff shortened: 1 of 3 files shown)


fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
+from typing import Callable
```
nh13, line 1: I'd move this to `fgpyo/alignment.py` and then move the tests similarly.  Any reason to keep it this way?

nh13: Don't forget to add module docs.  Can you also add an example or two of the padded string method?  See examples in other modules: https://github.com/fulcrumgenomics/fgpyo/blob/5ef0f5e9ced1e8e3763e6858317b59e448193635/fgpyo/read_structure.py#L16-L49

```suggestion
"""
Classes for representing Alignments
-----------------------------------

<INSERT an example or TWO HERE>

Module Contents
~~~~~~~~~~~~~~~
The module contains the following public classes:
    - :class:`~fgpyo.alignment.Alignment` -- Describes the alignment between two sequences 

"""


from typing import Callable
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        cigar: Cigar,
+        score: int,
+    ):
+        """
+        Args:
+            query: the query sequence
+            target: the target sequence
+            query_start: the 1-based position in the query sequence where the alignment begins
+            target_start: the 1-based position in the target sequence where the alignment begins
+            cigar: a Cigar object describing the alignment of the two sequences
```
nh13, line 29: I tend not to use "class" or "object" in docstrings since they we know they are classes or objects (since that's what we're document)
```suggestion
            cigar: the Cigar describing the alignment of the two sequences
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+from typing import List
+
+from fgpyo.sam import Cigar
+from fgpyo.sam import CigarElement
+from fgpyo.sam import CigarOp
+
+
+class Alignment:
+    """
+    A general class to describe the alignment between two sequences or partial ranges thereof
```
nh13, line 11:
```suggestion
    Describes the alignment between two sequences.
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @property
+    def target_end(self) -> int:
+        """
+        Returns:
+            One based closed coordinate of the end of the alignment on the target sequence.
+        """
+        return self.target_start + self.cigar.length_on_target() - 1
+
+    def padded_string(
```
nh13, line 55: this is the main reason this class is here, so I'd highlight it in the module docs in an example

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        self,
+        match_char: str = "|",
+        mismatch_char: str = ".",
+        gap_char: str = " ",
+        pad_char: str = "-",
+    ) -> List[str]:
```
nh13, line 61: prefer a tuple when we have a fixed # of items returned (or use a namedtuple/dataclass, but a tuple is best here).  That allows us to unpack on assignment: `q, a, t = alignment.padded_string(...)`
```suggestion
    ) -> Tuple[str, str, str]:
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            match_char: The character to use in the alignment line when the bases match.
+            mismatch_char: The character to use in the alignment line when the bases mismatch.
+            gap_char: The character to use in the alignment line when one of the
+                sequences is gapped.
+            pad_char: The character to use in the query or target sequence lines when
+                padding is required.
+
+        Returns:
+            Three lines representing the alignment.
+        """
```
nh13, line 79: check that all the `*_char` _strings_ are of the same length

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        query_buffer = []
+        align_buffer = []
+        target_buffer = []
+
+        q_offset = self.query_start - 1
+        t_offset = self.target_start - 1
+
+        for cig in self.cigar.elements:
```
nh13, line 87: This bugged me in fgbio
```suggestion
        for elem in self.cigar.elements:
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            length = cig.length
+
+            if cig.operator == CigarOp.I:
+                for _ in range(length):
+                    query_buffer.append(self.query[q_offset])
+                    align_buffer.append(gap_char)
+                    target_buffer.append(pad_char)
+                    q_offset += 1
```
nh13, line 95: I think you can remove the loops here and in the `CigarOp.D` case
```suggestion
                query_buffer.append(self.query[q_offset: q_offset + length])
                align_buffer.append(gap_char * length)
                target_buffer.append(pad_char * length)
                q_offset += length
```

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+                sequences is gapped.
+            pad_char: The character to use in the query or target sequence lines when
+                padding is required.
+
+        Returns:
+            Three lines representing the alignment.
+        """
+        query_buffer = []
+        align_buffer = []
+        target_buffer = []
```
nh13, line 82: Why append to lists and the join at the end?  Why not just build the strings directly
```suggestion
        query_buffer: str = ""
        align_buffer: str = ""
        target_buffer: str = ""
```
then later: `query_buffer += self.query[q_offset]`

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
+from typing import Callable
+from typing import List
+
+from fgpyo.sam import Cigar
+from fgpyo.sam import CigarElement
+from fgpyo.sam import CigarOp
+
+
+class Alignment:
```
nh13, line 9: Any reason not to make this a frozen (i.e. immutable) attr class? See: https://github.com/fulcrumgenomics/fgpyo/blob/5ef0f5e9ced1e8e3763e6858317b59e448193635/fgpyo/sam/__init__.py#L511

For container classes, we should use attrs, dataclasses, or named tuples I think.

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+                sequence to include.
+
+        Returns:
+            A new Alignment with updated coordinates and cigar.
+        """
+        if not (
+            self.query_start <= start <= self.query_end
+            and self.query_start <= end <= self.query_end
+        ):
+            raise ValueError("start or end is outside of aligned region of target sequence")
```
nh13, line 142: it's more useful to have the conditions separated so you know if it start or end that's invalid.  And why not use `assert`?

```suggestion
        assert self.query_start <= start <= self.query_end, f"start '{start}' is outside of aligned region of query sequence "
        assert self.query_start <= end <= self.query_end, f"end '{end}' is outside of aligned region of query sequence "
```

Also, we carried over a bug from fgbio (it's query, not target): https://github.com/fulcrumgenomics/fgbio/pull/914

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        Args:
+            start (int): The 1-based inclusive position of the first base on the target sequence
+                to include.
+            end (int): The 1-based inclusive position of the last base on the target sequence
+                to include.
+
+        Returns:
+            A new Alignment with updated coordinates and cigar.
+        """
+        if not (
```
nh13, line 161: ditto

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+                        t_start += diff
+                    curr_start += diff
+
+                if curr_end > end:
+                    length -= curr_end - end
+
+                elems.append(CigarElement(length, elem.operator))
+                curr_start += length
+
+        return Alignment(
```
nh13, line 225: if you make this an attr class, you can do `attr.evolve` :)

fgpyo/alignment/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            self.target_start <= start <= self.target_end
+            and self.target_start <= end <= self.target_end
+        ):
+            raise ValueError("start or end is outside of aligned region of target sequence")
+
+        return self.sub(
+            start, end, self.target_start, lambda elem: elem.operator.consumes_reference
+        )
+
+    def sub(
```
nh13, line 171: Make "private" as the docstring suggest
```suggestion
    def _sub(
```

fgpyo/sam/__init__.py
```diff
@@ -332,7 +332,7 @@ class CigarOp(enum.Enum):
         code (int): The :py:mod:`~pysam` cigar operator code.
         character (int): The single character cigar operator.
         consumes_query (bool): True if this operator consumes query bases, False otherwise.
-        consumes_target (bool): True if this operator consumes target bases, False otherwise.
+        consumes_reference (bool): True if this operator consumes target bases, False otherwise.
```
nh13, line 335:
```suggestion
        consumes_reference (bool): True if this operator consumes reference bases, False otherwise.
```

nh13 (changes requested):
Looking really good, a number of comments to address and then I'll take one final look!

PR #48, opened 2023-07-27
Suppress stderr printing when sam indices are older than sam file

TedBrookings:
Merging into the [unmerged feature/add-vcf](https://github.com/fulcrumgenomics/fgpyo/pull/47) PR because it moves the `redirect_dev_null` context manager to a common location (`fgpyo.io`) so that it can be used by `vcf` and `sam` modules.

I'm actually not sure how to write unit tests for this functionality, I've weirdly not been successful in extracting text from pysam errors by redirecting to a non-dev-null file. However, I did verify in manual testing that this does what it says on the tin.

### DIFF
--- a/fgpyo/io/__init__.py
+++ b/fgpyo/io/__init__.py
@@ -44,9 +44,12 @@ fgpyo.io Examples:
 import gzip
 import io
 import os
+import sys
+from contextlib import contextmanager
 from pathlib import Path
 from typing import IO
 from typing import Any
+from typing import Generator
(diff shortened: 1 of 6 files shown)


TedBrookings:
I liked the latter suggestion so I did that.

tfenne (approved):

tfenne:
Looks good to me @TedBrookings - sorry for taking so long to circle back.  Want to resolve conflicts and merge?

PR #50, opened 2023-08-04
Metric.read should support skipping extra columns

nh13:
This is a breaking change, since previously it would default to an Assertion Error.

I also fixed a few docstring issues.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -169,17 +169,18 @@ class Metric(ABC, Generic[MetricType]):
             missing_from_class = file_fields.difference(class_fields)
             missing_from_file = class_fields.difference(file_fields)
 
-            # ignore optional class fields that are missing from the file (via header), they'll get
-            # a default of None
+            field_name_to_attribute = attr.fields_dict(cls)
+
+            # ignore class fields that are missing from the file (via header) if they're optional
+            # or have a default
             if len(missing_from_file) > 0:
-                field_name_to_attribute = attr.fields_dict(cls)
-                missing_from_file_optionals = [
+                fields_with_defaults = [
                     field
                     for field in missing_from_file
-                    if inspect.attribute_is_optional(field_name_to_attribute[field])
+                    if inspect.attribute_has_default(field_name_to_attribute[field])
                 ]
                 # remove optional class fields from the fields
-                missing_from_file = missing_from_file.difference(missing_from_file_optionals)
+                missing_from_file = missing_from_file.difference(fields_with_defaults)
 
             # raise an exception if there are non-optional class fields missing from the file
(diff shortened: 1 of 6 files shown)


fgpyo/util/metric.py
```diff
@@ -148,18 +148,25 @@ def _parsers(cls) -> Dict[type, Callable[[str], Any]]:
         return {}
 
     @classmethod
-    def read(cls, path: Path) -> Iterator[Any]:
+    def read(cls, path: Path, skip_extra: bool = True) -> Iterator[Any]:
```
tfenne, line 151: How about `ignore_extra_fields`?  In the docstring you also explain it as "ignore" which I think is a better term.

fgpyo/util/metric.py
```diff
@@ -148,18 +148,25 @@ def _parsers(cls) -> Dict[type, Callable[[str], Any]]:
...
+            skip_extra: True to ignore any extra columns, False to raise an exception.
         """
         parsers = cls._parsers()
         with path.open("r") as reader:
             header: List[str] = reader.readline().rstrip("\r\n").split("\t")
-            assert header == cls.header(), "Header did not match"
+            cls_header = cls.header()
+            # check the header
+            for field in cls_header:
+                assert field in header, f"Missing field '{field}' in file: {path}"
```
tfenne, line 166: I would _love_ it if this check was also allowed (optionally?) to skip over situations where a) the class has a field, b) it's not in the file, c) the field is `Optional[_]` in the class.  I.e. would be nice to be able to evolve classes adding optional fields, and be able to parse old files.

fgpyo/util/metric.py
```diff
@@ -148,18 +148,25 @@ def _parsers(cls) -> Dict[type, Callable[[str], Any]]:
...
+            if not skip_extra:
+                for field in header:
+                    assert field in cls_header, f"Extra field '{field}' in file: {path}"
```
tfenne, line 169: For both of these it would be nicer to do:

```python
class_fields = set(cls_header)
file_fields = set(header)
missing_from_class = file_fields.difference(class_fields)
missing_from_file = class_fields.difference(file_fields)

if len(missing_from_class) > 0:
   raise ValueError("Fields in file missing from class: {missing_from_class}")
...
```

tfenne: Also would be helpful to state the class name in the errors along with the file name.

tfenne (approved):
After addressing comments :)

fgpyo/util/metric.py
nh13: I love that idea

nh13: I did this!

tfenne (approved):
Very nice, thank you!

PR #58, opened 2023-10-17
feat: add header property

msto:
Added a property to expose the private `_samheader` attribute

### DIFF
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -558,3 +558,8 @@ class SamBuilder:
             path = self.to_path(path=Path(filename), index=False)
             bam = sam.reader(path)
             return list(bam)
+
+    @property
+    def header(self) -> AlignmentHeader:
+        """Returns the builder's SAM header."""
+        return self._samheader


tfenne (approved):

fgpyo/sam/builder.py
```diff
@@ -558,3 +558,8 @@ def to_sorted_list(self) -> List[pysam.AlignedSegment]:
             path = self.to_path(path=Path(filename), index=False)
             bam = sam.reader(path)
             return list(bam)
+
+    @property
+    def header(self) -> AlignmentHeader:
+        """Returns the builder's SAM header."""
+        return self._samheader
```
nh13, line 565: Should this be a deep copy?  Otherwise folks can mutate it. 

What's the behavior if the header is mutated _after_ records have been created (since the header is used to create an `AlignedSegment`)?  See:

https://github.com/fulcrumgenomics/fgpyo/blob/e3349697bb69e0f8cdb0d1be5fa53ec71a54353d/fgpyo/sam/builder.py#L174

nh13 (changes requested):
We have to be careful about giving access to the sam header since it is used to create the `AlignedSegment`.  What are the tradeoffs?

fgpyo/sam/builder.py
msto: I don't think you can mutate an `AlignmentHeader`, e.g.

```py
In [22]: header['HD']
Out[22]: {'VN': '1.5', 'SO': 'coordinate'}

In [23]: header['HD'] = {'VN': '1.5', 'SO': 'unsorted'}
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[23], line 1
----> 1 header['HD'] = {'VN': '1.5', 'SO': 'unsorted'}

File ~/.mambaforge-x86/envs/uditas/lib/python3.10/site-packages/pysam/libcalignmentfile.pyx:550, in pysam.libcalignmentfile.AlignmentHeader.__setitem__()

TypeError: AlignmentHeader does not support item assignment (use header.to_dict()
```

and

```py
In [32]: header['RG']
Out[32]:
[{'ID': '1',
  'SM': '1_AAAAAA',
  'LB': 'default',
  'PU': 'xxx.1',
  'PL': 'ILLUMINA'}]

In [33]: header['RG'].pop()
Out[33]: {'ID': '1', 'SM': '1_AAAAAA', 'LB': 'default', 'PU': 'xxx.1', 'PL': 'ILLUMINA'}

In [34]: header['RG'][0].pop('PL')
Out[34]: 'ILLUMINA'

In [35]: header['RG']
Out[35]:
[{'ID': '1',
  'SM': '1_AAAAAA',
  'LB': 'default',
  'PU': 'xxx.1',
  'PL': 'ILLUMINA'}]
  ```

Comment (msto) [reply fgpyo/sam/builder.py:565]: shall I update to 

```py
        return self._samheader.copy()
```

?

nh13: Nah, this is good enough for me.  Ship it

nh13 (approved):

PR #59, opened 2023-10-17
feat: add set_tag method to Template

msto:
This is something I've implemented twice in client projects, so thought it'd be sensible to add directly to the library


fgpyo/sam/__init__.py
```diff
@@ -843,6 +843,21 @@ def all_recs(self) -> Iterator[AlignedSegment]:
...
+        tag: str,
+        value: Union[str, int, float],
+    ) -> None:
+        """Add a tag to all records associated with the template.
+
+        Args:
+            tag: The name of the tag.
+            value: The value of the tag.
+        """
+
```
tfenne, line 857: want to require hat `tag` is two characters?

fgpyo/sam/__init__.py
```diff
@@ -843,6 +843,21 @@ def all_recs(self) -> Iterator[AlignedSegment]:
             for rec in recs:
                 yield rec
 
+    def set_tag(
+        self,
+        tag: str,
+        value: Union[str, int, float],
```
tfenne, line 849: I think in pysam, if you want to clear a tag you set it to `None`, if that's right could we add `None` to the union too please?

tfenne (approved):
Approved with suggestions.

fgpyo/sam/__init__.py
msto: done

fgpyo/sam/__init__.py
msto: done

nh13 (approved):
one minor suggestion to say the tag!

Issue #55, opened 2023-09-08
SamBuilder should permit addition of discordant pairs


msto:
`add_pair()` requires both reads to align to the same contig, if aligned

msto:
fixed in #61

PR #61, opened 2023-10-17
feat: add chrom1/chrom2 arguments to add_pair

msto:
Addressing #55, it would be helpful to be able to create pairs where the reads map to different contigs.

I've changed the signature for `SamBuilder.add_pair()` to accept `chrom1` and `chrom2`, instead of `chrom`. When `chrom2` is not specified, the value passed to `chrom1` will be used for both reads.

As written, this will be a breaking change for most calls to `add_pair()`, which likely use `chrom`. To minimize side effects elsewhere, we could use `chrom` and `chrom2` instead, but this conflicts with the pattern used for all other arguments (`start1`/`start2`, etc). I think it'd be better to rip the bandaid once and have this fixed, but lmk if there's a cleaner way to make the update.

### DIFF
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -340,7 +340,8 @@ class SamBuilder:
         bases2: Optional[str] = None,
         quals1: Optional[List[int]] = None,
         quals2: Optional[List[int]] = None,
-        chrom1: str = sam.NO_REF_NAME,
+        chrom: str = sam.NO_REF_NAME,
+        chrom1: Optional[str] = None,
         chrom2: Optional[str] = None,
         start1: int = sam.NO_REF_POS,
         start2: int = sam.NO_REF_POS,
@@ -380,8 +381,9 @@ class SamBuilder:
             bases2: The bases for R2. If None is given a random sequence is generated.
             quals1: The list of int qualities for R1. If None, the default base quality is used.
             quals2: The list of int qualities for R2. If None, the default base quality is used.
-            chrom1: The chromosome to which R1 is mapped. Defaults to the unmapped value.
-            chrom2: The chromosome to which R2 is mapped. If None, `chrom1` is used.
+            chrom: The chromosome to which both reads are mapped. Defaults to the unmapped value.
+            chrom1: The chromosome to which R1 is mapped. If None, `chrom` is used.
+            chrom2: The chromosome to which R2 is mapped. If None, `chrom` is used.
             start1: The start position of R1. Defaults to the unmapped value.
             start2: The start position of R2. Defaults to the unmapped value.
             cigar1: The cigar string for R1. Defaults to None for unmapped reads, otherwise all M.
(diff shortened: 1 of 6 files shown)


msto:
This would be backwards compatible but is ugly imo

```py
def add_pair(
    # ...,
    chrom: str = sam.NO_REF_NAME,
    chrom1: Optional[str] = None,
    chrom2: Optional[str] = None,
    # ...,
) -> Tuple[AlignedSegment, AlignedSegment]:
    # ...

    # Use `chrom` for both reads if `chrom1` and `chrom2` are not specified.
    if chrom1 is None and chrom2 is None:
        chrom1 = chrom2 = chrom
    elif chrom1 is None or chrom2 is None:
        raise ValueError("When using chrom1 or chrom2, both must be specified.")


```

tfenne:
I don't feel good about breaking back-compat for this - it will likely break 100s of tests across many code-bases.  @nh13 what do you think?

I can think of a couple of other options:

1. Define a new function that takes `chrom1` and `chrom2` and vector the other function to it.  I'm having a hard time coming up with a clear and short name though.
2. Leave `chrom` in add_pair (and perhaps deprecate it?) _and_ add `chrom1` and `chrom2` and have clear rules about precedence (or just fail if called with values for `chrom` and `chrom1`?)

Or just leave it at `chrom`/`chrom2` and accept the annoying difference from all the other fields.

I think I'd vote for (2) above.

msto:
I updated so the following combinations of parameters are permitted:

```
# Backwards compatible:
- chrom, start1, start2
- chrom, start1
- chrom, start2

# New:
- chrom1, start1, chrom2, start2
- chrom1, start1
- chrom2, start2
```

nh13 (approved):
Nice, very thoughtful.

fgpyo/sam/builder.py
```diff
@@ -405,15 +420,49 @@ def add_pair(
...
+        # - chrom1, start1, chrom2, start2
+        # - chrom1, start1
+        # - chrom2, start2
+
+        # Fail if chrom is specified with chrom1 or chrom2
+        if chrom != sam.NO_REF_NAME and (chrom1 is not None or chrom2 is not None):
+            raise ValueError("chrom cannot be specified together with chrom1 or chrom2.")
+
+        # Backwards compatibility
+        # Use `chrom` for both reads if neither `chrom1` nor `chrom2` are specified.
```
tfenne, line 437: I _think_ this is overly restrictive, but I'm having a hard time thinking through all the cases.  Would it not suffice to do something like:

```
if chrom is None and (chrom1 is not None or chrom2 is not None):
    raise ValueError("Cannot use chrom in combination with chrom1 or chrom2")

if chrom is None and chrom1 is None and chrom2 is None:
    chrom = sam.NO_REF_NAME

chrom1 = next(c for c in (chrom2, chrom) if c is not None)
chrom2 = next(c for c in (chrom1, chrom) if c is not None)
```

tfenne (commented):
Might make sense to huddle and discuss how to handle chrom/chrom1/chrom2.

fgpyo/sam/builder.py
```diff
@@ -405,15 +420,49 @@ def add_pair(
...
+        # - chrom, start1, start2
+        # - chrom, start1
+        # - chrom, start2
+        # Valid parameterizations for contig mapping (new):
+        # - chrom1, start1, chrom2, start2
+        # - chrom1, start1
+        # - chrom2, start2
+
+        # Fail if chrom is specified with chrom1 or chrom2
+        if chrom != sam.NO_REF_NAME and (chrom1 is not None or chrom2 is not None):
```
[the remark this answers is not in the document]
msto: I hadn't wanted to change the function signature but I agree

fgpyo/sam/builder.py
msto: Yeah I find this clearer.

Just pushed, with additional checks to ensure a chromosome is always specified when providing a start position

https://github.com/fulcrumgenomics/fgpyo/pull/61/commits/8a55ab18d68aa57b09ec59edbada88e8ef30226f

tfenne (approved):

PR #64, opened 2023-10-18
ci: require that the poetry lock file is up to date

nh13:


nh13:
@msto so basically I had to:
1. update the version of poetry we require to 1.5
2. add `poetry lock --check` to our github action
3. add the `sphinx_rtd_theme` dependency to the `pyproject.yaml`.  I think you had added it manually as part of the `docs`, but didn't add it to the main dependency.  Take a look.  The typical way to add a new dependency is with `poetry add`.
4. cherry-pick ed6df15a585c0ba691bf3efd7fb7033fae96fd13 from #65

msto (approved):
🙌

Issue #67, opened 2023-10-21
Support stdin/stdout in sam.reader and sam.writer


msto:
It would be helpful if `sam.reader()` and `sam.writer()` supported the special cases of `"-"` and/or `"stdin"` and `"stdout"` to open `sys.stdin` and `sys.stdout` directly for reading/writing

to avoid verbosity like the following

```py
in_bam_path: SamPath = sys.stdin if str(in_bam) in ["-", "stdin"] else in_bam

out_bam_path: SamPath
out_file_type: Optional[SamFileType]
if str(out_bam) in ["-", "stdout"]:
    out_bam_path = sys.stdout
    out_file_type = SamFileType.SAM
else:
    out_bam_path = out_bam
    out_file_type = None
    
with (
    reader(in_bam_path) as bam_reader,
    writer(
        out_bam_path,
        header=bam_reader.header,
        file_type=out_file_type,
    ) as bam_writer,
```

nh13:
Implemented in https://github.com/fulcrumgenomics/fgpyo/pull/69

PR #69, opened 2023-10-25
feat: Support stdin/stdout special cases in sam.reader/writer

msto:
Addressing https://github.com/fulcrumgenomics/fgpyo/issues/67

@nh13 @tfenne would you entertain something like this?

Two open questions 

- whether writing to `stdout` should default to SAM or BAM
- I'm not sure how best to test these. In the past when unit testing against `sys.stdin` I've used `pytest.MonkeyPatch` and a `StringIO`, but `pysam` doesn't support reading from `StringIO`

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -153,6 +153,7 @@ The module contains the following methods:
 
 import enum
 import io
+import sys
 from pathlib import Path
 from typing import IO
 from typing import Any
@@ -188,6 +189,12 @@ NO_REF_POS: int = -1
 
 _IOClasses = (io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase)
 """The classes that should be treated as file-like classes"""
+
+_STDIN_PATHS: List[str] = ["-", "stdin", "/dev/stdin"]
+"""Paths that should be opened as stdin."""
+
+_STDOUT_PATHS: List[str] = ["-", "stdout", "/dev/stdout"]
+"""Paths that should be opened as stdout."""
 
 
 @enum.unique
@@ -230,6 +237,12 @@ def _pysam_open(
 ) -> SamFile:
     """Opens a SAM/BAM/CRAM for reading or writing.
 
+    This function permits reading from standard input and writing to standard output. The specified
+    path may be the UNIX conventional `"-"`, the more explicit `"stdin"` or `"stdout"`, or an
+    absolute path to either of the standard streams `"/dev/stdin"` or `"/dev/stdout"`.
+
+    When writing to standard output, the file type must be specified.
+
     Args:
         path: a file handle or path to the SAM/BAM/CRAM to read or write.
         open_for_reading: True to open for reading, false otherwise.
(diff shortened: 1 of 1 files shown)


fgpyo/sam/__init__.py
```diff
@@ -187,6 +188,12 @@
 _IOClasses = (io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase)
 """The classes that should be treated as file-like classes"""
 
+_STDIN_PATHS = ["-", "stdin", "/dev/stdin"]
```
nh13, line 191:
```suggestion
_STDIN_PATHS: List[str] = ["-", "stdin", "/dev/stdin"]
```

fgpyo/sam/__init__.py
```diff
@@ -187,6 +188,12 @@
...
+"""Paths that should be opened as stdin."""
+
+_STDOUT_PATHS = ["-", "stdout", "/dev/stdout"]
```
nh13, line 194:
```suggestion
_STDOUT_PATHS: List[str]  = ["-", "stdout", "/dev/stdout"]
```

nh13 (changes requested):
1. I think we should remove it default to SAM when writing to stdout and keep the requirement that `file_type` must be given when writing (the Args needed to be updated otherwise).
2. See here for how pysam tests stdin/stdout:
https://github.com/pysam-developers/pysam/blob/cdc0ed12fbe2d7633b8fa47534ab2c2547f66b84/tests/StreamFiledescriptors_test.py#L47-L65

msto:
I would argue that it is helpful to set a default `file_type` for stdout.

Essentially, I want to be able to do the following:

```
def my_tool(
    *,
    in_bam: Path,
    out_bam: Path,
):
    """
    Run my tool.
    """

    with (
        reader(in_bam) as bam_reader,
        writer(out_bam, header=bam_reader.header) as bam_writer
    ):
        pass
```

If `writer` doesn't have a default `file_type` when writing to `stdout`, the user has to manually check whether the output path is `stdout`. Avoiding that was the motivation for this PR

nh13:
I think it's a footgun to have it default to standard output in library code.  Let's go without it defaulting please.

tfenne:
I agree with Nils about not defaulting to SAM.  If anything I would say it should default to uncompressed BAM as that's what I'd want when piping between tools ... and if a user wrote directly to stdout you'd quickly see it.  But this points out why it should probably not have a default - because I'd argue for a different one.

I think the PR also needs to handle what if a user passes e.g. `Path("/dev/stdin")`

msto:
Will circle back to add tests, but made the requested changes.

Follow-up question: When opening from a filepath, should we require that the file type be inferred from the file extension, rather than giving the user the opportunity to specify a `file_type` that conflicts with the file extension?

https://github.com/fulcrumgenomics/fgpyo/blob/002d75bea5df7a268a512c21e8565f4a8736dd8d/fgpyo/sam/__init__.py#L255

nh13:
I think the follow-up question deserves a separate issue.  See comment in the code https://github.com/pysam-developers/pysam/issues/655 and a similar issue our other rust repo: https://github.com/fulcrumgenomics/fgoxide/pull/9.

fgpyo/sam/__init__.py
```diff
@@ -241,8 +254,14 @@ def _pysam_open(
     """
 
     if isinstance(path, (str, Path)):  # type: ignore
-        file_type = file_type or SamFileType.from_path(path)
-        path = str(path)
+        if str(path) in _STDIN_PATHS and open_for_reading:
+            path = sys.stdin
```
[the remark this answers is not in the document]
msto: https://github.com/fulcrumgenomics/fgpyo/issues/131

PR #71, opened 2023-11-07
Fix install and test

TedBrookings:
* Fix failing unit test
* Bump pysam version to fix install problems


nh13 (approved):
Ignore code cov for this one

PR #75, opened 2023-11-22
Fix type problems found by latest mypy

### LINKED ISSUE #73: Newer versions of mypy reject multiple lines of fgpyo
* I inadvertently bumped the version of `mypy`  (to 1.7.0) while making changes to other packages and discovered that there are a lot of lines of `fgpyo` that fail newer versions of mypy.
* Not super-urgent at the moment, although eventually old versions get deprecated, `pytest` depends on newer ones, etc.
* Most of the fixes appear to be very simple, but there is one complicated section that I think would benefit from @nh13 looking at it:
There were some issues with the use of `Literal` in [fgpyo/util/types.py.](https://github.com/fulcrumgenomics/fgpyo/blob/ca4a572524d2e5f79838250603e2fcf4d5d63155/fgpyo/util/types.py#L58) Basically it's illegal to define a generic `Literal`, so I think it's deliberately impossible to create something like `LiteralType = TypeVar("LiteralType", bound="Literal")`. Now mypy has gotten more hardcore and rejects that line. There are several related functions in `util/types.py` and `util/inspect.py`.
* I created a branch with all the simple fixes: [feature/support-updated-mypy](https://github.com/fulcrumgenomics/fgpyo/tree/feature/support-updated-mypy)

TedBrookings:
* Move to current latest version of pytest to avoid deprecation warning
* Move to current latest version of mypy
* Correct multiple problems with type and import flagged by newer mypy
Closes #73

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -118,7 +118,7 @@ import attr
 from fgpyo import io
 from fgpyo.util import inspect
 
-MetricType = TypeVar("MetricType")
+MetricType = TypeVar("MetricType", bound="Metric")
 
 
 @attr.s
@@ -137,7 +137,7 @@ class Metric(ABC, Generic[MetricType]):
         """An iterator over attribute values in the same order as the header."""
         return iter(attr.astuple(self, recurse=False))
 
-    def formatted_values(self) -> Iterator[str]:
+    def formatted_values(self) -> List[str]:
         """An iterator over formatted attribute values in the same order as the header."""
         return [self.format_value(value) for value in self.values()]
 
--- a/fgpyo/util/types.py
+++ b/fgpyo/util/types.py
@@ -1,5 +1,6 @@
 import collections
 import inspect
+import sys
 import typing
 from enum import Enum
 from functools import partial
@@ -9,20 +10,19 @@ from typing import Type
 from typing import TypeVar
 from typing import Union
 
-try:
-    from typing import Literal
-except ImportError:
-    from typing_extensions import Literal
-
-
 # `get_origin_type` is a method that gets the outer type (ex list in a List[str])
-if hasattr(typing, "get_origin"):  # py>=38
+# `get_arg_types` is a method that gets the inner type (ex str in a List[str])
+if sys.version_info >= (3, 8):
+    from typing import Literal
+
     get_origin_type = typing.get_origin
-else:  # py<38
+    get_arg_types = typing.get_args
+else:
+    import typing_inspect
+    from typing_extensions import Literal
 
     def get_origin_type(tp: Type) -> Type:
         """Returns the outer type of a Typing object (ex list in a List[T])"""
-        import typing_inspect
 
         if type(tp) is type(Literal):  # Py<=3.6.
             return Literal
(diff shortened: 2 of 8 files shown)


fgpyo/util/metric.py
```diff
@@ -137,7 +137,7 @@ def values(self) -> Iterator[Any]:
         """An iterator over attribute values in the same order as the header."""
         return iter(attr.astuple(self, recurse=False))
 
-    def formatted_values(self) -> Iterator[str]:
+    def formatted_values(self) -> List[str]:
         """An iterator over formatted attribute values in the same order as the header."""
         return [self.format_value(value) for value in self.values()]
```
nh13, line 142: I think we rely on this API elsewhere, so lets keep it an iterator and instead change the actual return type from a List to a generator (which is an iterator).  I am not sure why mypy didn't find this sooner.

```suggestion
    def formatted_values(self) -> Iterator[str]:
        """An iterator over formatted attribute values in the same order as the header."""
        return (self.format_value(value) for value in self.values())
```

fgpyo/util/metric.py
```diff
@@ -226,7 +226,7 @@ def parse(cls, fields: List[str]) -> Any:
         return inspect.attr_from(cls=cls, kwargs=dict(zip(header, fields)), parsers=parsers)
 
     @classmethod
-    def write(cls, path: Path, *values: MetricType) -> None:
+    def write(cls, path: Path, *values: "Metric") -> None:
```
nh13, line 229: Why loose this type hint?  It only writes metrics of this type, so should it not be `MetricType`?

fgpyo/util/types.py
```diff
@@ -9,20 +10,19 @@
 from typing import TypeVar
 from typing import Union
 
-try:
+# `get_origin_type` is a method that gets the outer type (ex list in a List[str])
+# `get_arg_types` is a method that gets the inner type (ex str in a List[str])
```
nh13, line 14: thank-you for adding this explanation, I forget every time.

fgpyo/util/types.py
```diff
@@ -92,7 +87,7 @@ def make_enum_parser(enum: Type[EnumType]) -> partial:
     return partial(_make_enum_parser_worker, enum)
 
 
-def is_constructible_from_str(type_: T) -> bool:
+def is_constructible_from_str(type_: type) -> bool:
```
nh13, line 90: can we remove `T = TypeVar("T")` in this case?

fgpyo/util/types.py
```diff
@@ -122,15 +117,16 @@ def _make_union_parser_worker(
...
     """Worker function behind union parsing. Iterates through possible parsers for the union and
     returns the value produced by the first parser that works. Otherwise raises an error if none
     work"""
     # Need to do this in the case of type Optional[str], because otherwise it'll return the string
     # 'None' instead of the object None
     if _is_optional(union):
         try:
-            return none_parser(value)
+            none_parser(value)
+            return None
```
nh13, line 129: why is this changed?  Now it always returns `None`!

nh13 (changes requested):
I see a few regressions to discuss, happy to chat through live.

fgpyo/util/metric.py
nh13: Add the bound on `MetricType` to have `Metric`

fgpyo/util/types.py
nh13, line 129:
```suggestion
            # mypy does not like functions that return None always, so return None seperately
            assert none_parser(value) is None
            return None
```

nh13 (approved):
LGTM based on in-person review

PR #80, opened 2023-11-27
Feature/metric with dataclass or attr

TedBrookings:
Solving the issue was not particularly easy, but the real battle is with `mypy`. Reviewers should be aware that both [dataclasses](https://mypy.readthedocs.io/en/stable/additional_features.html#caveats-known-issues) and [attr](https://mypy.readthedocs.io/en/stable/additional_features.html#the-attrs-package) modules have weird enough issues with type hints that they basically lie; and `mypy` has special code to figure out what's going on. Thus I had to make use of `# type: ignore` in multiple locations in actual code, and had to completely exclude `test_metric.py` from even being checked.

Also in an offline conversation with @tfenne I dropped changing `attr` from a package requirement to a dev/test requirement, because several attr-decorated classes in the main fgpyo codebase use the `slots` and `kw_only` keywords, which are only supported in `dataclasses` for python versions >= 3.10.

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -386,7 +386,7 @@ class CigarOp(enum.Enum):
         return self == CigarOp.S or self == CigarOp.H
 
 
-@attr.s(frozen=True, slots=True)
+@attr.s(frozen=True, slots=True, auto_attribs=True)
 class CigarElement:
     """Represents an element in a Cigar
 
@@ -395,14 +395,13 @@ class CigarElement:
         - operator (CigarOp): the operator of the element
     """
 
-    length: int = attr.ib()
-    operator: CigarOp = attr.ib()
+    length: int
+    operator: CigarOp
 
-    @length.validator
-    def _validate_length(self, attribute: Any, value: int) -> None:
+    def __attrs_post_init__(self) -> None:
         """Validates the length attribute is greater than zero."""
-        if value <= 0:
-            raise ValueError(f"Cigar element must have a length > 0, found {value}")
+        if self.length <= 0:
+            raise ValueError(f"Cigar element must have a length > 0, found {self.length}")
 
     @property
     def length_on_query(self) -> int:
@@ -424,7 +423,7 @@ class CigarParsingException(Exception):
     pass
 
 
-@attr.s(frozen=True, slots=True)
+@attr.s(frozen=True, slots=True, auto_attribs=True)
 class Cigar:
     """Class representing a cigar string.
 
@@ -432,7 +431,7 @@ class Cigar:
         - elements (Tuple[CigarElement, ...]): zero or more cigar elements
     """
 
-    elements: Tuple[CigarElement, ...] = attr.ib(default=())
+    elements: Tuple[CigarElement, ...] = ()
 
     @classmethod
     def from_cigartuples(cls, cigartuples: Optional[List[Tuple[int, int]]]) -> "Cigar":
@@ -519,7 +518,7 @@ class Cigar:
         return sum([elem.length_on_target for elem in self.elements])
 
 
-@attr.s(auto_attribs=True, frozen=True)
+@attr.s(frozen=True, auto_attribs=True)
 class SupplementaryAlignment:
     """Stores a supplementary alignment record produced by BWA and stored in the SA SAM tag.
 
@@ -532,12 +531,12 @@ class SupplementaryAlignment:
         nm: the number of edits
     """
 
-    reference_name: str = attr.ib()
-    start: int = attr.ib()
-    is_forward: bool = attr.ib()
-    cigar: Cigar = attr.ib()
-    mapq: int = attr.ib()
-    nm: int = attr.ib()
+    reference_name: str
+    start: int
+    is_forward: bool
+    cigar: Cigar
+    mapq: int
+    nm: int
 
     def __str__(self) -> str:
         return ",".join(
@@ -639,7 +638,7 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
     r2.template_length = -insert_size
 
 
-@attr.s(auto_attribs=True, frozen=True)
+@attr.s(frozen=True, auto_attribs=True)
 class ReadEditInfo:
     """
     Counts various stats about how a read compares to a reference sequence.
@@ -728,7 +727,7 @@ def calculate_edit_info(
     )
 
 
-@attr.s(auto_attribs=True, frozen=True)
+@attr.s(frozen=True, auto_attribs=True)
 class Template:
     """A container for alignment records corresponding to a single sequenced template
     or insert.
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -1,28 +1,112 @@
-import functools
 import sys
+import types as python_types
 import typing
-from enum import Enum
-from functools import partial
-from pathlib import PurePath
 from typing import Any
-from typing import Callable
 from typing import Dict
+from typing import FrozenSet
 from typing import Iterable
 from typing import List
 from typing import Literal
-from typing import Optional
+from typing import Mapping
+from typing import Protocol
 from typing import Tuple
 from typing import Type
 from typing import Union
 
-if sys.version_info >= (3, 12):
+if sys.version_info >= (3, 10):
     from typing import TypeAlias
 else:
     from typing_extensions import TypeAlias
 
-import attr
+import dataclasses
+import functools
+from dataclasses import MISSING as DATACLASSES_MISSING
+from dataclasses import fields as get_dataclasses_fields
+from dataclasses import is_dataclass as is_dataclasses_class
+from enum import Enum
+from functools import partial
+from pathlib import PurePath
+from typing import TYPE_CHECKING
+from typing import Callable
+from typing import Optional
+from typing import TypeVar
 
 import fgpyo.util.types as types
+
+attr: Optional[python_types.ModuleType]
+MISSING: FrozenSet[Any]
+
+try:
+    import attr
+
+    _use_attr = True
+    from attr import fields as get_attr_fields
+    from attr import fields_dict as get_attr_fields_dict
+
+    Attribute: TypeAlias = attr.Attribute  # type: ignore[name-defined, no-redef]
+    # dataclasses and attr have internal tokens for missing values, join into a set so that we can
+    # check if a value is missing without knowing the type of backing class
+    MISSING = frozenset({DATACLASSES_MISSING, attr.NOTHING})
+except ImportError:  # pragma: no cover
+    _use_attr = False
+    attr = None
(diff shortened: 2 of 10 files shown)


TedBrookings:
I should add a few tests to cover a handful of helper functions I created in `util.inspect`. That should likely bring codecov back up to target (or close to it)

TedBrookings:
I tried to address codecov. There were three categories of issue
1. Lines that should be covered by tests. I added tests to cover them.
2. Lines that probably shouldn't be covered. e.g. try/catch blocks for `ImportError` when `attr` is missing. I could drop these since the issue goal is no longer to remove `attr`. It feels like a plausible _someday_ goal, and adds relatively little cost. I'm inclined to leave those lines unchecked, but am happy to remove them if that's the preference.
3. Lines that are in _test files_! Those should not be checked at all. I added a codecov.yml file to the project to exclude them.

The final math makes codecov unhappy. I'm inclined to leave things as they stand.

fgpyo/util/inspect.py
```diff
@@ -9,24 +9,85 @@
...
+    from attr import NOTHING as ATTR_NOTHING
+    from attr import fields as get_attr_fields
+    from attr import fields_dict as get_attr_fields_dict
+
+    Attribute = attr.Attribute
+
+    MISSING = {DATACLASSES_MISSING, ATTR_NOTHING}
+except ImportError:
+    _use_attr = False
+    attr = None
```
clintval, line 49: Could you document what this variable is for? It is in the public namespace and I don't see an equivalent of it in the `import attr` success branch.

fgpyo/util/inspect.py
```diff
@@ -9,24 +9,85 @@
...
+    ATTR_NOTHING = None
+    Attribute = TypeVar("Attribute", bound=object)  # type: ignore
+
+    def get_attr_fields(cls: type) -> Tuple[dataclasses.Field, ...]:  # type: ignore
+        return ()
+
+    def get_attr_fields_dict(cls: type) -> Dict[str, dataclasses.Field]:  # type: ignore
+        return {}
```
clintval, line 57: Docstrings on all public functions please.

fgpyo/util/inspect.py
```diff
@@ -9,24 +9,85 @@
...
+        __attrs_attrs__: Dict[str, Any]
+
+
+def is_attr_class(cls: type) -> bool:  # type: ignore
+    return hasattr(cls, "__attrs_attrs__")
+
+
+MISSING_OR_NONE = {*MISSING, None}
+DataclassesOrAttrClass = Union[DataclassesProtocol, AttrsInstance]
+FieldType: TypeAlias = Union[dataclasses.Field, attr.Attribute]
```
clintval, line 83: Above in the imports you define `Attribute`, why not use that here?

fgpyo/util/inspect.py
```diff
@@ -254,9 +315,36 @@ def dict_parse(dict_string: str) -> Dict[Any, Any]:
...
+    """
+    Get the fields dict from either a dataclasses or attr dataclass.
+
+    Combine results in case someone chooses to mix them through inheritance.
+    """
+    if not (is_dataclasses_class(cls) or is_attr_class(cls)):
+        raise TypeError("cls must a dataclasses or attr class")
+    return {
+        **(get_dataclasses_fields_dict(cls) if is_dataclasses_class(cls) else {}),
+        **(get_attr_fields_dict(cls) if is_attr_class(cls) else {}),  # type: ignore
```
clintval, line 328: Since we have such type trickery going on. Could you add a refinement to all type ignore statements in this PR? It can be done with the syntax `# type: ignore[...]` where `...` is the type checker exception kind.

fgpyo/util/inspect.py
```diff
@@ -254,9 +315,36 @@ def dict_parse(dict_string: str) -> Dict[Any, Any]:
...
+    }
+
+
+def get_fields(cls: Type[DataclassesOrAttrClass]) -> Tuple[FieldType, ...]:
+    if not (is_dataclasses_class(cls) or is_attr_class(cls)):
+        raise TypeError("cls must a dataclasses or attr class")
+    return (get_dataclasses_fields(cls) if is_dataclasses_class(cls) else ()) + (
+        get_attr_fields(cls) if is_attr_class(cls) else ()  # type: ignore
+    )
```
clintval, line 337: Should this return a set so no one relies on the order for anything meaningful?

fgpyo/util/inspect.py
```diff
@@ -254,9 +315,36 @@ def dict_parse(dict_string: str) -> Dict[Any, Any]:
...
+
+
+AttrFromType = TypeVar("AttrFromType")
```
clintval, line 340: Docs please.

fgpyo/util/metric.py
```diff
@@ -10,14 +10,25 @@
...
+
+   >>> from fgpyo.util.metric import Metric
+   >>> import dataclasses
+   >>> @dataclasses.dataclass(frozen=True)
+   ... class Person(Metric["Person"]):
+   ...     name: str
+   ...     age: int
+
+or using attr:
 .. code-block:: python
```
clintval, line 32: Does this format correctly?

fgpyo/util/metric.py
```diff
@@ -75,15 +86,15 @@
...
    ...     last: str
    ...     @classmethod
    ...     def parse(cls, value: str) -> "Name":
    ...          fields = value.split(" ")
    ...          return Name(first=fields[0], last=fields[1])
-   >>> @attr.s(auto_attribs=True, frozen=True)
+   >>> @dataclasses.dataclass(frozen=True)
    ... class Person(Metric["Person"]):
    ...     name: Name
    ...     age: int
```
clintval, line 100: Is it worth it to have docs showing what happens if you mix dataclasses and attr?

fgpyo/util/tests/test_inspect.py
```diff
@@ -1,16 +1,20 @@
 from typing import Optional
```
clintval, line 1: Should we add a test some of the functions work for classes that are both dataclass and attr?

clintval (approved):
Only documentation requested in places where I couldn't quite understand what some of the variables were necessary for unless I scanned elsewhere in the code. I'm assuming attrs is used elsewhere in fgpyo so we can't make it an install optional can we?

fgpyo/sam/__init__.py
```diff
@@ -394,14 +394,13 @@ class CigarElement:
...
     """
 
-    length: int = attr.ib()
-    operator: CigarOp = attr.ib()
+    length: int
+    operator: CigarOp
 
-    @length.validator
-    def _validate_length(self, attribute: Any, value: int) -> None:
+    def __post_init__(self) -> None:
```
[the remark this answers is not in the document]
TedBrookings: It was a typo, should have been `__attrs_post_init__`. Good catch, thanks! I added a test.

As for your top-level question, attrs is used extensively throughout fgpyo. The original mandate was to replace attrs usages with dataclasses, but as I mentioned in the top comment, this can't be done until we drop support for python < 3.10.

fgpyo/util/inspect.py
TedBrookings: I added documentation. This is the success branch, and there is actually a corresponding value on the import failure branch.

fgpyo/util/inspect.py
TedBrookings: I added them. In this case these functions will never be called, and are basically there for static checking. I added an additional comment explaining this too.

fgpyo/util/inspect.py
TedBrookings: Done

fgpyo/util/inspect.py
TedBrookings: Added. This a typevar allowing the `attr_from` function to create either an attr class or a dataclasses class, so I added documentation there as well.

fgpyo/util/metric.py
TedBrookings: I'm not sure how to check. Is it easy to do?

fgpyo/util/metric.py
TedBrookings: It turns out you literally can't mix them. The constructors don't work if you try. I added a short blurb in the docs to mention that, but no one will be able to make such an object and presumably will discover this in debugging.

fgpyo/util/metric.py
TedBrookings: Nevermind, figured it out and fixed it.

fgpyo/util/inspect.py
clintval: In the future, you can document variables like:

```python
AttrFromType = TypeVar("AttrFromType")
"""TypeVar to allow attr_from to be used with either an attr class or a dataclasses class."""
```

And they will be picked up by Sphinx and compiled into rich docs (e.g. the fgpyo docs website). It might worth going back through and swapping line comments to true public variable docstrings.

fgpyo/util/inspect.py
```diff
@@ -20,10 +16,86 @@
...
+    from attr import AttrsInstance
+else:
+
+    class AttrsInstance(Protocol):  # type: ignore[no-redef]
+        __attrs_attrs__: Dict[str, Any]
+
+
+def is_attr_class(cls: type) -> bool:  # type: ignore[arg-type]
+    """Return True if the class is an attr class, and False otherwise"""
+    return hasattr(cls, "__attrs_attrs__")
```
clintval, line 85: Is this unit tested?

fgpyo/util/inspect.py
```diff
@@ -20,10 +16,86 @@
...
+
+
+MISSING_OR_NONE = {*MISSING, None}
+DataclassesOrAttrClass = Union[DataclassesProtocol, AttrsInstance]
+FieldType: TypeAlias = Union[dataclasses.Field, Attribute]
```
clintval, line 90: I would still document these. It makes sense what they are, but a reader will wonder what they are meant to be used for. Document like this so Sphinx picks it up:

```python
AttrFromType = TypeVar("AttrFromType")
"""TypeVar to allow attr_from to be used with either an attr class or a dataclasses class."""
```

fgpyo/util/inspect.py
```diff
@@ -1,16 +1,12 @@
-import functools
```
clintval, line 1: It's too late to propose making `inspect.py` "private" (e.g. `_inspect.py`) but I'm wondering if you think any of the public API is too tempting or dangerous for users to accidentally import and use and rely on. If yes, then we could go and prefix lots of what you added with `_` to hide more things.

clintval (approved):
This looks like a masterclass in Python type trickery.

Thanks for your hard and careful work!

My only lingering concern is exposing more inspect code to the public API and then having users accidentally (or purposefully) import some of the code when we probably don't want that to happen. How do you feel about that? Do we have any easy options since the inspect namespace is already "public"?

nh13 (approved):
Approve based on conversation to add a little more docs to the test metric module and perhaps a type hint or two in inspect, but don't spend more than 15 minutes, and if it doesn't work it doesn't work.

PR #85, opened 2024-01-30
feat: add SupplementaryAlignment constructor from read

msto:
I thought it'd be helpful to abstract away the logic for validating the presence of an SA tag and type-checking, since that's going to be necessary most of the time `parse_sa_tag` is used in practice

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -581,7 +581,7 @@ class SupplementaryAlignment:
         """Returns a list of SupplementaryAlignments for a given pysam.AlignedSegment."""
         if read.has_tag("SA"):
             sa_tag: str = cast(str, read.get_tag("SA"))
-            return SupplementaryAlignment.parse_sa_tag(sa_tag)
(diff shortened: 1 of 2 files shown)


nh13 (approved):
LGTM, thank-you!

fgpyo/sam/__init__.py
```diff
@@ -575,6 +576,24 @@ def parse_sa_tag(tag: str) -> List["SupplementaryAlignment"]:
...
+        """
+        Construct a list of SupplementaryAlignments from the SA tag in a pysam.AlignedSegment.
+
+        Args:
+            read: An alignment. The presence of the "SA" tag is not required.
+
+        Returns:
+            A list of all SupplementaryAlignments present in the SA tag.
+            If the SA tag is not present, or it is empty, an empty list will be returned.
+        """
```
msto, line 590: @nh13 to follow up on our slack convo, I added some detail to the docstring to clarify the behavior when the SA tag is absent. 

lmk if you have any suggested edits to the docs before I merge

PR #86, opened 2024-01-30
chore: isort

msto:
isort insisted on fixing these when I ran `ci/check.sh`  - afaict I'm using the version specified in the lock file, 5.10.1


clintval:
This was probably my mistake. I was relying on poetry and the checks in the GitHub Actions workflow.

Looks like isort isn't wired into GitHub Actions (neither is that CI helper script)...

I'll open an issue to resolve that script vs command in the GH Action.

There are MyPy, Black, and isort plugins for pytest, so I think whoever tackles this should get the harness down to just:

```
poetry run pytest
```

And everything can be run underneath that command.

clintval (approved):

nh13 (approved):

PR #95, opened 2024-02-21
Nested call to `type` is not necessary

### LINKED ISSUE #94: `from_attr` returns a `none_parser` rather than throw `ParserNotFound`
The offending line: https://github.com/fulcrumgenomics/fgpyo/blob/main/fgpyo/util/inspect.py#L229

`type(type(None))` returns `<class type>`, and all (I think) classes are instances of `type`, so `isinstance(type_, type(type(None)))` always returns `True`.

jdidion:
Fixes the check for `NoneType`. `type(type(None))` returns `<class type>`, and so `isinstance(type_, type(type(None)))` always returns `True`. Instead, we want `type_ is type(None)`.

Closes #94

### DIFF
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -60,18 +60,24 @@ def split_at_given_level(
             decrease_in_depth += high_level_split.count(char)
         outer_depth_of_split += increase_in_depth - decrease_in_depth
 
-        assert outer_depth_of_split >= 0, "Unpaired depth character! Likely incorrect output"
+        assert (
+            outer_depth_of_split >= 0
+        ), "Unpaired depth character! Likely incorrect output"
 
         current_outer_splits.append(high_level_split)
         if outer_depth_of_split == 0:
             out_vals.append(split_delim.join(current_outer_splits))
             current_outer_splits = []
-    assert outer_depth_of_split == 0, "Unpaired depth character! Likely incorrect output!"
+    assert (
+        outer_depth_of_split == 0
+    ), "Unpaired depth character! Likely incorrect output!"
     return out_vals
 
 
 def _get_parser(
-    cls: Type, type_: TypeAlias, parsers: Optional[Dict[type, Callable[[str], Any]]] = None
+    cls: Type,
+    type_: TypeAlias,
+    parsers: Optional[Dict[type, Callable[[str], Any]]] = None,
 ) -> partial:
     """Attempts to find a parser for a provided type.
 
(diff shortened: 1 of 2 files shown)


fgpyo/util/inspect.py
```diff
@@ -226,7 +228,7 @@ def dict_parse(dict_string: str) -> Dict[Any, Any]:
                 return types.make_enum_parser(type_)
             elif types.is_constructible_from_str(type_):
                 return functools.partial(type_)
-            elif isinstance(type_, type(type(None))):
+            elif type_ is type(None): # type: ignore[E721]
```
jdidion, line 231: mypy complains about this check, but using `isinstance` here is incorrect, thus ignoring it.

jdidion:
I think the codecov failure is due to extra lines being added by black reformatting. Would appreciate if we can merge despite that failure.

fgpyo/util/inspect.py
```diff
@@ -226,7 +228,7 @@ def dict_parse(dict_string: str) -> Dict[Any, Any]:
                 return types.make_enum_parser(type_)
             elif types.is_constructible_from_str(type_):
                 return functools.partial(type_)
-            elif isinstance(type_, type(type(None))):
+            elif type_ is type(None):  # noqa: E721
```
nh13, line 231: can you please add a test?

nh13 (changes requested):
Thank-you for fixing all this.  Nonetheless, there's a lot of other fixes that are not related to this PR, though nice to have in general (again thank-you!).  I'd much prefer to see separate PRs for each (e.g. dropping 3.7 support, adding 3.11 support, reformatting TOML, fix black formatting, etc).  If that's too onerous, I would want to see individual commits for the above, to make it clear the intent of each set of changes.  Is it too onerous to do either at this point?

fgpyo/util/inspect.py
msto: @jdidion following up on your question from Slack, I think this is the right approach.

as far as I'm aware, flake8 guards against `type(x) == int` and the like. We're doing equality checks against types above (e.g. `if type_ == bool`) and flake8 doesn't complain, I'm guessing because we don't call `type()` in the same line. 

If you wanted to avoid the `noqa`, you could declare `NoneType = type(None)` above (or in `fgpyo.types` and import it), then replace this line with `elif type_ is NoneType`. But this feels like an edge case in flake8, so I think what you have is fine if you're fine with it

PR #96, opened 2024-02-22
Fix the check for `None` type in `inspect._get_parser`

### LINKED ISSUE #94: `from_attr` returns a `none_parser` rather than throw `ParserNotFound`

jdidion:
`type(type(None))` returns `<class type>`, and so `isinstance(type_, type(type(None)))` always returns `True`. Instead, we want `type_ is type(None)`. I used @msto suggestion to alias `NoneType = type(None)` to avoid the flake8 E721 error.

Closes #94.

### DIFF
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -70,6 +70,9 @@ def split_at_given_level(
     return out_vals
 
 
+NoneType = type(None)
+
+
 def _get_parser(
     cls: Type, type_: TypeAlias, parsers: Optional[Dict[type, Callable[[str], Any]]] = None
 ) -> partial:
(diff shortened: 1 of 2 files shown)


TedBrookings (approved):
The `types` and `inspect` modules are quite complex and the fact that this weird edge-case resulted in incorrect behavior suggests to me that many more tests are needed. But that's a separate issue. I think this PR correctly addresses the issue at hand.

PR #98, opened 2024-02-22
Drop support for python 3.7

jdidion:


fgpyo/vcf/tests/test_builder.py
```diff
@@ -81,9 +81,11 @@ def _get_random_variant_inputs(
...
             random_generator.randint(0, 100)
             if value_type == VcfFieldType.INTEGER
-            else random_generator.uniform(0, 1)
-            if value_type == VcfFieldType.FLOAT
-            else random_generator.choice(["Up", "Down"])
+            else (
+                random_generator.uniform(0, 1)
+                if value_type == VcfFieldType.FLOAT
+                else random_generator.choice(["Up", "Down"])
+            )
```
TedBrookings, line 88: I don't understand the need for this change. The original seems more pythonic to me.

TedBrookings (approved):
There is one change I don't like (style, not function). Otherwise looks good to me.

fgpyo/vcf/tests/test_builder.py
jdidion: I can try to revert the change and see if black complains.

TedBrookings: I feel like a series of v1 if c1 else v2 if c2 else c3 is pretty typical for python. But the version black made isn't some unreadable abomination. If you don't care either way I'd say try to revert it. If black doesn't let you, that's that. I approved.

jdidion:
I'm not sure why the codecov check is failing. I deleted lots of code that was not covered by the tests, so I would have expected coverage to go up not down.

TedBrookings:
I suspect it's some crazy math to do with the fraction of lines in the change that are tested? In any case, the 2nd codecov test (overall project coverage) increased. I think you've done your duty and should merge.

jdidion:
Awesome thanks!

PR #99, opened 2024-02-22
Support python 3.11 and 3.12 in CI, update black, flake8, and poetry versions

jdidion:
Includes reformatting based on updated black version.


fgpyo/vcf/tests/test_builder.py
```diff
@@ -81,9 +81,11 @@ def _get_random_variant_inputs(
...
             random_generator.randint(0, 100)
             if value_type == VcfFieldType.INTEGER
-            else random_generator.uniform(0, 1)
-            if value_type == VcfFieldType.FLOAT
-            else random_generator.choice(["Up", "Down"])
+            else (
+                random_generator.uniform(0, 1)
+                if value_type == VcfFieldType.FLOAT
+                else random_generator.choice(["Up", "Down"])
+            )
```
TedBrookings, line 88: Same section we talked about yesterday. I'm not sure if this is black saying "no really, I insist" or if this change just predated the revert on the previous PR. If it's easy and you don't mind reverting, I'd prefer it. Otherwise it's fine..

TedBrookings (approved):
Same issue as previous PR, but I'm not sure if it predates your revert. Do what you think is best. Otherwise looks good to me.

jdidion:
Previously it was because I was formatting locally using a newer version of black than was pinned in the lockfile, and the older version of black was ambivalent.

In this PR, the pinned version of black is updated, and it now has an Opinion about this (ie --check will fail).

msto (approved):
lgtm

PR #103, opened 2024-03-15
feat: Support reading Picard-style metrics files with comments above the header

msto:
It would be helpful if `Metric.read` could parse metrics files with comments above the header, such as those produced by Picard. I've updated `Metric.read` to use the `read_header()` method introduced in #124 , and also simplified the typing on `io.to_reader()` and `io.to_writer()` (per Clint's suggestion)

---

**note** I want to keep this separate from the `MetricWriter` feature branch, but it's dependent on the `Metric._read_header` method introduced in that branch.

I am waiting to merge until #123 is merged.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -214,7 +214,7 @@ class Metric(ABC, Generic[MetricType]):
         cls,
         path: Path,
         ignore_extra_fields: bool = True,
-        comment_prefix: str = "#",
+        comment_prefix: Optional[str] = None,
     ) -> Iterator["Metric[MetricType]"]:
         """Reads in zero or more metrics from the given path.
 
@@ -397,7 +397,7 @@ class Metric(ABC, Generic[MetricType]):
     def _read_header(
         reader: TextIOWrapper,
         delimiter: str = "\t",
-        comment_prefix: str = "#",
+        comment_prefix: Optional[str] = None,
     ) -> MetricFileHeader:
         """
         Read the header from an open file.
@@ -422,12 +422,17 @@ class Metric(ABC, Generic[MetricType]):
         preamble: List[str] = []
 
         for line in reader:
-            if line.strip().startswith(comment_prefix) or line.strip() == "":
-                # Skip any commented or empty lines before the header
-                preamble.append(line.strip())
+            line = line.strip()
+
+            if line == "":
+                # Skip any empty lines before the header
+                preamble.append(line)
+            elif comment_prefix is not None and line.startswith(comment_prefix):
+                # Skip any commented lines before the header
+                preamble.append(line)
             else:
                 # The first line with any other content is assumed to be the header
-                fieldnames = line.strip().split(delimiter)
+                fieldnames = line.split(delimiter)
                 break
         else:
             # If the file was empty, kick back an empty header
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -237,8 +237,8 @@ class Metric(ABC, Generic[MetricType]):
             try:
                 header = cls._read_header(reader, comment_prefix=comment_prefix)
                 fieldnames: List[str] = header.fieldnames
-            except ValueError:
-                raise ValueError(f"No header found in file: {path}")
+            except ValueError as e:
+                raise ValueError(f"No header found in file: {path}") from e
 
             # check the header
             class_fields = set(cls.header())
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -234,11 +234,11 @@ class Metric(ABC, Generic[MetricType]):
         """
         parsers = cls._parsers()
         with io.to_reader(path) as reader:
-            try:
-                header = cls._read_header(reader, comment_prefix=comment_prefix)
-                fieldnames: List[str] = header.fieldnames
-            except ValueError as e:
-                raise ValueError(f"No header found in file: {path}") from e
+            header = cls._read_header(reader, comment_prefix=comment_prefix)
+            fieldnames: List[str] = header.fieldnames
+
+            if not fieldnames:
+                raise ValueError(f"No header found in file: {path}")
 
             # check the header
             class_fields = set(cls.header())
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -210,7 +210,12 @@ class Metric(ABC, Generic[MetricType]):
         return {}
 
     @classmethod
-    def read(cls, path: Path, ignore_extra_fields: bool = True) -> Iterator[Any]:
+    def read(
+        cls,
+        path: Path,
+        ignore_extra_fields: bool = True,
+        comment_prefix: str = "#",
+    ) -> Iterator["Metric[MetricType]"]:
         """Reads in zero or more metrics from the given path.
 
         The metric file must contain a matching header.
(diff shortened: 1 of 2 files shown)


fgpyo/io/__init__.py
```diff
@@ -59,9 +59,19 @@
...
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
+
+
 COMPRESSED_FILE_EXTENSIONS: Set[str] = {".gz", ".bgz"}
 
 
+Reader: TypeAlias = Union[io.TextIOWrapper, TextIO, IO[Any]]
+Writer: TypeAlias = Union[IO[Any], io.TextIOWrapper]
```
msto, line 72: I added these type aliases to make it easier to reference the return types of `to_reader` and `to_writer` in other modules.

These were how the respective function signatures were defined, but is there any reason not to make them the same?

fgpyo/util/metric.py
```diff
@@ -322,3 +332,29 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+            A list of field names found in the header line.
+        """
+
+        for line in reader:
+            if not line.startswith(comment_char) and not line.strip() == "":
+                break
+        else:
+            raise ValueError("No header found")
+
+        return line.rstrip("\r\n").split("\t")
```
msto, line 371: I noticed `Metric` always assumes a tab-delimited file.

Would it be sensible to provide the delimiter as an argument to the `read()` and `write()` methods?

fgpyo/io/__init__.py
TedBrookings: Based on my understanding of the outputs, they should be `TextIO` or `io.TextIOWrapper`. The `IO[Any]` is superfluous.

TedBrookings (approved):
LGTM. I think you should wait to merge until Nils signs off. My general sense is that Metric at least *originated* as an internal tool for Fulcrum code, not a general-purpose TSV <-> data utility. It would then read and write in a way that Fulcrum approved of. Header comments and CSV instead of TSV were deliberately excluded. I personally don't see the harm in relaxing the allowed inputs while retaining strict outputs.

fgpyo/io/__init__.py
TedBrookings: To be more specific, I don't think it's helpful to have overly broad types for _output_. It should be as specific as possible to help type-checking.

msto:
That's fair. 

I would be open to refactoring so `.read()` remains specific to `Metric`-formatted files. In that case I'd suggest providing a `.read_picard()` method or similar, which would filter comment lines before calling `.read()` under the hood.

I do think it would be generally valuable to have built-in support for reading Picard metrics into a `Metric`, and it would be very specifically helpful for one of my clients.

msto:
@nh13 thoughts?

nh13:
@msto I'd like to review this one before it gets merged

clintval (commented):
I'm indifferent on the PR because for Picard files I use this library:

- https://pypi.org/project/crimson/

It's not typed but it's been good enough for me since it doesn't force a co-install of things like pysam.

fgpyo/util/metric.py
```diff
@@ -149,7 +152,12 @@ def _parsers(cls) -> Dict[type, Callable[[str], Any]]:
         return {}
 
     @classmethod
-    def read(cls, path: Path, ignore_extra_fields: bool = True) -> Iterator[Any]:
+    def read(
+        cls,
+        path: Path,
+        ignore_extra_fields: bool = True,
+        header_comment_char: str = DEFAULT_HEADER_COMMENT_CHAR,
```
tfenne, line 159: A couple of thoughts:

1. I think this should probably be `: Optional[str] = None` for backwards compatibility
2. Should we also allow for blank lines before the header?
3. Should lines starting with the comment character (and blank lines?) be allowed in the body of the data or only pre-header?

fgpyo/util/metric.py
```diff
@@ -322,3 +338,34 @@ def fast_concat(*inputs: Path, output: Path) -> None:
             io.write_lines(
                 path=output, lines_to_write=list(io.read_lines(input_path))[1:], append=True
             )
+
+    @staticmethod
+    def _read_header(
+        reader: io.Reader,
+        comment_char: str = DEFAULT_HEADER_COMMENT_CHAR,
```
tfenne, line 345: I would prefer to remove defaults from non-public methods to make sure that the values are being vectored appropriately.

fgpyo/util/metric.py
tfenne: And one more ... since python has no `char` type, maybe this should be `comment_prefix`?  And then it could support e.g. `##` or `//`.

fgpyo/util/metric.py
tfenne: I wouldn't object to it, so long as the default is always `tab`.  However, that does raise the question of whether we should then do better at quoting/escaping since tabs don't often show up in values, but commas and spaces will.

tfenne (changes requested):
Some reasonably minor suggestions.

fgpyo/util/metric.py
msto: 1. Yep, agreed
2. This PR does allow for blank lines before the header - see line 366. Any lines that start with the comment character or are blank are skipped.
3. I don't think so. I'm not familiar with any format where that's valid, and I think it's more likely to be a sign of a malformatted file or a concatenation gone wrong.
4. Yep, agreed

I'll update (1) and (4)

fgpyo/util/metric.py
msto: 👍

fgpyo/util/metric.py
msto: Good thought - I'll open a separate issue to address custom delimiters

nh13 (changes requested):
Minor comments from me, but a few others to fixup from others.

fgpyo/util/metric.py
msto: See #124

fgpyo/util/metric.py
```diff
@@ -192,7 +192,12 @@ def _parsers(cls) -> Dict[type, Callable[[str], Any]]:
         return {}
 
     @classmethod
-    def read(cls, path: Path, ignore_extra_fields: bool = True) -> Iterator["Metric[MetricType]"]:
+    def read(
+        cls,
+        path: Path,
+        ignore_extra_fields: bool = True,
+        comment_prefix: str = "#",
```
nh13, line 199: Is this backwards incompatible?  Should we bump the minor version?

fgpyo/util/metric.py
```diff
@@ -203,13 +208,23 @@ def read(cls, path: Path, ignore_extra_fields: bool = True) -> Iterator["Metric[
...
+            comment_prefix: Any lines beginning with this string will be ignored before parsing the
+                header.
+
+        Raises:
+            ValueError: if the file does not contain a header.
         """
         parsers = cls._parsers()
         with io.to_reader(path) as reader:
-            header: List[str] = reader.readline().rstrip("\r\n").split("\t")
+            try:
```
nh13, line 219:
```suggestion
            fieldnames: List[str]
            try:
```

nh13 (approved):
LGTM, take it or leave the additional comments/suggestions.

fgpyo/util/metric.py
msto: Good question, it isn't.

One possible solution for backwards compatibility - I could update `Metric._read_header` to take an `Optional` comment prefix, and return the first line if the prefix is `None` (i.e. no skipping of prefixed _or_ blank lines)

Is that something you'd like to see, or should I leave as is and bump the version?

msto:
I want to keep this separate from the `MetricWriter` feature branch, but it's dependent on the `Metric._read_header` method introduced in that branch.

This PR is done but I am waiting to merge until #123 is merged.

fgpyo/util/metric.py
tfenne: I might suggest `comment_symbol`?

Also i agree with @nh13 that ideally the method gets expanded to include both:

```python
comment_symbol: Optional[str] = None,
skip_blank_lines: bool = False
```

I might even argue for `skip_blank_lines` to even be a regex that defines lines to skip ... so e.g. I could use `skip_lines_matching = "$\s*^"` to skip stupid lines that are all tabs/spaces but no content!

tfenne (approved):
I think this looks fine, but I'm curious why change the default comment prefix, and remove the `strip()` when parsing the header?

Would this require classes that read a Picard metrics file to use all upper-case attribute names, or is case-translation managed somehow?

PR #105, opened 2024-04-17
feat: Add end property on SupplementaryAlignment

msto:
I have a context where it'd be helpful if the `end` were readily accessible.


tfenne (approved):
LGTM

PR #107, opened 2024-05-05
feat: Add MetricWriter class

### LINKED ISSUE #88: Add MetricWriter to permit streamed writing of metrics
`Metric.write()` does not support appending rows to an existing file, it only supports batch writing

It'd be helpful if there were an `.append()` or `.write_one()` function. 

Nils requested this be made available via a new `MetricWriter` class, to more obviously couple the header write and the write of individual metrics, e.g.

https://github.com/fulcrumgenomics/fgbio/blob/0a03aa9b19def665b92ad8887e743332daa3bceb/src/main/scala/com/fulcrumgenomics/util/Metric.scala#L68

msto:
Closes #88.

I've adapted the `DataclassWriter` from https://github.com/msto/dataclass_io/ to work with Metrics.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -119,19 +119,25 @@ Formatting and parsing the values for custom types is supported by overriding th
 import dataclasses
 import sys
 from abc import ABC
+from contextlib import AbstractContextManager
+from csv import DictWriter
 from dataclasses import dataclass
 from enum import Enum
 from inspect import isclass
 from io import TextIOWrapper
 from pathlib import Path
+from types import TracebackType
 from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import Generic
+from typing import Iterable
 from typing import Iterator
 from typing import List
+from typing import Optional
 from typing import Type
 from typing import TypeVar
+from typing import Union
 
 if sys.version_info >= (3, 10):
     from typing import TypeGuard
@@ -479,6 +485,140 @@ def asdict(metric: Metric) -> Dict[str, Any]:
         )
 
 
+class MetricWriter(Generic[MetricType], AbstractContextManager):
+    _metric_class: Type[Metric]
+    _fieldnames: List[str]
+    _fout: TextIOWrapper
+    _writer: DictWriter
+
+    def __init__(
+        self,
+        filename: Union[Path, str],
+        metric_class: Type[Metric],
+        append: bool = False,
+        delimiter: str = "\t",
+        include_fields: Optional[List[str]] = None,
+        exclude_fields: Optional[List[str]] = None,
+    ) -> None:
+        """
+        Args:
+            path: Path to the file to write.
+            metric_class: Metric class.
+            append: If `True`, the file will be appended to. Otherwise, the specified file will be
+                overwritten.
+            delimiter: The output file delimiter.
+            include_fields: If specified, only the listed fieldnames will be included when writing
+                records to file. Fields will be written in the order provided.
+                May not be used together with `exclude_fields`.
+            exclude_fields: If specified, any listed fieldnames will be excluded when writing
+                records to file.
+                May not be used together with `include_fields`.
+
+        Raises:
+            TypeError: If the provided metric class is not a dataclass- or attr-decorated
+                subclass of `Metric`.
+            AssertionError: If the provided filepath is not writable.
+            AssertionError: If `append=True` and the provided file is not readable. (When appending,
+                we check to ensure that the header matches the specified metric class. The file must
+                be readable to get the header.)
+            ValueError: If `append=True` and the provided file does not include a header.
+            ValueError: If `append=True` and the header of the provided file does not match the
+                specified metric class and the specified include/exclude fields.
+        """
+
+        filepath: Path = Path(filename)
+        ordered_fieldnames: List[str] = _validate_output_fieldnames(
+            metric_class=metric_class,
+            include_fields=include_fields,
+            exclude_fields=exclude_fields,
+        )
+
+        _assert_is_metric_class(metric_class)
+        io.assert_path_is_writeable(filepath)
+        if append:
+            io.assert_path_is_readable(filepath)
+            _assert_file_header_matches_metric(
+                path=filepath,
+                metric_class=metric_class,
+                ordered_fieldnames=ordered_fieldnames,
+                delimiter=delimiter,
+            )
+
+        self._metric_class = metric_class
+        self._fieldnames = ordered_fieldnames
+        self._fout = io.to_writer(filepath, append=append)
+        self._writer = DictWriter(
+            f=self._fout,
+            fieldnames=self._fieldnames,
+            delimiter=delimiter,
+        )
+
+        # If we aren't appending to an existing file, write the header before any rows
+        if not append:
+            self._writer.writeheader()
+
+    def __enter__(self) -> "MetricWriter":
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_value: BaseException,
+        traceback: TracebackType,
+    ) -> None:
+        self.close()
+        super().__exit__(exc_type, exc_value, traceback)
+
+    def close(self) -> None:
+        """Close the underlying file handle."""
+        self._fout.close()
+
+    def write(self, metric: Metric) -> None:
+        """
+        Write a single Metric instance to file.
+
+        The Metric is converted to a dictionary and then written using the underlying
+        `csv.DictWriter`. If the `MetricWriter` was created using the `include_fields` or
+        `exclude_fields` arguments, the fields of the Metric are subset and/or reordered
+        accordingly before writing.
+
+        Args:
+            metric: An instance of the specified Metric.
+
+        Raises:
+            TypeError: If the provided `metric` is not an instance of the Metric class used to
+            parametrize the writer.
+        """
+        if not isinstance(metric, self._metric_class):
+            raise TypeError(f"Must provide instances of {self._metric_class.__name__}")
+
+        # Serialize the Metric to a dict for writing by the underlying `DictWriter`
+        row = asdict(metric)
+
+        # Filter and/or re-order output fields if necessary
+        row = {fieldname: row[fieldname] for fieldname in self._fieldnames}
+
+        # Format values
+        row = {fieldname: self._metric_class.format_value(val) for fieldname, val in row.items()}
+
+        self._writer.writerow(row)
+
+    def writeall(self, metrics: Iterable[Metric]) -> None:
+        """
+        Write multiple Metric instances to file.
+
+        Each Metric is converted to a dictionary and then written using the underlying
+        `csv.DictWriter`. If the `MetricWriter` was created using the `include_fields` or
+        `exclude_fields` arguments, the attributes of each Metric are subset and/or reordered
+        accordingly before writing.
+
+        Args:
+            metrics: A sequence of instances of the specified Metric.
+        """
+        for metric in metrics:
+            self.write(metric)
+
+
 def _get_fieldnames(metric_class: Type[Metric]) -> List[str]:
     """
     Get the fieldnames of the specified metric class.
@@ -499,10 +639,47 @@ def _get_fieldnames(metric_class: Type[Metric]) -> List[str]:
         assert False, "Unreachable"
 
 
+def _validate_output_fieldnames(
+    metric_class: Type[MetricType],
+    include_fields: Optional[List[str]] = None,
+    exclude_fields: Optional[List[str]] = None,
+) -> List[str]:
+    """
+    Subset and/or re-order the Metric's fieldnames based on the specified include/exclude lists.
+
+    * Only one of `include_fields` and `exclude_fields` may be specified.
+    * All fieldnames specified in `include_fields` must be fields on `metric_class`. If this
+      argument is specified, fields will be returned in the order they appear in the list.
+    * All fieldnames specified in `exclude_fields` must be fields on `metric_class`. (This is
+      technically unnecessary, but is a safeguard against passing an incorrect list.)
+    * If neither `include_fields` or `exclude_fields` are specified, return the `metric_class`'s
+      fieldnames, in the order they are defined on the `metric_class`.
+
+    Raises:
+        ValueError: If both `include_fields` and `exclude_fields` are specified.
+    """
+
+    if include_fields is not None and exclude_fields is not None:
+        raise ValueError(
+            "Only one of `include_fields` and `exclude_fields` may be specified, not both."
+        )
+    elif exclude_fields is not None:
+        _assert_fieldnames_are_metric_attributes(exclude_fields, metric_class)
+        output_fieldnames = [f for f in _get_fieldnames(metric_class) if f not in exclude_fields]
+    elif include_fields is not None:
+        _assert_fieldnames_are_metric_attributes(include_fields, metric_class)
+        output_fieldnames = include_fields
+    else:
+        output_fieldnames = _get_fieldnames(metric_class)
+
+    return output_fieldnames
+
+
 def _assert_file_header_matches_metric(
     path: Path,
     metric_class: Type[MetricType],
     delimiter: str,
+    ordered_fieldnames: Optional[List[str]] = None,
 ) -> None:
     """
     Check that the specified file has a header and its fields match those of the provided Metric.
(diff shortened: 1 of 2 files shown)


fgpyo/io/__init__.py
```diff
@@ -56,11 +56,15 @@
...
 from typing import Set
 from typing import TextIO
+from typing import TypeAlias
 from typing import Union
 from typing import cast
 
 COMPRESSED_FILE_EXTENSIONS: Set[str] = {".gz", ".bgz"}
 
+ReadableFileHandle: TypeAlias = Union[io.TextIOWrapper, TextIO, IO[Any]]
+WritableFileHandle: TypeAlias = Union[IO[Any], io.TextIOWrapper]
```
clintval, line 66: I _**think**_ you only need `io.TextIOWrapper` https://github.com/fulcrumgenomics/fgpyo/pull/103#discussion_r1539758711

But I also recently found these structural types in typeshed too:

https://github.com/python/typeshed/blob/2d33fe212221a05661c0db5215a91cf3d7b7f072/stdlib/_typeshed/__init__.pyi#L253-L271

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+
+
+def is_metric(cls: Any) -> bool:
+    """True if the given class is a Metric."""
+
+    return (
+        isclass(cls)
+        and issubclass(cls, Metric)
+        and (dataclasses.is_dataclass(cls) or attr.has(cls))
+    )
```
clintval, line 356: Return type can be a `TypeGuard`

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+@dataclass(kw_only=True)
+class MetricFileFormat:
+    """
+    Parameters describing the format and configuration of a delimited Metric file.
+
+    Most of these parameters, if specified, are passed through to `csv.DictReader`/`csv.DictWriter`.
+    """
+
+    delimiter: str = "\t"
+    comment: str = "#"
```
clintval, line 368: Should we anticipate multiple comment prefixes?

```suggestion
    comment_prefixes: set[str] = {"#"}
```

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+def asdict(metric: Metric) -> dict[str, Any]:
+    """Convert a Metric instance to a dictionary."""
+    assert_is_metric(type(metric))
+
+    if dataclasses.is_dataclass(metric):
+        return dataclasses.asdict(metric)
+    elif attr.has(metric):
+        return attr.asdict(metric)
+    else:
+        assert False, "Unreachable"
```
clintval, line 397: For your unreachables, I'm wondering if you could use this:

https://docs.python.org/3/library/typing.html#typing.assert_never

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+                May not be used together with `include_fields`.
+
+        Raises:
+            AssertionError: If the provided metric class is not a dataclass- or attr-decorated
+                subclass of `Metric`.
+            AssertionError: If the provided filepath is not writable. (Or readable, if
+                `append=True`.)
+        """
+
+        filepath: Path = filename if isinstance(filename, Path) else Path(filename)
```
clintval, line 460: Path will return itself if it is already a path:

```suggestion
        filepath: Path = Path(filename)
```

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+
+    if dataclasses.is_dataclass(metric_class):
+        return [f.name for f in dataclasses.fields(metric_class)]
+    elif attr.has(metric_class):
+        return [f.name for f in attr.fields(metric_class)]
+    else:
+        assert False, "Unreachable"
+
+
+class MetricWriter:
```
clintval, line 423: Do all our Python versions support generic typing? If so:

```suggestion
class MetricWriter(Generic[MetricType]):
```

Can also add context manager type:

```suggestion
class MetricWriter(Generic[MetricType], ContextManager):
```

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+    def __enter__(self) -> "MetricWriter":
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Type[BaseException],
+        exc_value: BaseException,
+        traceback: TracebackType,
+    ) -> None:
+        self.close()
```
clintval, line 495:
```suggestion
        self.close()
        return super().__exit__(exc_type, exc_value, traceback)
```

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+            metrics: A sequence of instances of the specified Metric.
+        """
+        for metric in metrics:
+            self.write(metric)
+
+
+def _get_header(
+    reader: io.ReadableFileHandle,
+    file_format: MetricFileFormat,
+) -> Optional[MetricFileHeader]:
```
clintval, line 543: If/when we implement a `MetricReader` in a similar style, you might consider refactoring to re-use some of the reading logic. I can imagine a `MetricReader` would wrap a readable file handle in a peekable iterator so you could consume the header but still be able to use the iterator with the header included for actual reading (maybe with a `DictReader`?)

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        cls: A class object.
+
+    Raises:
+        TypeError: If the given class is not a Metric.
+    """
+    if not is_metric(cls):
+        raise TypeError(f"Not a dataclass or attr decorated Metric: {cls}")
+
+
+def assert_file_header_matches_metric(
```
clintval, line 632: These functions and those further down could be private (with a `_` prefix).

clintval (commented):
I like the direction this is taking! Could you add some examples of how to use this in the `__doc__` and check the docs compile correctly and look good too?

fgpyo/io/__init__.py
nh13: 1. It'd be nice to have comment or two about what each of these types are trying to alias
2. Why not just `Readable` and `Writeable`?

fgpyo/util/metric.py
nh13: TIL

fgpyo/util/metric.py
nh13: 👍

fgpyo/util/metric.py
nh13: TIL

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        accordingly before writing.
+
+        Args:
+            metric: An instance of the specified Metric.
+        """
+        if not isinstance(metric, self._metric_class):
+            raise ValueError(f"Must provide instances of {self._metric_class.__name__}")
+
+        # Serialize the Metric to a dict for writing by the underlying `DictWriter`
+        row = asdict(metric)
```
nh13, line 517: This bypasses the `format_value` method on `Metric`, which we'll want to use.

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+    """
+    Read the header from an open file.
+
+    The first row after any commented or empty lines will be used as the fieldnames.
+
+    Lines preceding the fieldnames will be returned in the `preface.`
+
+    NB: This function returns `Optional` instead of raising an error because the name of the
+    source file is not in scope, making it difficult to provide a helpful error message. It is
+    the responsibility of the caller to raise an error if the file is empty.
```
nh13, line 553: I'd probably raise an error (it's more pythonic) since the caller still needs to check for `None` in this case and handle it.   If an exception is raised, then one could still use exception chaining (`raise Exception from ex`) in a try block.

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        None if the file was empty or contained only comments or empty lines.
+    """
+
+    preface: list[str] = []
+
+    for line in reader:
+        if line.startswith(file_format.comment) or line.strip() == "":
+            preface.append(line.strip())
+        else:
+            break
```
nh13, line 573: I saved a line (feel free to ignore)
```suggestion
        if not line.startswith(file_format.comment) and line.strip() != "":
            break
        preface.append(line.strip())
```

fgpyo/util/metric.py
```diff
@@ -334,3 +344,330 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+    file_format: MetricFileFormat,
+) -> None:
+    """
+    Check that the specified file has a header and its fields match those of the provided Metric.
+    """
+    with path.open("r") as fin:
+        header: MetricFileHeader = _get_header(fin, file_format=file_format)
+
+    if header is None:
+        raise ValueError(f"Could not find a header in the provided file: {path}")
```
nh13, line 644: here's where exception chaining could occur

nh13 (changes requested):
Minor comments

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
             "The provided metric is not an instance of a `dataclass` or `attr.s`-decorated Metric "
             f"class: {metric.__class__}"
         )
+
+
+class MetricWriter(Generic[MetricType], AbstractContextManager):
```
msto, line 486: TODO docs

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+        """
+        Write a single Metric instance to file.
+
+        The Metric is converted to a dictionary and then written using the underlying
+        `csv.DictWriter`. If the `MetricWriter` was created using the `include_fields` or
+        `exclude_fields` arguments, the attributes of the dataclass are subset and/or reordered
+        accordingly before writing.
+
+        Args:
+            metric: An instance of the specified Metric.
```
msto, line 574:
```suggestion
        Args:
            metric: An instance of the specified Metric.
            
        Raises:
            TypeError: If the provided `metric` is not an instance of the Metric class used to parametrize the writer.
```

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+        """
+        Write multiple Metric instances to file.
+
+        Each Metric is converted to a dictionary and then written using the underlying
+        `csv.DictWriter`. If the `MetricWriter` was created using the `include_fields` or
+        `exclude_fields` arguments, the attributes of each Metric are subset and/or reordered
+        accordingly before writing.
+
+        Args:
+            metrics: A sequence of instances of the specified Metric.
```
msto, line 600:
```suggestion
            metrics: A sequence of instances of the specified Metric.
        
        Raises:
            TypeError: If any of the provided `metrics` are not an instance of the Metric class used to parametrize the writer.
```

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+
+
+def assert_file_header_matches_metric(
+    path: Path,
+    metric_class: type[Metric],
+    delimiter: str,
+) -> None:
+    """
+    Check that the specified file has a header and its fields match those of the provided Metric.
+    """
```
msto, line 636: TODO make private and add `Raises` block

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+        raise ValueError(
+            "The provided file does not have the same field names as the provided dataclass:\n"
+            f"\tDataclass: {metric_class.__name__}\n"
+            f"\tFile: {path}\n"
+            f"\tDataclass fields: {', '.join(get_fieldnames(metric_class))}\n"
+            f"\tFile: {', '.join(header.fieldnames)}\n"
+        )
+
+
+def assert_fieldnames_are_metric_attributes(
```
msto, line 653: TODO make private

fgpyo/util/metric.py
```diff
@@ -466,3 +481,240 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    elif include_fields is not None:
+        assert_fieldnames_are_metric_attributes(include_fields, metric_class)
+        output_fieldnames = include_fields
+    else:
+        output_fieldnames = get_fieldnames(metric_class)
+
+    return output_fieldnames
+
+
+def assert_is_metric_class(cls: type[Metric]) -> None:
```
msto, line 709:
```suggestion
def _assert_is_metric_class(cls: type[Metric]) -> None:
```

fgpyo/util/metric.py
msto:
```suggestion
def _assert_fieldnames_are_metric_attributes(
```

fgpyo/util/metric.py
msto: I think this is unlikely to occur in practice and makes the interface clunkier.

Could you provide an example of a file format that would have multiple prefixes? I can't think of one off the top of my head.

fgpyo/util/metric.py
msto: You won't need a peekable iterator, the handle will be left open to the first row of data, and metrics can be consumed after calling `_read_header`

proof of concept: https://github.com/msto/dataclass_io/blob/c3d1d622788e58c555efb2e266aa3622a95a6786/dataclass_io/reader.py#L57-L62

fgpyo/util/metric.py
```diff
@@ -517,10 +694,12 @@ def _assert_file_header_matches_metric(
         path: A path to a `Metric` file.
         metric_class: The `Metric` class to validate against.
         delimiter: The delimiter to use when reading the header.
+        ordered_fieldnames: An optional ordering of the fieldnames in the header.
 
     Raises:
         ValueError: If the provided file does not include a header.
-        ValueError: If the header of the provided file does not match the provided Metric.
+        ValueError: If the header of the provided file does not match the provided Metric (or list
+            of ordered fieldnames, if provided).
```
nh13, line 702: 1. or -> and
2. where does `ordered_fieldnames` get checked?

fgpyo/util/metric.py
```diff
@@ -502,10 +642,47 @@ def _get_fieldnames(metric_class: Type[Metric]) -> List[str]:
...
+    exclude_fields: Optional[List[str]] = None,
+) -> List[str]:
+    """
+    Subset and/or re-order the Metric's fieldnames based on the specified include/exclude lists.
+
+    * Only one of `include_fields` and `exclude_fields` may be specified.
+    * All fieldnames specified in `include_fields` must be fields on `metric_class`. If this
+      argument is specified, fields will be returned in the order they appear in the list.
+    * All fieldnames specified in `exclude_fields` must be fields on `metric_class`. (This is
+      technically unnecessary, but is a safeguard against passing an incorrect list.)
```
nh13, line 657: do we strongly about this safeguard?

nh13 (changes requested):

fgpyo/util/metric.py
clintval: UCSC BED files come to mind, but they also lack a header with named fields entirely. This is one of those API decisions that's hard to go back on since it will be a breaking change in the future, but not now.

clintval: If I'm understanding the API correctly, this wouldn't only activate for mixed comment prefixes in the same file, but allow for setting multiple comment prefixes across files too?

fgpyo/util/metric.py
```diff
@@ -334,3 +372,390 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        traceback: TracebackType,
+    ) -> None:
+        self.close()
+        super().__exit__(exc_type, exc_value, traceback)
+
+    def close(self) -> None:
+        """Close the underlying file handle."""
+        self._fout.close()
+
+    def write(self, metric: Metric) -> None:
```
nh13, line 576:
```suggestion
    def write(self, *metric: Metric) -> None:
```
why not make this `varargs`?  Can the return be a `MetricType`?  For both, see: https://github.com/fulcrumgenomics/fgpyo/blob/cd554be7f376ed30504ec2c7fafc041012732b0d/fgpyo/util/metric.py#L243

fgpyo/util/metric.py
```diff
@@ -334,3 +372,390 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+
+        # Filter and/or re-order output fields if necessary
+        row = {fieldname: row[fieldname] for fieldname in self._fieldnames}
+
+        # Format values
+        row = {fieldname: self._metric_class.format_value(val) for fieldname, val in row.items()}
+
+        self._writer.writerow(row)
+
+    def writeall(self, metrics: Iterable[Metric]) -> None:
```
nh13, line 606: Is this needed if we change `write`?

fgpyo/util/metric.py
nh13: bump

PR #108, opened 2024-05-06
fix: DataclassInstance and AttrInstance protocols

msto:
This PR includes two minor QoL fixes

- I found the aliased import of `DataclassInstance` to `DataclassesProtocol` unnecessary, especially since it conflicted with the naming convention of `AttrInstance`
- I updated the `*Instance` definitions to reflect the canonical definitions for each. (In particular, the implementation for `AttrInstance` was bugged since `__attrs_attrs__` can be a tuple of attributes, not a dict)


TedBrookings (approved):
Looks good to me. Good catch on the `__attrs_attrs__` definition.

nh13 (approved):
Thank-you! LGTM

PR #109, opened 2024-05-06
feat: Add asdict()

msto:
I'd like to add a function to convert a `Metric` instance to a dict. (In support of #107 )

I'm currently receiving the following mypy errors, which I suspect are due to a bug in how `Metric` is typed - as mypy appears to be inferring that `Metric` is a union of `DataclassInstance` and `type[DataclassInstance]`, when it should only be the former.

```
fgpyo/util/metric.py:346: error: Argument 1 to "asdict" has incompatible type "DataclassInstance | type[DataclassInstance]"; expected "DataclassInstance"  [arg-type]
fgpyo/util/metric.py:347: error: Argument 1 to "has" has incompatible type "Metric[Any]"; expected "type"  [arg-type]
fgpyo/util/metric.py:348: error: Argument 1 to "asdict" has incompatible type "type[AttrsInstance]"; expected "AttrsInstance"  [arg-type]
fgpyo/util/metric.py:348: note: ClassVar protocol member AttrsInstance.__attrs_attrs__ can never be matched by a class object
Found 3 errors in 1 file (checked 46 source files)
Failed Type Checking: [mypy -p fgpyo --config /Users/msto/code/fulcrumgenomics/fgpyo/ci/mypy.ini]
```

NB: it may be valuable to permit formatting/casting the values as part of this function (e.g. by adding a parameter `format: bool = False`), but to do so we'll probably want to extract `Metric.format_value()` to a standalone function instead of a classmethod

---

Edit: copying my explanation of the updated solution from Slack:

For the curious - the issue was two-fold, and due to my own incorrect typing.

`dataclasses.is_dataclass` accepts either an instance or a class object, and has a `TypeGuard` to narrow the type of the argument to `DataclassInstance | type[DataclassInstance]`. I had to add a helper function to override this guard and narrow the type further to `DataclassInstance`.

Meanwhile, `attr.has` **only** accepts a class object. I was passing an instance, which was the source of one type error. Fixing this by calling `attr.has(metric.__class__` was insufficient, because this did not narrow the type of the metric instance, so I added a similar helper for `AttrsInstance`.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -382,7 +382,7 @@ def is_attrs_instance(metric: Metric) -> TypeGuard[AttrsInstance]:
     return not isclass(metric) and attr.has(metric.__class__)
 
 
-def asdict(metric: Metric) -> dict[str, Any]:
+def asdict(metric: Metric) -> Dict[str, Any]:
     """
     Convert a Metric instance to a dictionary.
 
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -117,8 +117,10 @@ Formatting and parsing the values for custom types is supported by overriding th
 """
 
 import dataclasses
+import sys
 from abc import ABC
 from enum import Enum
+from inspect import isclass
 from pathlib import Path
 from typing import Any
 from typing import Callable
@@ -128,10 +130,17 @@ from typing import Iterator
 from typing import List
 from typing import TypeVar
 
+if sys.version_info >= (3, 10):
+    from typing import TypeGuard
+else:
+    from typing_extensions import TypeGuard
+
 import attr
 
 from fgpyo import io
 from fgpyo.util import inspect
+from fgpyo.util.inspect import AttrsInstance
+from fgpyo.util.inspect import DataclassInstance
 
 MetricType = TypeVar("MetricType", bound="Metric")
 
@@ -339,12 +348,53 @@ class Metric(ABC, Generic[MetricType]):
             )
 
 
-def asdict(metric: Metric) -> dict[str, Any]:
-    """Convert a Metric instance to a dictionary."""
+def is_dataclass_instance(metric: Metric) -> TypeGuard[DataclassInstance]:
+    """
+    Test if the given metric is a dataclass instance.
 
-    if dataclasses.is_dataclass(metric):
+    NB: `dataclasses.is_dataclass` returns True for both dataclass instances and class objects, and
+    we need to override the built-in function's `TypeGuard`.
+
+    Args:
+        metric: An instance of a Metric.
+
+    Returns:
+        True if the given metric is an instance of a dataclass-decorated Metric.
+        False otherwise.
+    """
+    return not isclass(metric) and dataclasses.is_dataclass(metric)
+
+
+def is_attrs_instance(metric: Metric) -> TypeGuard[AttrsInstance]:
+    """
+    Test if the given metric is an attr.s instance.
+
+    NB: `attr.has` provides a type guard, but only on the class object - we want to narrow the type
+    of the metric instance, so we implement a guard here.
+
+    Args:
+        metric: An instance of a Metric.
+
+    Returns:
+        True if the given metric is an instance of an attr.s-decorated Metric.
+        False otherwise.
+    """
+    return not isclass(metric) and attr.has(metric.__class__)
+
+
+def asdict(metric: Metric) -> dict[str, Any]:
+    """
+    Convert a Metric instance to a dictionary.
+
+    Args:
+        metric: An instance of a Metric.
+
+    Returns:
+        A dictionary representation of the given metric.
+    """
+    if is_dataclass_instance(metric):
         return dataclasses.asdict(metric)
-    elif attr.has(metric):
+    elif is_attrs_instance(metric):
         return attr.asdict(metric)
     else:
         assert False, "Unreachable"
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -391,10 +391,17 @@ def asdict(metric: Metric) -> Dict[str, Any]:
 
     Returns:
         A dictionary representation of the given metric.
+
+    Raises:
+        TypeError: If the given metric is not an instance of a `dataclass` or `attr.s`-decorated
+        Metric.
     """
     if _is_dataclass_instance(metric):
         return dataclasses.asdict(metric)
     elif _is_attrs_instance(metric):
         return attr.asdict(metric)
     else:
-        assert False, "Unreachable"
+        raise TypeError(
+            "The provided metric is not an instance of a `dataclass` or `attr.s`-decorated Metric "
+            f"class: {metric.__class__}"
+        )
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -386,6 +386,9 @@ def asdict(metric: Metric) -> Dict[str, Any]:
     """
     Convert a Metric instance to a dictionary.
 
+    No formatting is performed on the values, and they are returned as contained (and typed) in the
+    underlying dataclass. Use `Metric.format_value` to format the values to string.
+
     Args:
         metric: An instance of a Metric.
 
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -116,6 +116,7 @@ Formatting and parsing the values for custom types is supported by overriding th
 
 """
 
+import dataclasses
 from abc import ABC
 from enum import Enum
 from pathlib import Path
@@ -126,6 +127,8 @@ from typing import Generic
 from typing import Iterator
 from typing import List
 from typing import TypeVar
+
+import attr
 
 from fgpyo import io
 from fgpyo.util import inspect
@@ -334,3 +337,14 @@ class Metric(ABC, Generic[MetricType]):
             io.write_lines(
                 path=output, lines_to_write=list(io.read_lines(input_path))[1:], append=True
             )
+
+
+def asdict(metric: Metric) -> dict[str, Any]:
+    """Convert a Metric instance to a dictionary."""
+
+    if dataclasses.is_dataclass(metric):
+        return dataclasses.asdict(metric)
+    elif attr.has(metric):
+        return attr.asdict(metric)
+    else:
+        assert False, "Unreachable"
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -348,7 +348,7 @@ class Metric(ABC, Generic[MetricType]):
             )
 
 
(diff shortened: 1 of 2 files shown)


msto:
Note - I've tried changing the type hint to `metric: MetricType`, as in `Metric.write()`, but get the same errors

TedBrookings:
I'm sorry for the delay, today had a lot of non-work emergencies. This is my suggestion:
1. Inside the metric class, directly below the existing `values` method, add `asdict`
```
    def values(self) -> Iterator[Any]:
        """An iterator over attribute values in the same order as the header."""
        for field in inspect.get_fields(self.__class__):  # type: ignore[arg-type]
            yield getattr(self, field.name)

    def asdict(self) -> Dict[str, Any]:
        """A dictionary of attribute values in the same order as the header."""
        return {
            field.name: getattr(self, field.name)
            for field in inspect.get_fields(self.__class__)  # type: ignore[arg-type]
        }
```
You need to add the `# type: ignore[arg-type]` because we are insisting that all actual `Metric` subclasses will be attr.s or dataclasses. As far as I know there isn't any way to signal to `mypy` that this will be the case though. This task has really made me appreciate the advantage of a typing system that prioritizes traits over types.

2. Add this test to test_metric.py, directly below the existing test_metric_values
```
@pytest.mark.parametrize("data_and_classes", (attr_data_and_classes, dataclasses_data_and_classes))
def test_metric_values(data_and_classes: DataBuilder) -> None:
    assert list(data_and_classes.Person(name="name", age=42).values()) == ["name", 42]


@pytest.mark.parametrize("data_and_classes", (attr_data_and_classes, dataclasses_data_and_classes))
def test_metric_asdict(data_and_classes: DataBuilder) -> None:
    assert data_and_classes.Person(name="name", age=42).asdict() == {"name": "name", "age": 42}
```

I created and pushed a branch that does this: [tb-add-asdict](https://github.com/fulcrumgenomics/fgpyo/tree/tb-add-asdict). The tests [pass](https://github.com/fulcrumgenomics/fgpyo/actions/runs/8976866311).

TedBrookings (approved):
I've never heard of `TypeGuard` before, that's cool. I think that's probably the recipe to clean up a lot of the type ignore statements currently in inspect.

My only remaining suggestion is that I think `asdict` could be a member function of the `Metric` class, with all the statements just acting on `self`.

msto:
I had the same thought, but I'm of two minds.

I strongly prefer being consistent with established convention when possible, and both `dataclasses` and `attr` implement `asdict()` as a standalone function rather than an instance method.

However, packaging it as a method removes the need for an import (and possibly makes it more discoverable).

Curious what @nh13 @tfenne @clintval think

fgpyo/util/tests/test_metric.py
```diff
@@ -519,3 +522,24 @@ def test_metric_columns_out_of_order(tmp_path: Path, data_and_classes: DataBuild
...
+
+    assert is_dataclass_instance(dataclasses_data_and_classes.Person(name="name", age=42))
+    assert not is_dataclass_instance(attr_data_and_classes.Person(name="name", age=42))
+
+
+def test_is_attrs_instance() -> None:
+    """Test that is_attrs_instance works as expected."""
+
+    assert not is_attrs_instance(dataclasses_data_and_classes.Person(name="name", age=42))
+    assert is_attrs_instance(attr_data_and_classes.Person(name="name", age=42))
```
clintval, line 538: I would also add a `not` test to make sure it won't return `True` on class objects instead of instances.

```
assert not is_*_instance(Person)
```

clintval (commented):
I see why this convenience function is desired. It is useful when you have a `Metric` and you don't know if it is an attrs-defined or dataclass-defined instance. In my experience, I usually know which flavor I'm dealing with and use the respective import accordingly:

```python
attr.asdict(metric)
dataclasses.asdict(metric)
```

I lean towards letting the user import their specific "as dict" implementation for their use case over providing another import to do functionally the same thing but I won't let that opinion of mine block this PR! Here's another idea though, what about adding an `__iter__(self)` dunder method on the base Metric class so we can start doing `dict(metric)` instead, which uses a built-in? I'm also a bigger fan of `as_dict()` instead of `asdict()` if we're allowed to vote on function naming too!

Will this function be needed when we eventually remove attrs support? Should we consider not adding it to the public API because eventually all Metrics should be using `@dataclass` and can use the corresponding `dataclasses.asdict` built-in?

fgpyo/util/metric.py
```diff
@@ -334,3 +346,55 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+
+    Returns:
+        A dictionary representation of the given metric.
+    """
+    if is_dataclass_instance(metric):
+        return dataclasses.asdict(metric)
+    elif is_attrs_instance(metric):
+        return attr.asdict(metric)
+    else:
+        assert False, "Unreachable"
```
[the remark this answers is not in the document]
msto: Not at the moment. I've realized it's not actually unreachable, because `Metric` doesn't enforce that the class is decorated with `dataclass` or `attr.s`.

For now, we should probably raise a `TypeError` instead.

https://github.com/fulcrumgenomics/fgpyo/issues/114

fgpyo/util/tests/test_metric.py
msto: good call!

fgpyo/util/metric.py
```diff
@@ -334,3 +346,55 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        metric: An instance of a Metric.
+
+    Returns:
+        True if the given metric is an instance of an attr.s-decorated Metric.
+        False otherwise.
+    """
+    return not isclass(metric) and attr.has(metric.__class__)
+
+
+def asdict(metric: Metric) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
msto: Both `asdict` implementations return `dict[str, Any]` because they do recursively `dict`-ify the dataclass/attrs instance. I don't think we can further narrow the type here because of that.

https://stackoverflow.com/a/52229565

msto (commented):
> It is useful when you have a Metric and you don't know if it is an attrs-defined or dataclass-defined instance. In my experience, I usually know which flavor I'm dealing with and use the respective import accordingly:

Exactly, it's a convenience function to abstract the concern away.

IMO if we intend to support both `attr.s` and `dataclass`, then we should have an API that works with `Metric` (as a pseudo-alias for the union of `AttrsInstance` and `DataclassInstance`). 

This is also intended to facilitate the `MetricWriter` in #107 and other such utilities which may not be able to assume which import to use.

> what about adding an __iter__(self) dunder method on the base Metric class so we can start doing dict(metric) instead, which uses a built-in?

See below comment - `dict()` and `asdict()` [do different things](https://stackoverflow.com/a/52229565), and I think the latter implementation is preferable here.

> I'm also a bigger fan of as_dict() instead of asdict() if we're allowed to vote on function naming too!

I agree that snakecasing is generally preferable, but I have a stronger preference for not conflicting with the established naming convention from `dataclass` and `attr.s`

> Will this function be needed when we eventually remove attrs support? Should we consider not adding it to the public API because eventually all Metrics should be using @dataclass and can use the corresponding dataclasses.asdict built-in?

At that time we could simply import `dataclasses.asdict` into this module to avoid breakage? Or replace `from fgpyo.util.metric import asdict` with `from dataclasses import asdict` (which I consider another argument in favor of leaving the naming as is)

clintval (approved):
Make sense. I'm onboard.

My approval contingent on cleaning up the actually reachable-unreachable branches with a `TypeError`. Thanks Matt!

fgpyo/util/metric.py
```diff
@@ -334,3 +346,55 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+    """
+    Convert a Metric instance to a dictionary.
+
+    Args:
+        metric: An instance of a Metric.
+
+    Returns:
+        A dictionary representation of the given metric.
+    """
```
nh13, line 394: This overlooks the `format_value` method on `Metric` that is used to format values when written to a file.

nh13 (changes requested):

fgpyo/util/metric.py
msto: @nh13 the omission was deliberate. I left a comment on the topic in the PR description, but given how much conversation this one has attracted it was easy to overlook 🙂 

> NB: it may be valuable to permit formatting/casting the values as part of this function (e.g. by adding a parameter `format: bool = False`), but to do so we'll probably want to extract `Metric.format_value()` to a standalone function instead of a classmethod

I do not think we should have an `asdict()` function that changes the value types by default. This function is primarily intended as a dispatcher, selecting the correct (`dataclasses` or `attr.s`) function depending on how the `Metric` in question was selected. 

Ideally, this function will be deprecated once we drop support for `attrs`. As I mentioned in my comments above to Clint, when that happens it would be preferable to be able to transparently replace this with `dataclasses.asdict` 

> At that time we could simply import `dataclasses.asdict` into this module to avoid breakage? Or replace `from fgpyo.util.metric import asdict` with `from dataclasses import asdict`

I am open to adding `format: bool = False` to this function, with the stipulation that it should be `False` by default and the caveat that I expect it to add debt and increase friction when we deprecate `attr.s`.

However, I'd prefer to leave as is, and then call `format_value` on the resulting dict when necessary, e.g.
```py
metric_dict = {key: format_value(val) for key, val in asdict(metric)}
```
as I find the above relatively easy to understand

Thoughts?

nh13: Let’s omit it but document it clearly. 

I do like that the format_value is a class method so that any new Metric type can perform custom formatting but overriding the class method. I think passing in a parsing function, like defopt, in rare situations causes a conflict when we want to format the type differently.

fgpyo/util/metric.py
```diff
@@ -334,3 +346,62 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+    Returns:
+        True if the given metric is an instance of an attr.s-decorated Metric.
+        False otherwise.
+    """
+    return not isclass(metric) and attr.has(metric.__class__)
+
+
+def asdict(metric: Metric) -> Dict[str, Any]:
+    """
+    Convert a Metric instance to a dictionary.
```
nh13, line 387:
```suggestion
    Convert a Metric instance to a dictionary.  No formatting is performed on the values and so 
    they are returned as-is. 
```

fgpyo/util/metric.py
msto: I like the idea of overriding the classmethod, but can't imagine doing so in practice. Given that it specifies formatting behavior for a wide variety of primitive and compound types, it doesn't seem to lend itself to easy extension or modification. (Since in order to override formatting for one type, you would have to override them all.)

Issue #92, opened 2024-02-21
Feature request: UMI extraction from read name


msto:
It would be great to have a utility function to extract a UMI from a read name, similar to the functionality provided by `fgbio CopyUmiFromReadName`

http://fulcrumgenomics.github.io/fgbio/tools/latest/CopyUmiFromReadName.html

emmcauley:
Closed via #118

PR #118, opened 2024-05-20
feat: Add methods for UMI extraction from read name

emmcauley:
Addresses https://github.com/fulcrumgenomics/fgpyo/issues/92

* Added extract_umis_from_read_name() (based on gist provided by Matt); modified to accept a bool `strict`: 
-- if `strict=True`, we count number of colons in read name to determine whether to return None or the last element in the read name, which is assumed to be UMI (should be consistent with `fgbio` [functionality](https://github.com/fulcrumgenomics/fgbio/blob/ba0788ea4ea9667d74ab68e156c0877553ad999c/src/main/scala/com/fulcrumgenomics/umi/Umis.scala#L70)

* Added copy_umi_from_read_name() with bool `remove_umi`:
-- Copies the UMI to the record's `RX tag`
-- If `remove_umi = True`, then we update the `rec.qname` in place to include everything but the UMI

* Added testing coverage in [test_umi_methods.py ](fgpyo/sam/tests/test_umi_methods.py)
-- test_strict_extract_umi_from_read_name(): if `strict=True`, colon count of read_name becomes relevant
-- test_strict_extract_umi_from_read_name_raises(): test that when `strict=True` a `ValueError()` is raised when number of colons is not 7 or 8 as expected
-- test_copy_umi_from_read_name() with and without `remove_umi=True`


fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,111 @@ class SamOrder(enum.Enum):
...
     Unknown = "unknown"  # Unknown SAM / BAM / CRAM sort order
+
+
+def extract_umis_from_read_name(
+    read_name: str,
+    read_name_delimiter: str = ILLUMINA_READ_NAME_DELIMITER,
+    umi_delimiter: str = ILLUMINA_UMI_DELIMITER,
+    strict: bool = False,
+) -> Optional[str]:
+    """Extract UMI(s) from a read name.
```
nh13, line 960:
```suggestion
    """Extract UMI(s) from an Illumina-style read name.
```

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,111 @@ class SamOrder(enum.Enum):
...
+        if strict:
+            raise ValueError(
+                f"Invalid UMIs found in read name: {read_name}",
+                f"  (Invalid UMIs: {', '.join(invalid_umis)})",
+            )
+        else:
+            return None
+
+    else:
+        return SAM_UMI_DELIMITER.join(umis)
```
nh13, line 1010: I think we can re-order this to remove the nesting.  Strictly speaking, we could change the `elif` to an `if`, and then remove the `else`, but I think this is clearer:

```suggestion
    if len(invalid_umis) == 0:
        return SAM_UMI_DELIMITER.join(umis)
    elif strict:
        raise ValueError(
            f"Invalid UMIs found in read name: {read_name}",
            f"  (Invalid UMIs: {', '.join(invalid_umis)})",
        )
    else:
        return None
```

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,111 @@ class SamOrder(enum.Enum):
...
+
+
+def copy_umi_from_read_name(
+    rec: AlignedSegment, strict: bool = False, remove_umi: bool = False
+) -> None:
+    """
+    Copy a UMI from an alignment's read name to its `RX` SAM tag. UMI will not be copied to RX
+     tag if invalid.
```
nh13, line 1018:
```suggestion
    tag if invalid.
```

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,111 @@ class SamOrder(enum.Enum):
...
+        else:
+            return None
+
+    else:
+        return SAM_UMI_DELIMITER.join(umis)
+
+
+def copy_umi_from_read_name(
+    rec: AlignedSegment, strict: bool = False, remove_umi: bool = False
+) -> None:
```
nh13, line 1015: consider returning a `bool` indicating if the UMI was valid.  This could be useful for testing when strict is false.

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,111 @@ class SamOrder(enum.Enum):
...
+        if strict:
+            raise ValueError(f"Invalid UMI {umi} extracted from {rec.query_name}")
+        else:
+            return
+
+    rec.set_tag(tag="RX", value=umi)
+
+    if remove_umi:
+        last_index = rec.query_name.rfind(ILLUMINA_READ_NAME_DELIMITER)
+        rec.query_name = rec.query_name[:last_index] if last_index != -1 else rec.query_name
```
nh13, line 1046: I _think_ this can be simplified by reorganizing the branches
```suggestion
    if umi is not None:
        rec.set_tag(tag="RX", value=umi)
        if remove_umi:
            last_index = rec.query_name.rfind(ILLUMINA_READ_NAME_DELIMITER)
            rec.query_name = rec.query_name[:last_index] if last_index != -1 else rec.query_name
    elif strict:
          raise ValueError(f"Invalid UMI {umi} extracted from {rec.query_name}")
```

fgpyo/sam/__init__.py
```diff
@@ -189,6 +190,20 @@
...
+
+VALID_UMI_CHARACTERS: Set[str] = set("ACGTN")
+"""Illumina's restricted UMI characters;
+https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/FileFormat_FASTQ-files_swBS.htm."""  # noqa
+
+ILLUMINA_UMI_DELIMITER: str = "+"
+"""Multiple UMIs are delimited with a plus-sign in Illumina FASTQs; see docs above."""
+
+ILLUMINA_READ_NAME_DELIMITER: str = ":"
+"""Illumina read names are delimited with a colon."""
```
nh13, line 205: Can these be private?

fgpyo/sam/tests/test_umi_methods.py
```diff
@@ -0,0 +1,117 @@
...
+        extract_umis_from_read_name(read_name, strict=True)
+
+
+@pytest.mark.parametrize("remove_umi, strict", [[True, False], [True, False]])
+def test_copy_valid_umi_from_read_name(remove_umi: bool, strict: bool) -> None:
+    """Test that we populate the RX field with a valid UMI if remove_umi and strict
+    are both True; otherwise do not remove UMI from read.query_name"""
+    builder = SamBuilder()
+    read = builder.add_single(name="abc:def:ghi:jfk:lmn:opq:rst:GATTACA")
+    copy_umi_from_read_name(read, strict=strict, remove_umi=remove_umi)
```
nh13, line 102: if the method returned a value, which I suggested above, you can test it here.

nh13 (changes requested):
This is looking really good.  Just a few suggestions to make it a little more concise.

fgpyo/sam/__init__.py
```diff
@@ -934,3 +947,95 @@ class SamOrder(enum.Enum):
...
+    invalid_umis = [umi for umi in umis if not _is_valid_umi(umi)]
+    if len(invalid_umis) > 0:
+        raise ValueError(
+            f"Invalid UMIs found in read name: {read_name}",
+            f"  (Invalid UMIs: {', '.join(invalid_umis)})",
+        )
+    return SAM_UMI_DELIMITER.join(umis)
+
+
+def copy_umi_from_read_name(rec: AlignedSegment, remove_umi: bool = False) -> None:
```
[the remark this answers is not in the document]
emmcauley: @nh13 @tfenne just bumping this comment to get your thoughts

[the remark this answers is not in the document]
nh13: I think it's clear that this method has side-effects, so I'd keep it as is.  We can always add an `in_place: bool = False` argument later

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,112 @@ class SamOrder(enum.Enum):
...
+        ValueError: If the record already has a populated `RX` SAM tag.
+    """
+
+    umi = extract_umis_from_read_name(
+        read_name=rec.query_name, strict=strict, umi_delimiter=_ILLUMINA_READ_NAME_DELIMITER
+    )
+    if umi is not None:
+        if rec.has_tag("RX"):
+            return False
+            raise ValueError(f"Record {rec.query_name} already has a populated RX tag")
```
nh13, line 1037:
```suggestion
            raise ValueError(f"Record {rec.query_name} already has a populated RX tag")
```
The `raise` line can never be reached since there's a return before it.  I don't think you want to return here.

fgpyo/sam/__init__.py
```diff
@@ -934,3 +949,112 @@ class SamOrder(enum.Enum):
...
+        rec.set_tag(tag="RX", value=umi)
+        if remove_umi:
+            last_index = rec.query_name.rfind(_ILLUMINA_READ_NAME_DELIMITER)
+            rec.query_name = rec.query_name[:last_index] if last_index != -1 else rec.query_name
+        return True
+    elif strict:
+        return False
+        raise ValueError(f"Invalid UMI {umi} extracted from {rec.query_name}")
```
nh13, line 1045:
```suggestion
        raise ValueError(f"Invalid UMI {umi} extracted from {rec.query_name}")
```
ditto, the `raise` line can never be reached.  As per your docstring, if `strict` is `true`, you wan to raise the exception.

fgpyo/sam/tests/test_umi_methods.py
```diff
@@ -0,0 +1,127 @@
...
+    ],
+)
+def test_extract_umi_from_read_name_raises(read_name: str) -> None:
+    """Test that we raise an error when the read name includes an invalid UMI
+    and strict=True."""
+    with pytest.raises(ValueError):
+        extract_umis_from_read_name(read_name=read_name, strict=True)
+
+
+def test_extract_umi_from_read_name_strict_False() -> None:
```
nh13, line 59: naming conventions
```suggestion
def test_extract_umi_from_read_name_strict_false() -> None:
```

fgpyo/sam/tests/test_umi_methods.py
```diff
@@ -0,0 +1,127 @@
...
+    else:
+        assert read.query_name == "abc:def:ghi:jfk:lmn:opq:rst:GATTACA"
+
+
+def test_populated_rx_tag() -> None:
+    """Test that we raise a ValueError when a record already has a populated RX tag"""
+    builder = SamBuilder()
+    read = builder.add_single(name="abc:def:ghi:jfk:lmn:opq:rst:GATTACA")
+    read.set_tag(tag="RX", value="NNNNACGT")
+    assert copy_umi_from_read_name(read, strict=False, remove_umi=True) is False
```
nh13, line 115: I don't think we want the return value to be `False`, we want it to raise an exception.

fgpyo/sam/__init__.py
nh13: Fix the associated unit test to check for the exception please

fgpyo/sam/tests/test_umi_methods.py
```diff
@@ -0,0 +1,127 @@
...
+    [
+        ("abc:def:ghi:jfk"),
+        ("abc:def:ghi:jfk:lmn:opq:rst:uvw:xyz"),
+        ("abc:def:ghi:jfk:lmn:opq:rst:"),  # Invalid UMI
+    ],
+)
+def test_strict_extract_umi_from_read_name_raises(read_name: str) -> None:
+    """Test that we raise an error when strict=True and number of colons is not 7 or 8."""
+    with pytest.raises(ValueError):
+        extract_umis_from_read_name(read_name, strict=True)
```
nh13, line 93: How is this unit test passing as-is?  I think it hits this branch, which doesn't raise but returns:

```python
    elif strict:
        return False
        raise ValueError(f"Invalid UMI {umi} extracted from {rec.query_name}")
```

nh13 (changes requested):
Can you carefully check that `copy_umi_from_read_name` raises an exception or returns a value (but not both)?  And then make sure we have unit tests covering each return/raise?  There's one unit test which I can't understand why it was passing (it asserts a raise, but it looks like the function returns False before it raises.

fgpyo/sam/tests/test_umi_methods.py
emmcauley: I think this unit-test is overloaded because we have 2 different `ValueError` conditions: one for whether we have 7 or 8 colon-delimited fields, and one for when we encounter an invalid UMI (which is the last condition in the testing input data). The unit test was passing because it was raising a `ValueError`, just not the one that corresponds to the code snippet you included (it instead raised an earlier `ValueError` corresponding to `f"Trying to extract UMIs from read with {colons + 1} parts, (7 or 8 expected): {read_name}"`).

I separated out the two input data conditions and also added a `match` to the raise condition so we can track exactly which `ValueError` is getting triggered.

fgpyo/sam/__init__.py
```diff
@@ -189,6 +190,20 @@
 _IOClasses = (io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase)
 """The classes that should be treated as file-like classes"""
 
+SAM_UMI_DELIMITER: str = "-"
+"""Multiple UMI delimiter, which SAM specification recommends should be a hyphen;
+see specification here: https://samtools.github.io/hts-specs/SAMtags.pdf"""
+
+_VALID_UMI_CHARACTERS: Set[str] = set("ACGTN")
```
nh13, line 197: can you move this method into a `fgpyo/vendor/illumina.py` file?  See comment from @tfenne here: https://github.com/fulcrumgenomics/fgpyo/pull/151#issuecomment-2192098876

nh13 (changes requested):
One more suggestion for where this function should live.

fgpyo/sam/__init__.py
msto: Just to clarify, `fgpyo/platform/illumina.py` is the destination we decided on

nh13 (changes requested):
tests are in the wrong place

nh13 (approved):

PR #121, opened 2024-05-27
fix: r/w assertions

### LINKED ISSUE #117: `io.assert_path_is_writeable()` should return True for "/dev/stdout"
```py
In [5]: io.assert_path_is_writeable(Path("/dev/stdout"))
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[5], line 1
----> 1 io.assert_path_is_writeable(Path("/dev/stdout"))

File ~/.mambaforge-x86/envs/gta/lib/python3.10/site-packages/fgpyo/io/__init__.py:108, in assert_path_is_writeable(path, parent_must_exist)
    106 # If file exists, it must be writeable
    107 if path.exists():
--> 108     assert path.is_file() and os.access(
    109         path, os.W_OK
    110     ), f"File exists but is not writeable: {path}"
    112 # Else if file doesnt exist and parent_must_exist is True then check
    113 # that path.absolute().parent exists, is a directory and is writeable
    114 elif parent_must_exist:

AssertionError: File exists but is not writeable: /dev/stdout

In [6]: with Path("/dev/stdout").open("w") as fout:
   ...:     fout.write("abc\n")
   ...:
abc

```

msto:
Closes #117 

- I updated `assert_path_is_readable()` and `assert_path_is_writable()` to specially handle `/dev/stdin` and `/dev/stdout`, respectively
- I updated `assert_path_is_readable()` to specifically require that the input `path` is a file, not just that it is not a directory. (This disparity between the two functions is why `stdin` was ok for the `readable` check, but `stdout` was not for the `writable` check)
- I lightly refactored the chain of assertions in `assert_path_is_writable()` to produce more specific error messages and for legibility.
- I updated `assert_path_is_writable()` so the absolute path is considered regardless of the value of `parent_must_exist`
- Minor housekeeping in the corresponding unit tests

### DIFF
--- a/fgpyo/io/__init__.py
+++ b/fgpyo/io/__init__.py
@@ -148,7 +148,7 @@ def assert_path_is_writable(path: Path, parent_must_exist: bool = True) -> None:
     # Else if file doesn't exist and parent_must_exist is False, test parent until
     # you find the first extant path, and check that it is a directory and is writable.
     else:
-        for parent in path.parents:
+        for parent in path.absolute().parents:
             if parent.exists():
                 assert os.access(parent, os.W_OK), f"Parent directory is not writable: {parent}"
                 break
--- a/fgpyo/io/__init__.py
+++ b/fgpyo/io/__init__.py
@@ -72,8 +72,12 @@ def assert_path_is_readable(path: Path) -> None:
     Example:
         assert_file_exists(path = Path("some_file.csv"))
     """
+    # stdin is readable
+    if path == Path("/dev/stdin"):
+        return
+
     assert path.exists(), f"Cannot read non-existent path: {path}"
-    assert not path.is_dir(), f"Cannot read path becasue it is a directory: {path}"
+    assert path.is_file(), f"Cannot read path becasue it is not a file: {path}"
     assert os.access(path, os.R_OK), f"Path exists but is not readable: {path}"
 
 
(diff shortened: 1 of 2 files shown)


fgpyo/io/__init__.py
```diff
@@ -124,30 +128,32 @@ def assert_path_is_writable(path: Path, parent_must_exist: bool = True) -> None:
...
-        ), f"Path does not exist and parent isn't extant/writable: {path}"
+        assert parent.exists(), f"Parent directory does not exist: {parent}"
+        assert parent.is_dir(), f"Parent directory exists but is not a directory: {parent}"
+        assert os.access(parent, os.W_OK), f"Parent directory exists but is not writable: {parent}"
 
     # Else if file doesn't exist and parent_must_exist is False, test parent until
     # you find the first extant path, and check that it is a directory and is writable.
     else:
-        for parent in path.parents:
+        for parent in path.absolute().parents:
```
msto, line 151: @nh13 @tfenne it wasn't clear to me why we only considered the absolute path when `parent_must_exist=True`. If the omission was intentional please lmk and I'll revert

fgpyo/io/__init__.py
```diff
@@ -72,8 +72,12 @@ def assert_path_is_readable(path: Path) -> None:
     Example:
         assert_file_exists(path = Path("some_file.csv"))
     """
+    # stdin is readable
+    if path == Path("/dev/stdin"):
+        return
```
nh13, line 77:
```suggestion
        return None
```

fgpyo/io/__init__.py
```diff
@@ -124,30 +128,32 @@ def assert_path_is_writable(path: Path, parent_must_exist: bool = True) -> None:
     Example:
         assert_path_is_writable(path = Path("example.txt"))
     """
-    # If file exists, it must be writable
+    # stdout is writable
+    if path == Path("/dev/stdout"):
+        return
```
nh13, line 133:
```suggestion
        return None
```

fgpyo/io/__init__.py
nh13: Taking a shit in the dark, what if the path includes a symbolic link?

nh13 (approved):
LGTM, with two nit suggestions (avoid empty returns), and one possible explanation for reverting the use of absolute.

fgpyo/io/__init__.py
msto: Does it matter? Won't the symbolic link inherit the permissions of the target?

nh13:
@msto ready to merge.  I think my objection is too rare to happen in practice, if at all.

fgpyo/io/__init__.py
msto: Resolved per slack discussion

> `Path.is_file()` returns `True` for symlinks


> Let's say you have the following
> ```
> /heres/the/symlinked/path/to/file -> /and/the/actual/path/to/file
> ```
> In the context of the assert, the question is basically only relevant when `file`, `to`, and `path` do not exist
> and afaik a symlinked directory is writable iff the target directory is writable

PR #123, opened 2024-06-04
feat: MetricWriter

### LINKED ISSUE #88: Add MetricWriter to permit streamed writing of metrics

### LINKED ISSUE #90: Metric.write should support excluding attributes
It is sometimes helpful to declare attributes on a `Metric` that are useful for internal logic, but would clutter the output.

Could we update `Metric.write` to accept a list of attribute names that will be excluded from the output?

msto:
Closes #88 
Closes #90 

## Summary

This PR introduces a `MetricWriter` class. 

This PR is motivated by a limitation in `Metric.write()`, which does not permit appending metric instances to a file. The new class supports appending and streamed writing of metrics, validating that the existing file's structure matches the metrics being written. `MetricWriter` also supports optionally reordering or removing fields when writing.

## Added

- Added the `MetricWriter` class, which supports context-managed appending and writing of Metrics to file.
- Added `Metric.keys()`, `Metric.items()`, and `Metric.formatted_items()`. These are companions to the existing `Metric.values()` and `Metric.formatted_values()`, and provide analogous key/item views
- Added `Metric._read_header` and the associated `MetricFileHeader` class, to parse a file's header (and any preceding comment lines). 
    - NB: This code is employed in #103 to support reading of Picard-style metrics files.

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -479,7 +479,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
         """
 
         filepath: Path = Path(filename)
-        ordered_fieldnames: List[str] = _validate_output_fieldnames(
+        ordered_fieldnames: List[str] = _validate_and_generate_final_output_fieldnames(
             metric_class=metric_class,
             include_fields=include_fields,
             exclude_fields=exclude_fields,
@@ -566,7 +566,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
             self.write(metric)
 
 
-def _validate_output_fieldnames(
+def _validate_and_generate_final_output_fieldnames(
     metric_class: Type[MetricType],
     include_fields: Optional[List[str]] = None,
     exclude_fields: Optional[List[str]] = None,
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -142,8 +142,6 @@ if sys.version_info >= (3, 10):
 else:
     from typing_extensions import TypeGuard
 
-import attr
-
 from fgpyo import io
 from fgpyo.util import inspect
 
@@ -432,11 +430,15 @@ class Metric(ABC, Generic[MetricType]):
 def _is_metric_class(cls: Any) -> TypeGuard[Metric]:
     """True if the given class is a Metric."""
 
-    return (
-        isclass(cls)
-        and issubclass(cls, Metric)
-        and (dataclasses.is_dataclass(cls) or attr.has(cls))
-    )
+    is_metric_cls: bool = isclass(cls) and issubclass(cls, Metric)
+
+    try:
+        import attr
+        is_metric_cls = is_metric_cls and (dataclasses.is_dataclass(cls) or attr.has(cls))
+    except ImportError:
+        is_metric_cls = is_metric_cls and dataclasses.is_dataclass(cls)
+
+    return is_metric_cls
 
 
 class MetricWriter(Generic[MetricType], AbstractContextManager):
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -397,7 +397,8 @@ class Metric(ABC, Generic[MetricType]):
 
         The first row after any commented or empty lines will be used as the fieldnames.
 
-        Lines preceding the fieldnames will be returned in the `preamble`.
+        Lines preceding the fieldnames will be returned in the `preamble`. Leading and trailing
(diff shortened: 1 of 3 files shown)


fgpyo/util/metric.py
```diff
@@ -334,3 +365,104 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        reader: TextIOWrapper,
+        delimiter: str = "\t",
+        comment_prefix: str = "#",
+    ) -> MetricFileHeader:
+        """
+        Read the header from an open file.
+
+        The first row after any commented or empty lines will be used as the fieldnames.
+
+        Lines preceding the fieldnames will be returned in the `preamble`.
```
nh13, line 380: mention that leading and trailing whitespaces are ignored.

fgpyo/util/metric.py
```diff
@@ -334,3 +365,104 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+            A `MetricFileHeader` containing the field names and any preceding lines.
+
+        Raises:
+            ValueError: If the file was empty or contained only comments or empty lines.
+        """
+
+        preamble: List[str] = []
+
+        for line in reader:
+            if line.startswith(comment_prefix) or line.strip() == "":
```
nh13, line 397: comment can start with a whitespace right?

fgpyo/util/metric.py
```diff
@@ -334,3 +365,104 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+                preamble.append(line.strip())
+            else:
+                break
+        else:
+            raise ValueError(f"File {reader.name} did not contain a header line.")
+
+        fieldnames = line.strip().split(delimiter)
```
nh13, line 404: probably want to check for an empty "line" since that's what we'd get if the file was empty (good thing to add a unit test for too)

fgpyo/util/metric.py
```diff
@@ -334,3 +365,104 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+        metric: An instance of a Metric.
+
+    Returns:
+        True if the given metric is an instance of an attr.s-decorated Metric.
+        False otherwise.
+    """
+    return not isclass(metric) and attr.has(metric.__class__)
+
+
+def asdict(metric: Metric) -> Dict[str, Any]:
```
nh13, line 443: why not be an instance method since it only takes one argument (a Metric)?

fgpyo/util/tests/test_metric.py
```diff
@@ -519,3 +522,71 @@ def test_metric_columns_out_of_order(tmp_path: Path, data_and_classes: DataBuild
...
+
+        def __init__(self, foo: int, bar: str):
+            self.foo = foo
+            self.bar = bar
+
+    with pytest.raises(TypeError, match="The provided metric is not an instance"):
+        asdict(UndecoratedMetric(foo=1, bar="a"))
+
+
+def test_read_header_can_read_picard(tmp_path: Path) -> None:
```
nh13, line 573: add tests:
1. empty file (fails: no fieldnames)
2. file with no comment lines

nh13 (dismissed):
Looking good so far, just a couple small things

fgpyo/util/metric.py
```diff
@@ -331,3 +385,300 @@ def fast_concat(*inputs: Path, output: Path) -> None:
...
+
+    Raises:
+        ValueError: If the provided file does not include a header.
+        ValueError: If the header of the provided file does not match the provided Metric (or list
+            of ordered fieldnames, if provided).
+    """
+    _assert_is_metric_class(metric_class)
+
+    with path.open("r") as fin:
+        header: MetricFileHeader = metric_class._read_header(fin, delimiter=delimiter)
```
tfenne, line 632: Do you maybe want to check that `path` is a regular path first?  What if we're asked to append to stdout?

tfenne (changes requested):
Could you also replace metric.write() with an implementaiton that uses the new writer so we don't have duplicate code paths/

tfenne (approved):

PR #129, opened 2024-06-06
feat: Add assertions to support `MetricWriter`

msto:
I split the remaining code in #107 into two PRs to facilitate review.

This PR introduces several of the assertion methods used when constructing the `MetricWriter`. (See #107 for how they are used in practice).

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -130,6 +130,7 @@ from typing import Dict
 from typing import Generic
 from typing import Iterator
 from typing import List
+from typing import Type
 from typing import TypeVar
 
 if sys.version_info >= (3, 10):
@@ -406,6 +407,16 @@ class Metric(ABC, Generic[MetricType]):
         return MetricFileHeader(preamble=preamble, fieldnames=fieldnames)
 
 
+def _is_metric_class(cls: Any) -> TypeGuard[Metric]:
+    """True if the given class is a Metric."""
+
+    return (
+        isclass(cls)
+        and issubclass(cls, Metric)
+        and (dataclasses.is_dataclass(cls) or attr.has(cls))
+    )
+
+
 def _is_dataclass_instance(metric: Metric) -> TypeGuard[DataclassInstance]:
     """
     Test if the given metric is a dataclass instance.
@@ -466,3 +477,98 @@ def asdict(metric: Metric) -> Dict[str, Any]:
             "The provided metric is not an instance of a `dataclass` or `attr.s`-decorated Metric "
             f"class: {metric.__class__}"
         )
+
+
+def _get_fieldnames(metric_class: Type[Metric]) -> List[str]:
+    """
+    Get the fieldnames of the specified metric class.
+
+    Args:
+        metric_class: A Metric class.
+
+    Returns:
+        A list of fieldnames.
+    """
+    _assert_is_metric_class(metric_class)
+
+    if dataclasses.is_dataclass(metric_class):
+        return [f.name for f in dataclasses.fields(metric_class)]
+    elif attr.has(metric_class):
+        return [f.name for f in attr.fields(metric_class)]
+    else:
+        assert False, "Unreachable"
+
+
+def _assert_file_header_matches_metric(
+    path: Path,
+    metric_class: Type[MetricType],
+    delimiter: str,
+) -> None:
+    """
+    Check that the specified file has a header and its fields match those of the provided Metric.
+
+    Args:
+        path: A path to a `Metric` file.
+        metric_class: The `Metric` class to validate against.
+        delimiter: The delimiter to use when reading the header.
+
+    Raises:
+        ValueError: If the provided file does not include a header.
+        ValueError: If the header of the provided file does not match the provided Metric.
+    """
+    # NB: _get_fieldnames() will validate that `metric_class` is a valid Metric class.
+    fieldnames: List[str] = _get_fieldnames(metric_class)
+
+    header: MetricFileHeader
+    with path.open("r") as fin:
+        try:
+            header = metric_class._read_header(fin, delimiter=delimiter)
+        except ValueError:
+            raise ValueError(f"Could not find a header in the provided file: {path}")
+
+    if header.fieldnames != fieldnames:
+        raise ValueError(
+            "The provided file does not have the same field names as the provided Metric:\n"
+            f"\tMetric: {metric_class.__name__}\n"
+            f"\tFile: {path}\n"
+            f"\tExpected fields: {', '.join(fieldnames)}\n"
+            f"\tActual fields: {', '.join(header.fieldnames)}\n"
+        )
+
+
+def _assert_fieldnames_are_metric_attributes(
+    specified_fieldnames: List[str],
+    metric_class: Type[MetricType],
+) -> None:
+    """
+    Check that all of the specified fields are attributes on the given Metric.
+
+    Raises:
+        ValueError: if any of the specified fieldnames are not an attribute on the given Metric.
+    """
+    _assert_is_metric_class(metric_class)
+
+    invalid_fieldnames = {
+        f for f in specified_fieldnames if f not in _get_fieldnames(metric_class)
+    }
+
+    if len(invalid_fieldnames) > 0:
+        raise ValueError(
+            "One or more of the specified fields are not attributes on the Metric "
+            + f"{metric_class.__name__}: "
+            + ", ".join(invalid_fieldnames)
+        )
+
+
+def _assert_is_metric_class(cls: Type[Metric]) -> None:
+    """
+    Assert that the given class is a Metric.
+
+    Args:
+        cls: A class object.
+
+    Raises:
+        TypeError: If the given class is not a Metric.
+    """
+    if not _is_metric_class(cls):
+        raise TypeError(f"Not a dataclass or attr decorated Metric: {cls}")
--- a/fgpyo/util/tests/test_metric.py
+++ b/fgpyo/util/tests/test_metric.py
@@ -30,6 +30,8 @@ import pytest
 from fgpyo.util.inspect import is_attr_class
 from fgpyo.util.inspect import is_dataclasses_class
 from fgpyo.util.metric import Metric
+from fgpyo.util.metric import _assert_fieldnames_are_metric_attributes
+from fgpyo.util.metric import _assert_file_header_matches_metric
 from fgpyo.util.metric import _assert_is_metric_class
 from fgpyo.util.metric import _get_fieldnames
 from fgpyo.util.metric import _is_attrs_instance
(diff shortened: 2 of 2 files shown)


fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    """
+    _assert_is_metric_class(metric_class)
+
+    with path.open("r") as fin:
+        header: MetricFileHeader = metric_class._read_header(fin, delimiter=delimiter)
+
+    if header is None:
+        raise ValueError(f"Could not find a header in the provided file: {path}")
+
+    fieldnames: List[str] = _get_fieldnames(metric_class)
```
nh13, line 533: could move this to the start of the method, then remove the `_assert_is_metric_class` since `_get_fieldnames` calls the assert itself

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    """
+    Check that all of the specified fields are attributes on the given Metric.
+
+    Raises:
+        ValueError: if any of the specified fieldnames are not an attribute on the given Metric.
+    """
+    _assert_is_metric_class(metric_class)
+
+    invalid_fieldnames = [
+        f for f in specified_fieldnames if f not in _get_fieldnames(metric_class)
```
nh13, line 557: cache `_get_fieldnames(metric_class)`?  Also make it a `set`?

fgpyo/util/tests/test_metric.py
```diff
@@ -590,3 +593,70 @@ def test_read_header_can_read_picard(tmp_path: Path) -> None:
...
+def test_get_fieldnames(data_and_classes: DataBuilder) -> None:
+    """Test we can get the fieldnames of a metric."""
+
+    assert _get_fieldnames(data_and_classes.Person) == ["name", "age"]
+
+
+def test_fieldnames_raises_if_not_a_metric() -> None:
+    """Test we raise if we get a non-metric."""
+
+    @dataclass
```
nh13, line 608: but isn't this a dataclass?

nh13 (changes requested):
A few minor comments I'd like to see addressed.  Can you also add tests (pass and fail) for:

* ` _assert_fieldnames_are_metric_attributes`
* `_assert_file_header_matches_metric`

fgpyo/util/metric.py
```diff
@@ -406,6 +407,16 @@ def _read_header(
         return MetricFileHeader(preamble=preamble, fieldnames=fieldnames)
 
 
+def _is_metric_class(cls: Any) -> TypeGuard[Metric]:
```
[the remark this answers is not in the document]
msto: `TypeGuard` is an alias for `bool` that narrows the type of the function's argument

https://mypy.readthedocs.io/en/latest/type_narrowing.html

fgpyo/util/metric.py
msto: python sets are unordered, and `fieldnames` should appear in the order they are defined on the Metric

fgpyo/util/tests/test_metric.py
msto: But not a `Metric` subclass

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    path: Path,
+    metric_class: Type[MetricType],
+    delimiter: str,
+) -> None:
+    """
+    Check that the specified file has a header and its fields match those of the provided Metric.
+
+    If `ordered_fieldnames` is specified, the fieldnames in the header must match this list exactly.
+    Otherwise, the fieldnames in the header must match the fieldnames of the provided Metric.
+
```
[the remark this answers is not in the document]
msto: Sorry, artifact of splitting up #107. It's introduced in the next PR, I've removed reference to it here.

tfenne (changes requested):
Stopping since you're actively working on it.  Once you've addressed @nh13's comments I'll take another pass.

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+        delimiter: The delimiter to use when reading the header.
+
+    Raises:
+        ValueError: If the provided file does not include a header.
+        ValueError: If the header of the provided file does not match the provided Metric.
+    """
+    _assert_is_metric_class(metric_class)
+
+    with path.open("r") as fin:
+        header: MetricFileHeader = metric_class._read_header(fin, delimiter=delimiter)
```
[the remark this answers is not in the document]
msto: This should be a `try-except` since #124 changed the return type of `_read_header` to be non-`Optional`.

I've fixed

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+
+    Raises:
+        TypeError: If the given class is not a Metric.
+    """
+    _assert_is_metric_class(metric_class)
+
+    if dataclasses.is_dataclass(metric_class):
+        return [f.name for f in dataclasses.fields(metric_class)]
+    elif attr.has(metric_class):
+        return [f.name for f in attr.fields(metric_class)]
```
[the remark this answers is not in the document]
msto: Short answer: this is an artifact from the initial proof-of-concept for this class being implemented in https://github.com/msto/dataclass_io/ (i.e. only for dataclasses, not considering `Metric`)

Long answer: `metric.header()` and `inspect.get_fields()` have a *lot* of `type: ignore`s , and I'd rather use an implementation that's type-safe. (I have a mental TODO to clean up the volume of `type: ignores` in the repo, and I can leave a TODO that this function should be deprecated/merged into `inspect` once that happens.)

fgpyo/util/metric.py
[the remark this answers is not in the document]
tfenne: Thanks!

fgpyo/util/metric.py
```diff
@@ -466,3 +477,101 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+        raise ValueError(
+            "The provided file does not have the same field names as the provided Metric:\n"
+            f"\tMetric: {metric_class.__name__}\n"
+            f"\tFile: {path}\n"
+            f"\tExpected fields: {', '.join(fieldnames)}\n"
+            f"\tActual fields: {', '.join(header.fieldnames)}\n"
+        )
+
+
+def _assert_fieldnames_are_metric_attributes(
```
tfenne, line 542: Making a note here as I'm not sure where else to put it ... but since this is also leading up to the `picard` metric PR, do we want to allow for case insensivity, and where would be the right place to put it?  I.e. it would be nice to support picard metrics and not have the classes yelling at us.

fgpyo/util/metric.py
```diff
@@ -466,3 +477,101 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+) -> None:
+    """
+    Check that all of the specified fields are attributes on the given Metric.
+
+    Raises:
+        ValueError: if any of the specified fieldnames are not an attribute on the given Metric.
+    """
+    _assert_is_metric_class(metric_class)
+
+    invalid_fieldnames = [
```
tfenne, line 554: For defensiveness, want to make this a set?

fgpyo/util/metric.py
```diff
@@ -466,3 +477,101 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    Assert that the given class is a Metric.
+
+    Args:
+        cls: A class object.
+
+    Raises:
+        TypeError: If the given class is not a Metric.
+    """
+    if not _is_metric_class(cls):
+        raise TypeError(f"Not a dataclass or attr decorated Metric: {cls}")
```
tfenne, line 577: How do we feel about `_assert` functions raising errors other than `AssertionError`?  I would probably be surprised by this.

fgpyo/util/metric.py
msto: I think that's a good suggestion but outside the scope of the current PR (and outside the scope of the larger `MetricWriter` feature branch).

The open PR to support reading Picard metrics headers (#103) will work as-is to read Picard metrics files if the `Metric` class's attributes are all caps. I think case insensitivity would be a nice QoL improvement and I'll open a separate issue to support it.

fgpyo/util/metric.py
msto: I don't think `assert` should be used outside of unit tests, and I prefer to raise more specific exceptions.

The Google style guide [recommends/requires this as well](https://google.github.io/styleguide/pyguide.html#244-decision):
> Do not use assert statements in place of conditionals or validating preconditions. They must not be critical to the application logic. A litmus test would be that the assert could be removed without breaking the code. assert conditionals are [not guaranteed](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) to be evaluated. For [pytest](https://pytest.org/) based tests, assert is okay and expected to verify expectations.


The primary reason to avoid the use of assert statements is that [they are intended for debugging and can be disabled](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

Additionally, using `assert` makes it easy to fail to cover all branches of a program:
```python
# Covered by any unit test that calls the parent function
assert is_ok(foo), "Foo was not ok"

if not is_ok(foo):
    # Not covered unless the condition where `is_ok(foo)` is False is tested explicitly
    raise ValueError("Foo was not ok")  
```

I'm happy to rename the functions to something other than `assert_*` (I've used `validate_*` in the past), I was just trying to remain consistent with naming convention in the project.

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+    Get the fieldnames of the specified metric class.
+
+    Args:
+        metric_class: A Metric class.
+
+    Returns:
+        A list of fieldnames.
+
+    Raises:
+        TypeError: If the given class is not a Metric.
```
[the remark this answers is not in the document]
msto: The `TypeError` is raised by `_assert_is_metric_class()`, and I should remove it from the doc. (I prefer to only document exceptions directly raised by the function)

I will remove the `Raises` block because the assertion is unreachable.

fgpyo/util/metric.py
```diff
@@ -466,3 +477,103 @@ def asdict(metric: Metric) -> Dict[str, Any]:
...
+
+    if dataclasses.is_dataclass(metric_class):
+        return [f.name for f in dataclasses.fields(metric_class)]
+    elif attr.has(metric_class):
+        return [f.name for f in attr.fields(metric_class)]
+    else:
+        assert False, "Unreachable"
+
+
+def _assert_file_header_matches_metric(
```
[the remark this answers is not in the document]
msto: I can add an assertion that the `path` is not `/dev/stdin`. Are there other streams that are likely to be represented as a `Path`?

fgpyo/util/metric.py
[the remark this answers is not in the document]
nh13: I think we should use it and fix it?

fgpyo/util/metric.py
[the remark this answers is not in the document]
nh13: Just streams I think

fgpyo/util/metric.py
clintval: This might be a good case for having a custom exception:

```python
class ValueAssertion(AssertionError, ValueError):
    pass
```

Then you would catch the exception with either `AssertionError` or `ValueError`.

I do like having the explicit branch with `if` since code coverage analysis will identify it as a covered/uncovered branch depending on actual test coverage. With single-line statements (e.g. `assert`), you don't get any branch information so it is easy to forget to test the exception paths.

fgpyo/util/metric.py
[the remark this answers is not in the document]
nh13: Bump

nh13 (approved):
One more item to address, then lgtm

Issue #137, opened 2024-06-12
Update `pyproject.toml`


msto:
- [x] Tooling configuration should be moved out of per-tool config files and into` pyproject.toml`
    - [x] flake8.cfg #140 
    - [x] mypy.ini
- [x] `dev-dependencies` [is deprecated](https://python-poetry.org/docs/managing-dependencies/) and should be replaced with `group.dev.dependencies`
- [x] Sphinx/doc dependencies should be moved into a `docs` group

PR #135, opened 2024-06-07
Move to mkdocs from Sphinx

nh13:
There are _a lot_ of changes required for this PR, most of them minor:

1. Update RST syntax to Markdown (e.g. URLs, anchors to clases/func, code blocks, lists)
2. Make the linter happy (lines too long with new URLs)
3. Consolidate tests (see #120)
4. Fixed a few other docstring warnings

PR #140, opened 2024-06-12
chore: Switch to ruff

### LINKED ISSUE #139: Move to `ruff`
I would like to make the transition because ruff includes a number of classes of lints that we have enabled by default in the Python template (isort, quotes, bugbear, mccabe complexity) that I would like to add to fgpyo, but would require adding plugins to support via flake8. In my experience ruff has largely been a seamless drop-in replacement for flake8, so I think the transition is low-risk, and makes more sense to do now rather than spending effort on adding flake8 plugins.

https://docs.astral.sh/ruff/rules/

Future issues should address the addition of new lint classes

msto:
Closes #139 

This PR removes `flake8`, `isort`, and `black` from the dependencies, and replaces them with `ruff`. The `ci/check.sh` script has been updated to use `ruff` in place of these tools.


pyproject.toml
```diff
@@ -35,12 +35,7 @@ typing_extensions = { version = ">=3.7.4", python = "<3.12" }
 docs = ["sphinx", "sphinx_rtd_theme"]
 
 [tool.poetry.dev-dependencies]
-black = ">=24.2.0"
-flake8 = [
-	{ version = ">=5.0.0", python = "<3.12.0" },
-	{ version = ">=6.1.0", python = ">=3.12.0" },
-]
-isort = ">=5.10.1"
+ruff = "0.4.8"
```
msto, line 38: #137

pyproject.toml
```diff
@@ -49,3 +44,15 @@ setuptools = ">=68.0.0"
...
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.ruff]
+line-length    = 100
+target-version = "py38"
+output-format  = "full"
+
+[tool.ruff.lint]
+select    = ["E", "F", "I", "W"]
```
msto, line 54: Additional classes to be added in subsequent issues/PRs. This PR just makes the transition to `ruff`

.github/workflows/pythonpackage.yml
```diff
@@ -55,31 +55,26 @@ jobs:
...
-    - name: Run isort
-      shell: bash
-      run: |
-        poetry run isort --force-single-line-imports --profile black fgpyo
-
     - name: Style checking
       shell: bash
       run: |
-        poetry run black --line-length 99 --check fgpyo
+        poetry run ruff format --check fgpyo
```
msto, line 61: `ruff format` replaces both `black` and `isort`

.github/workflows/pythonpackage.yml
```diff
@@ -55,31 +55,26 @@ jobs:
...
     - name: Run mypy
       shell: bash
       run: |
         poetry run mypy -p fgpyo --config=ci/mypy.ini
 
+    - name: Run pytest
+      shell: bash
+      run: |
+        poetry run python -m pytest --cov=fgpyo --cov-report=xml --cov-branch 
+
```
msto, line 77: I find formatting, linting, type-checking, and unit testing to be a better order of operations. 

1) it's loosely ordered by how long I expect each to take (put quicker steps first to fail fast, especially on GHA), 

2) it's the order in which I think about fixing issues with the code. It's easier to debug a test when you're confident in the type-checking, and type checks are easier to fix when you're not also wrangling with lints or style issues

nh13 (approved):
lgtm, probably good to get @tfenne's approval too

ci/check.sh
```diff
@@ -44,10 +44,9 @@ if [[ "$1" == "--check" ]]; then
 fi
 
 banner "Executing in conda environment ${CONDA_DEFAULT_ENV} in directory ${root}"
+run "Style Checking" "ruff format fgpyo"
+run "Linting"        "ruff check --fix fgpyo"
```
tfenne, line 48: I'm loosing track of the various PRs ... but will we want to include `tests` in this too?

tfenne (approved):

ci/check.sh
```diff
@@ -44,10 +44,9 @@ if [[ "$1" == "--check" ]]; then
...
 run "Unit Tests"     "python -m pytest -vv -r sx fgpyo"
-run "Import Sorting" "isort --force-single-line-imports --profile black fgpyo"
-run "Style Checking" "black --line-length 99 $black_extra_args fgpyo"
```
geoffjentry, line 49: It looks like `black_extra_args` is either now superfluous or should be repurposed

ci/check.sh
```diff
@@ -44,10 +44,9 @@ if [[ "$1" == "--check" ]]; then
 fi
 
 banner "Executing in conda environment ${CONDA_DEFAULT_ENV} in directory ${root}"
```
geoffjentry, line 46: This is preexisting, feel free to ignore. Looks like somewhere along the way this either got copypasta'd from another repo that actually defined `root` or the `root` definition was removed. Either way, `root` doesn't reference anything.

msto: thanks for catching!

ci/check.sh
msto: Yes, but it will have to wait until `tests/` gets refactored into a top level directory, which I think is happening in #135

PR #144, opened 2024-06-12
chore: fix B017 violations

msto:
https://docs.astral.sh/ruff/rules/assert-raises-exception/

> ### What it does
> 
> Checks for `assertRaises` and `pytest.raises` context managers that catch `Exception` or `BaseException`.
> 
> ### Why is this bad?
> 
> These forms catch every `Exception`, which can lead to tests passing even if, e.g., the code under consideration raises a `SyntaxError` or `IndentationError`.
> 
> Either assert for a more specific exception (builtin or custom), or use `assertRaisesRegex` or `pytest.raises(..., match=<REGEX>)` respectively.

### DIFF
--- a/fgpyo/tests/test_read_structure.py
+++ b/fgpyo/tests/test_read_structure.py
@@ -97,12 +97,18 @@ def test_read_structure_variable_once_and_only_once_last_segment_ok(
     assert ReadStructure.from_string(segments=string).segments == segments
 
 
-@pytest.mark.parametrize(
-    "string",
-    ["++M", "5M++T", "5M70+T", "+M+T", "+M70T"],
-)
-def test_read_structure_variable_once_and_only_once_last_segment_exception(string: str) -> None:
-    with pytest.raises(Exception):
+@pytest.mark.parametrize("string", ["+M+T", "+M70T"])
+def test_read_structure_rejects_variable_length(string: str) -> None:
+    """Variable length should only be permitted in the last segment of the read structure."""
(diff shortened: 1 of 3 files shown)


nh13 (approved):
Thank-you for doing this!

fgpyo/tests/test_read_structure.py
```diff
@@ -97,12 +97,18 @@ def test_read_structure_variable_once_and_only_once_last_segment_ok(
...
+    """Variable length should only be permitted in the last segment of the read structure."""
+    expected_msg = r"Variable length \(\+\) can only be used in the last segment"
+    with pytest.raises(AssertionError, match=expected_msg):
+        ReadStructure.from_string(segments=string)
+
+
+@pytest.mark.parametrize("string", ["++M", "5M70+T", "5M++T"])
+def test_read_structure_rejects_unknown_type(string: str) -> None:
+    """Segments with unknown type should be rejected."""
+    with pytest.raises(ValueError, match="Read structure segment had unknown type"):
```
tfenne, line 111: Looks good, but I'm curious if you could have done `pytest.raises(Union[AssertionError, ValueError])` or similar here?

tfenne (approved):
Thank you!

fgpyo/tests/test_read_structure.py
msto: You could, but I think it should be avoided for similar reasons. We want to test that the expected exception is being raised, and no other.

PR #145, opened 2024-06-12
chore: fix B007 violations

msto:
https://docs.astral.sh/ruff/rules/unused-loop-control-variable/

> ### What it does
> 
> Checks for unused variables in loops (e.g., for and while statements).
>
> ### Why is this bad?
> 
> Defining a variable in a loop statement that is never used can confuse readers.
> 
> If the variable is intended to be unused (e.g., to facilitate destructuring of a tuple or other object), prefix it with an underscore to indicate the intent. Otherwise, remove the variable entirely.


nh13 (approved):
Thank-you for doing this!

tfenne (approved):

PR #146, opened 2024-06-12
chore: fix B028 violations

msto:
https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

> ### What it does
> 
> Checks for `warnings.warn` calls without an explicit `stacklevel` keyword argument.
>
> ### Why is this bad?
> 
> The `warnings.warn` method uses a `stacklevel` of 1 by default, which limits the rendered stack trace to that of the line on which the warn method is called.
> 
> It's recommended to use a `stacklevel` of 2 or higher, give the caller more context about the warning.


nh13 (approved):
LGTM

tfenne (approved):

PR #147, opened 2024-06-13
chore: fix B904 violations

nh13:
https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

> What it does
> Checks for raise statements in exception handlers that lack a from clause.
> 
> Why is this bad?
> In Python, raise can be used with or without an exception from which the current exception is derived. This is known as exception chaining. When printing the stack trace, chained exceptions are displayed in such a way so as make it easier to trace the exception back to its root cause.
> 
> When raising an exception from within an except clause, always include a from clause to facilitate exception chaining. If the exception is not chained, it will be difficult to trace the exception back to its root cause.

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -224,8 +224,8 @@ class SamFileType(enum.Enum):
         ext = Path(path).suffix
         try:
             return next(iter([tpe for tpe in SamFileType if tpe.extension == ext]))
-        except StopIteration:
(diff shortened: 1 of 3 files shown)


msto (approved):
🙌

tfenne (approved):
If I'm honest I have more mixed feelings about this.  E.g. in `get_parser()` is it really helpful, or noisy, to know that the reason we couldn't parse a type is because we couldn't find a parser for that type in a dict?

PR #148, opened 2024-06-13
chore: fix B010 violations

nh13:
https://docs.astral.sh/ruff/rules/set-attr-with-constant/

> What it does
> Checks for uses of setattr that take a constant attribute value as an argument (e.g., setattr(obj, "foo", 42)).
> 
> Why is this bad?
> setattr is used to set attributes dynamically. If the attribute is defined as a constant, it is no safer than a typical property access. When possible, prefer property access over setattr calls, as the former is more concise and idiomatic.

### DIFF
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -386,7 +386,7 @@ def _get_parser(
     # Set the name that the user expects to see in error messages (we always
     # return a temporary partial object so it's safe to set its __name__).
     # Unions and Literals don't have a __name__, but their str is fine.
-    setattr(parser, "__name__", getattr(type_, "__name__", str(type_)))
+    setattr(parser, "__name__", getattr(type_, "__name__", str(type_)))  # noqa: B010
     return parser
 
 


nh13:
Looks like we need to ignore this one

msto:
Yeah. 

I hadn't seen this error code before, and I had the thought that we might want to ignore it generally, not just here. 

However, I then had the realization that it's basically the linter politely asking if you're _sure_ you want to monkey-patch a new attribute onto this. And I don't mind the extra friction on something that should be generally but not dogmatically avoided. 

(tbh I think the `ruff` docs could maybe be made clearer that this is really a lint for monkey-patching)

Curious what you and @tfenne think

msto (approved):

nh13:
Agreed on wanting this, as I want to avoid setattr as much as possible. It turns out this is a special case where we want it, as the recommended solution (set it directly) doesn’t work for this type of attribute (I think)

tfenne (approved):

PR #149, opened 2024-06-18
chore: ignore existing C901 violations

msto:
https://docs.astral.sh/ruff/rules/complex-structure/

> ### Why is this bad?
> 
> The McCabe complexity of a function is a measure of the complexity of the control flow graph of the function. It is calculated by adding one to the number of decision points in the function. A decision point is a place in the code where the program has a choice of two or more paths to follow.
> 
> Functions with a high complexity are hard to understand and maintain.

### DIFF
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -311,7 +311,7 @@ def dict_parser(
     return functools.partial(dict_parse)
 
 
-def _get_parser(
+def _get_parser(  # noqa: C901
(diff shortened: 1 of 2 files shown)


msto:
IMO `get_parser` and `format_value` might be able to be simplified, and we should consider at least examining their tests for sufficient coverage, but I think that work can be deferred while we get the checks in

nh13 (approved):
LGTM

msto:
Merging [as Tim approved in Slack](https://fulcrumgenomics.slack.com/archives/C05RSE14QM6/p1718712229521609)

PR #150, opened 2024-06-20
ci: update code-cov in GHA to use a token

nh13:


geoffjentry (approved):
Despite the draft status, it does look like it succeeded in GHA so :+1: from me

Issue #112, opened 2024-05-06
Consider renaming modules which shadow stdlib modules


msto:
Consider renaming the `types`, `inspect`, `logging`, and `string` modules to avoid shadowing stdlib

PR #151, opened 2024-06-26
feat: Add `find_illumina_fastq`

msto:
I'm introducing a function to search a directory for a FASTQ belonging to a given sample, regardless of extension or [name formatting](https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm).

I've added it to a new submodule, `util.files`, but let me know if there's a more appropriate name or you would like to see it added to an existing module

### DIFF
--- a/fgpyo/platform/illumina.py
+++ b/fgpyo/platform/illumina.py
@@ -33,7 +33,7 @@ def find_illumina_fastq(
     Args:
         fastq_dir: The directory to search.
         sample_id: The sample ID. The function will search for a FASTQ prefixed with this string.
-        number: The read number.
+        read_number: The read number.
         recursive: If True, search subdirectories.
 
     Returns:
--- a/fgpyo/platform/illumina.py
+++ b/fgpyo/platform/illumina.py
@@ -1,0 +1,65 @@
+from enum import Enum
+from enum import unique
+from glob import glob
+from pathlib import Path
+from typing import List
+
+
+@unique
+class IlluminaReadNumber(Enum):
+    """The read number of an Illumina FASTQ."""
+
+    R1 = "R1"
+    R2 = "R2"
+    I1 = "I1"
+    I2 = "I2"
+
+
+FASTQ_EXTENSIONS = [".fq", ".fastq", ".fq.gz", ".fastq.gz"]
+
+
+def find_illumina_fastq(
+    fastq_dir: Path,
+    sample_id: str,
+    read_number: IlluminaReadNumber,
+    recursive: bool = True,
+) -> Path:
+    """
+    Search a directory for a FASTQ belonging to the specified sample.
+
+    The function will search for a FASTQ prefixed with `sample_id`, including the specified
+    `read_number`, and suffixed with `.fastq`, `.fq`, `.fastq.gz`, or `.fq.gz`.
+
+    Args:
+        fastq_dir: The directory to search.
+        sample_id: The sample ID. The function will search for a FASTQ prefixed with this string.
+        number: The read number.
+        recursive: If True, search subdirectories.
+
+    Returns:
+        The path to the discovered FASTQ file.
+
+    Raises:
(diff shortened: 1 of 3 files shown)


fgpyo/util/files.py
```diff
@@ -0,0 +1,65 @@
+from enum import Enum
+from enum import unique
+from glob import glob
+from pathlib import Path
+from typing import List
+
+
+@unique
+class IlluminaReadNumber(Enum):
```
nh13, line 9: Can use `StrEnum` and `auto()` below

nh13 (changes requested):
Should we support FASTQs split across lanes?

msto:
Probably. 🙂 

And update the validation to enforce that there may be only one fastq for the specified sample and read number per lane (e.g. a sample can't have two R2 fastqs from the same lane)

As I'm writing this, this should maybe be made optional as well, so the user can find all fastqs associated with a sample from multiple sequencing runs

tfenne:
Can take a closer look later, but I would like to veto the use of `util` and related names.  

I would probably suggest putting this into a `vendors.illumina` module as I think if I were looking for such a thing, the primary search route would be "looking for illumina specific stuff".

msto:
Thanks @tfenne. The `util` module already exists, and includes the `inspect`, `logging`, `metric`, `string`, and `types` submodules. I think refactoring the package to remove the existing `util` module falls outside the scope of this PR, but perhaps could be addressed as part of #112

nh13 (approved):
LGTM, can you create `fgpyo/vendor/illumina.py` and move this in there as per @tfenne 's comments: https://github.com/fulcrumgenomics/fgpyo/pull/151#issuecomment-2192098876

msto:
`vendor` is typically reserved for vendored copies of third-party packages.

e.g. in `vendor/Illumina` I might expect to find a (possibly patched) copy of [BeadArrayFiles](https://github.com/Illumina/BeadArrayFiles).

Could the two of you please agree on an alternative name and let me know?

tfenne:
I wasn't aware that `vendor` in the module hierarchy has any special meaning.  That feels the most natural to me as we often talk about "sequencing vendors".  The only other term I can think of would be `platform` - thoughts @nh13 ?

nh13:
I would suggest `platform` as well, given it's the synonym for `vendor` that's used in the SAM spec.  @msto does that work?

msto:
Works for me!

fgpyo/platform/illumina.py
```diff
@@ -0,0 +1,65 @@
...
+
+FASTQ_EXTENSIONS = [".fq", ".fastq", ".fq.gz", ".fastq.gz"]
+
+
+def find_illumina_fastq(
+    fastq_dir: Path,
+    sample_id: str,
+    read_number: IlluminaReadNumber,
+    recursive: bool = True,
+) -> Path:
```
msto, line 26: @nh13 Should this be updated to return `list[Path]`, such that if a sample was split across lanes the function returns all FASTQs from the sample?

nh13: Yes please.  Also consider if the read is chunked into multiple parts due to file size, which I have seen.

fgpyo/platform/illumina.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Search a directory for a FASTQ belonging to the specified sample.
+
+    The function will search for a FASTQ prefixed with `sample_id`, including the specified
+    `read_number`, and suffixed with `.fastq`, `.fq`, `.fastq.gz`, or `.fq.gz`.
+
+    Args:
+        fastq_dir: The directory to search.
+        sample_id: The sample ID. The function will search for a FASTQ prefixed with this string.
+        number: The read number.
```
nh13, line 36:
```suggestion
        read_number: The read number.
```

PR #154, opened 2024-06-26
- Added __getitem__ and __len__ to the Cigar class, pulling the answer…

yfarjoun:
… from the .elements field

- Added tests and documentation
- fixes #153

PR #155, opened 2024-06-27
feat: add len, iter, and getitem dunder methods to Cigar

msto:
Per [Yossi's suggestion in Slack yesterday](https://fulcrumgenomics.slack.com/archives/C04BQ9RMJ2G/p1719424987601389?thread_ts=1719352429.511779&cid=C04BQ9RMJ2G), I believe this will make `Cigar` more ergonomic

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -531,6 +531,15 @@ class Cigar:
         else:
             return "*"
 
+    def __len__(self) -> int:
+        return len(self.elements)
+
+    def __iter__(self) -> Iterator[CigarElement]:
+        return iter(self.elements)
+
+    def __getitem__(self, index: int) -> CigarElement:
+        return self.elements[index]
+
     def reversed(self) -> "Cigar":
         """Returns a copy of the Cigar with the elements in reverse order."""
         return Cigar(tuple(reversed(self.elements)))
--- a/fgpyo/sam/tests/test_sam.py
+++ b/fgpyo/sam/tests/test_sam.py
@@ -372,3 +372,25 @@ def test_calc_edit_info_with_aligned_Ns() -> None:
     assert info.deletions == 0
     assert info.deleted_bases == 0
     assert info.nm == 5
+
+
+@pytest.mark.parametrize("cigarstring", ["10M5I4M", "10M5D4M", "10M40S", "*"])
+def test_cigar_length(cigarstring: str) -> None:
+    """CIGAR length should be the number of elements."""
+    cigar = Cigar.from_cigarstring(cigarstring)
+    assert len(cigar) == len(cigar.elements)
+
+
+@pytest.mark.parametrize("cigarstring", ["10M5I4M", "10M5D4M", "10M40S", "*"])
+def test_cigar_get_item(cigarstring: str) -> None:
+    """CIGAR elements should be accessible by index."""
+    cigar = Cigar.from_cigarstring(cigarstring)
+    for i in range(len(cigar)):
+        assert cigar[i] == cigar.elements[i]
+
+
+@pytest.mark.parametrize("cigarstring", ["10M5I4M", "10M5D4M", "10M40S", "*"])
+def test_cigar_iterable(cigarstring: str) -> None:
+    """Iterating over a CIGAR should yield its elements."""
+    cigar = Cigar.from_cigarstring(cigarstring)
+    assert [e for e in cigar] == list(cigar.elements)


fgpyo/sam/__init__.py
```diff
@@ -531,6 +531,15 @@ def __str__(self) -> str:
...
 
+    def __len__(self) -> int:
+        return len(self.elements)
+
+    def __iter__(self) -> Iterator[CigarElement]:
+        return iter(self.elements)
+
+    def __getitem__(self, index: int) -> CigarElement:
+        return self.elements[index]
+
```
nh13, line 542: I love this.  Can we add to the module-level examples to demonstrate this new functionality?

nh13 (changes requested):
One minor request.

clintval (commented):
Duplicate? https://github.com/fulcrumgenomics/fgpyo/pull/154

msto:
I'll update to only add `__iter__` and `__len__`

fgpyo/sam/__init__.py
```diff
@@ -531,6 +531,15 @@ def __str__(self) -> str:
         else:
             return "*"
 
+    def __len__(self) -> int:
+        return len(self.elements)
```
tfenne, line 535: See https://github.com/fulcrumgenomics/fgpyo/pull/154#discussion_r1655564488

I think adding a len() support to CIGAR is a bad idea and we have explicitly chosen to not do this in the past.

nh13: bump

fgpyo/sam/tests/test_sam.py
```diff
@@ -372,3 +372,25 @@ def test_calc_edit_info_with_aligned_Ns() -> None:
     assert info.deletions == 0
     assert info.deleted_bases == 0
     assert info.nm == 5
+
+
+@pytest.mark.parametrize("cigarstring", ["10M5I4M", "10M5D4M", "10M40S", "*"])
+def test_cigar_length(cigarstring: str) -> None:
+    """CIGAR length should be the number of elements."""
+    cigar = Cigar.from_cigarstring(cigarstring)
+    assert len(cigar) == len(cigar.elements)
```
nh13, line 381:
```suggestion
```

fgpyo/sam/__init__.py
```diff
@@ -531,6 +531,15 @@ def __str__(self) -> str:
...
+
```
nh13, line 536:
```suggestion
```

fgpyo/sam/__init__.py
msto: I disagree. I would like to propose that `Cigar` subclass `collections.abc.Sequence`, as we have already introduced `__get_item__` in #154.

https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

The related dunder methods being introduced - `__get_item__` and `__iter__` - are both with respect to the underlying list of `CigarElement`s, and I think it makes sense for `__len__` to be defined similarly.

It is _less_ intuitive to me to have an indexable and iterable object that is not `Sized`.

msto: Put another way - if you have a `Collection` or `Sequence` of `Sized` elements, I don't think the user would expect the length of that collection to be the sum of the lengths of its elements. I'm not sure why that is being raised as an argument in opposition here

tfenne: Because we're representing a "cigar string" as an object here.  Conduct a straw-poll and ask some folks without any pre-amble or coaching "what do you think the length of `Cigar("76M")` is?" and I'd be willing to bet that you'll get a diversity of answers and that "1" will not be the most common.

clintval: From the peanut gallery: I would have wanted the answer to be `76` since a Cigar is a [run-length encoded](https://en.wikipedia.org/wiki/Run-length_encoding) sequence and I'd want the full length of what it represents, not what it is compressed.

For now I also vote avoiding the issue by leaving `__len__` unimplemented and adding properties that are more clear about the different "lengths".

But if I had to choose, I'd want `76` since in my mind these two objects should have the same length and it should be the uncompressed length: `Cigar("2M")` vs `Cigar("1M1M")`. They are representations of alignment operators first, and sequence-like second.

msto: Then I do not think we should have implemented `__get_item__` and we should revert #154 .

Flipping the question around, what is the second element in "2M"?

clintval: Perhaps I'm in the minority but I would have expected `__get_item__` to index into the uncompressed representation (e.g. what's at the second position of the alignment represented by the cigar?). Indexing into the second position of `Cigar("2M")` would return a match operator and indexing into the second position of `Cigar("3S4M2S")` would return a soft-clip operator. If I wanted to index into the second of the compressed elements, I'd explicitly use `cigar.elements[1]`.

Looks like we avoided this in `fgpyo` since `Cigar` is an iterable but is not sized or indexable: 

https://github.com/fulcrumgenomics/fgbio/blob/9a4b01d45a38c4990bff2b011cba7abdde134190/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala#L165

I'm interested in what others think. For me, thinking more carefully about this #154 seems like it could be a footgun for the same reasons as implementing `__len__()`.

msto: I think these need to be defined consistently, or not at all.

If we are defining `__get_item__` as an index into the elements, there should be a `__length__` defined with respect to the number of elements.

If we think the length of a CIGAR should be its run length, then `__get_item__` should be an index into the uncompressed sequence of operators (e.g. a single `M` operator)

If we cannot agree whether `Cigar` should be a collection of `CigarElement` or a collection of single base operators, then neither `__get_item__` nor `__length__` should be defined.

tfenne: I'd be happy with that outcome.  As @nh13 said elsewhere, I've never found it burdensome to have to do `cig.elements[1]` instead of `cig[1]`, and it's clear that `elements` is a sequence of `CigarElement`.

I'm ambivalent about having `__iter__`.  There are plenty of reasonable examples of providing iterators without knowing the number of items the iterator will produce.  On the other hand, it's also just not that hard to `for elem in cigar.elements` either.

PR #156, opened 2024-07-01
Feat: add gc_content() calculation method to fgpyo/sequence.py

emmcauley:
Add method to calculate fraction GC content in a string of bases. Corresponding unit tests also added.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,8 +10,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import os
-import sys
+import sys, os, glob
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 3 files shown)


fgpyo/sequence.py
```diff
@@ -70,3 +70,12 @@ def reverse_complement(bases: str) -> str:
...
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
+
+
+def gc_content(bases: str) -> float:
+    """Calculates the fraction of G and C bases in a sequence."""
+    if bases:
+        gc_count = sum(1 if (ch == "C" or ch == "G") else 0 for ch in bases)
+        return gc_count / len(bases)
+    else:
+        return 0
```
nh13, line 81: 1. Reminder to use `if len(str) == 0:` instead of the implicit conversion `if str:`.   This is included in my suggestion below.
2. Do we handle lower-case bases?  This is not included in my suggestion below.
```suggestion
    if len(bases) = 0:
        return 0
    gc_count = sum(1 if (ch == "C" or ch == "G") else 0 for ch in bases)
    return gc_count / len(bases)
```

fgpyo/tests/test_sequence.py
```diff
@@ -15,3 +16,11 @@ def test_reverse_complement() -> None:
...
     with pytest.raises(KeyError):
         reverse_complement("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+
+def test_gc_content() -> None:
+    assert gc_content("ATATATATTATATA") == 0.0
+    assert gc_content("GCGCGCGCGCGCG") == 1.0
+    assert gc_content("ACGT") == 0.5
+    assert gc_content("") == 0.0
+    assert gc_content("ACGTN") == 0.4
```
nh13, line 26: 1. Can you make these parameterized with pytest?
2. After (1), then can you test both upper and lower case (you can just have two asserts in the test method, one that upper-cases explicitly, and the other that lower-cases explicitly)

docs/conf.py
```diff
@@ -10,7 +10,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys, os, glob
+import os
```
nh13, line 13: where are these changes coming from?  I think these are unrelated.  Can you either remove, or make a separate PR?

nh13 (changes requested):

docs/conf.py
emmcauley: I think they resulted from a `ruff format` run; it should be fixed now

fgpyo/tests/test_sequence.py
emmcauley: 1. parameterized with pytest
2. unit test includes mixed base cases

fgpyo/sequence.py
emmcauley: thank you! Suggestion implemented and I use .upper() on the input sequence, so now we should be handling lower and upper cases appropriately

fgpyo/sequence.py
```diff
@@ -70,3 +70,12 @@ def reverse_complement(bases: str) -> str:
...
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
+
+
+def gc_content(bases: str) -> float:
+    """Calculates the fraction of G and C bases in a sequence."""
+    if len(bases) == 0:
+        return 0
+    else:
+        gc_count = sum(base == "C" or base == "G" for base in bases.upper())
+        return gc_count / len(bases)
```
nh13, line 81:
```suggestion
    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
    return gc_count / len(bases)
```

nh13 (changes requested):
Minor suggestion that avoids having to traverse `bases` twice (once to uppercase, once to compare).  Also avoids relying summing an iterator of bools, which seems nonsensical (bool is not a numeric), but that's python

nh13 (approved):
LGTM, merge when ready:
punch it

PR #158, opened 2024-07-05
feat: speed up longest_homopolymer

nh13:


yfarjoun (approved):

emmcauley (approved):
I like it!

fgpyo/sequence.py
```diff
@@ -75,16 +75,15 @@ def reverse_complement(bases: str) -> str:
 def longest_hp_length(bases: str) -> int:
     """Calculates the length of the longest homopolymer in the input sequence."""
     max_hp = 0
-    for i in range(0, len(bases)):
+    i = 0
+    # NB: if we have found a homopolymer of length `max_hp`, then we do not need
+    # to examine the last `max_hp` bases since we'll never find a longer one.
+    while i < len(bases) - max_hp:
```
tfenne, line 81: If going for speed, why not also cache `len(bases)`?

tfenne (approved):

fgpyo/sequence.py
nh13: It’s O(1) so as fast as an access, so really isn’t beneficial.

Issue #136, opened 2024-06-12
Move code out of `__init__.py` and into `module.py` files


msto:
It's more conventional to reserve `__init__.py` for imports and namespace management (i.e. with `__all__`).

If this is implemented, there should likely be (approximately) one PR per module

- [ ] collections
- [ ] fasta
- [ ] fastx
- [ ] io
- [ ] sam
- [ ] util
- [ ] vcf

PR #159, opened 2024-07-11
Add a lazy is_sorted function to collections

clintval:


msto (approved):

fgpyo/collections/__init__.py
```diff
@@ -47,29 +64,60 @@
...
+
+class _SupportsLessThanOrEqual(Protocol):
+    """A structural type for objects that support less-than-or-equal comparison."""
+
+    def __le__(self, other: Any) -> bool: ...
+
+
 IterType = TypeVar("IterType")
 
+_LessThanOrEqualType = TypeVar("_LessThanOrEqualType", bound=_SupportsLessThanOrEqual)
```
nh13, line 118: Is it appropriate for this type be "private" when it is used as a type parameter in a "public" method below?

nh13 (changes requested):

fgpyo/collections/__init__.py
clintval: Can do!

I notice fgpyo doesn't use `__init__.py` for public namespace definition which makes defining a public API trickier. For example, the function `pairwise` in this PR is now exposed as our public API instead of just as an import or a helper. I do see Matt is tracking that issue here, though:

- https://github.com/fulcrumgenomics/fgpyo/issues/136

nh13: Agreed, one way to manage what is public versus private is #136, but another way to solve it is to the way we're doing here.  Moving things out of `__init__.py` into submodules doesn't solve this, since the methods/classes can be public there, but it does help with namespace issues.  Lets move that discussion to #136 since the namespace issue is compelling, and any new method can always be private like we have here for pairwise.

nh13 (approved):
lgtm

PR #160, opened 2024-07-15
feat: adds classes and methods for sequence dictionaries

### LINKED ISSUE #152: Feature request: better modeling of a pysam sequence dictionary
Rough sketch of functionality:

* class that takes the `pysam` list of dictionaries `reader.header["SQ"]` and stores something similar to fgbio's [`SequenceMetadata`](https://github.com/fulcrumgenomics/fgbio/blob/main/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala)
* method that takes a FASTA and returns the path to the sequence dictionary (simple as `Path(fasta).with_suffix(".dict")`?)
* method that takes a FASTA and returns the sequence dictionary class

nh13:
Fixes #152

### DIFF
--- a/fgpyo/fasta/sequence_dictionary.py
+++ b/fgpyo/fasta/sequence_dictionary.py
@@ -1,0 +1,501 @@
+"""
+# Classes for representing sequencing dictionaries.
+
+## Examples of building and using sequence dictionaries
+
+Building a sequence dictionary from a `pysam.AlignmentHeader`:
+
+```python
+>>> import pysam
+>>> from fgpyo.fasta.sequence_dictionary import SequenceDictionary
+>>> sd: SequenceDictionary
+>>> with pysam.AlignmentFile("./fgpyo/sam/tests/data/valid.sam") as fh:
+    ...    sd = SequenceDictionary.from_sam(header=fh.header)
+...
+>>> print(sd)
+@SQ	SN:chr1	LN:101
+@SQ	SN:chr2	LN:101
+@SQ	SN:chr3	LN:101
+@SQ	SN:chr4	LN:101
+@SQ	SN:chr5	LN:101
+@SQ	SN:chr6	LN:101
+@SQ	SN:chr7	LN:404
+@SQ	SN:chr8	LN:202
+```
+
+Query based on index:
+
+```python
+>>> print(sd[3])
+@SQ	SN:chr4	LN:101
+```
+
+Query based on name:
+
+```python
+>>> print(sd["chr6"])
+@SQ	SN:chr6	LN:101
+```
+
+Add, get, and delete attributes:
+
+```python
+>>> meta = sd[0]
+>>> print(meta)
+@SQ	SN:chr1	LN:101
+>>> meta[Keys.ASSEMBLY] = "hg38"
+>>> print(meta))
+@SQ	SN:chr1	LN:101	AS:hg38
+>>> meta.get(Keys.ASSEMBLY)
+"hg38"
+>>> meta.get(Keys.SPECIES) is None
+True
+>>> Keys.MD5 in meta
+False
+>>> del meta[Keys.ASSEMBLY]
+>>> print(meta)
+@SQ	SN:chr1	LN:101
+```
+
+Get a sequence based on one of its aliases
+
+```python
+>>> meta[Keys.ALIASES] = "foo,bar,car"
+>>> sd = SequenceDictionary(infos=[meta] + sd.infos[1:])
+>>> print(sd)
+@SQ	SN:chr1	LN:101	AN:foo,bar,car
+@SQ	SN:chr2	LN:101
+@SQ	SN:chr3	LN:101
+@SQ	SN:chr4	LN:101
+@SQ	SN:chr5	LN:101
+@SQ	SN:chr6	LN:101
+@SQ	SN:chr7	LN:404
+@SQ	SN:chr8	LN:202
+>>> print(sd["chr1"])
+@SQ	SN:chr1	LN:101	AN:foo,bar,car
+>>> print(sd["bar"])
+@SQ	SN:chr1	LN:101	AN:foo,bar,car
+```
+
+Create a `pysam.AlignmentHeader` from a sequence dictionary:
+
+```python
+>>> sd.to_sam_header()
+<pysam.libcalignmentfile.AlignmentHeader object at 0x10e93f5f0>
+>>> print(sd.to_sam_header())
+@HD	VN:1.5
+@SQ	SN:chr1	LN:101	AN:foo,bar,car
+@SQ	SN:chr2	LN:101
+@SQ	SN:chr3	LN:101
+@SQ	SN:chr4	LN:101
+@SQ	SN:chr5	LN:101
+@SQ	SN:chr6	LN:101
+@SQ	SN:chr7	LN:404
+@SQ	SN:chr8	LN:202
+```
+
+Create a `pysam.AlignmentHeader` from a sequence dictionary with extra header items:
+
+```python
+>>> sd.to_sam_header(
+...     extra_header={"RG": [{"ID": "A", "LB": "a-library"}, {"ID": "B", "LB": "b-library"}]}
+... )
+<pysam.libcalignmentfile.AlignmentHeader object at 0x10e93fe30>
+>>> print(sd.to_sam_header(
+...     extra_header={"RG": [{"ID": "A", "LB": "a-library"}, {"ID": "B", "LB": "b-library"}]}
+... ))
+@HD	VN:1.5
+@SQ	SN:chr1	LN:101	AN:foo,bar,car
+@SQ	SN:chr2	LN:101
+@SQ	SN:chr3	LN:101
+@SQ	SN:chr4	LN:101
+@SQ	SN:chr5	LN:101
+@SQ	SN:chr6	LN:101
+@SQ	SN:chr7	LN:404
+@SQ	SN:chr8	LN:202
+@RG	ID:A	LB:a-library
+@RG	ID:B	LB:b-library
+```
+"""
+
+import copy
+import itertools
+import re
+import sys
+from dataclasses import dataclass
+from dataclasses import field
+from dataclasses import replace
+from enum import unique
+from typing import Any
+from typing import Dict
+from typing import Iterator
+from typing import List
+from typing import Mapping
+from typing import MutableMapping
+from typing import Optional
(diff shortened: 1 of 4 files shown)


fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+        self_m5 = self.md5
+        other_m5 = other.md5
+        if self_m5 is None or other_m5 is None:
+            return True
+        else:
+            return self_m5 == other_m5
+
+    def to_sam(self) -> Dict[str, Any]:
+        """Converts the sequence metadata to a dictionary equivalent to one item in the
+        list of sequences from `pysam.AlignmentHeader#to_dict()`."""
```
tfenne, line 301: Shouldn't this be:

```suggestion
        list of sequences from `pysam.AlignmentHeader#to_dict()["SD"]`."""
```

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+        return self.attributes[key]
+
+    def __setitem__(self, key: str, value: str) -> None:
+        self.attributes[key] = value
+
+    def __delitem__(self, key: str) -> None:
+        del self.attributes[key]
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self.attributes)
```
tfenne, line 340: I feel like these should probably be augmented to also handle `SN` and `LN`.
For get/set it should route access to the properties.
For del I assume it should fail.
for iter I think it should probably chain iterators?

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+            extra_header: a dictionary of extra values to add to the header, None otherwise.  See
+                          `:~pysam.AlignmentHeader` for more details.
+        """
+        header_dict: Dict[str, Any] = {
+            "HD": {"VN": "1.5"},
+            "SQ": [meta.to_sam() for meta in self.infos],
+        }
+        if extra_header is not None:
+            header_dict = {**header_dict, **extra_header}
+        return pysam.AlignmentHeader.from_dict(header_dict=header_dict)
```
tfenne, line 401: I think I would question here if this is the most useful thing to do.  I think it might be nicer to return just the dict generated by `[meta.to_sam() for meta in self.infos]` so it can be inserted into an existing sam header?

tfenne: Or perhaps this becomes `to_sam_file_header()`, and we also have a function that does just the above?

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+        header: Union[pysam.AlignmentHeader, List[Dict[str, Any]]],
+    ) -> "SequenceDictionary":
+        if isinstance(header, pysam.AlignmentHeader):
+            return SequenceDictionary.from_sam(header=header.to_dict()["SQ"])
+
+        infos: List[SequenceMetadata] = [
+            SequenceMetadata.from_sam(meta=meta, index=index) for index, meta in enumerate(header)
+        ]
+
+        return SequenceDictionary(infos=infos)
```
tfenne, line 422: Similarly I think this should probably work from `sam.header["SD"]`.

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+
+    # FIXME: mypyp doesn't like these
+    # @overload
+    # def __getitem__(self, key: str) -> SequenceMetadata: ...
+    #
+    # @overload
+    # def __getitem__(self, index: int) -> SequenceMetadata: ...
+
+    def __getitem__(self, key: Union[str, int]) -> SequenceMetadata:
+        return self._dict[key] if isinstance(key, str) else self.infos[key]
```
tfenne, line 432: I wonder if the `isinstance` has much overhead if e.g. calling this in a big loop?  I'd be tempted to add separate functions (in addition?) to access specifically by name and index.

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+
+    def __iter__(self) -> Iterator[str]:
+        return iter(self._dict)
```
tfenne, line 435: I would have expected iteration to go over the values ... but I can see why you did this.  I'm not sure what to suggest.

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,441 @@
...
+
+    def __len__(self) -> int:
+        return len(self.infos)
+
+    def __str__(self) -> str:
+        return "\n".join(f"{info}" for info in self.infos)
```
tfenne, line 441: Would be nice to wrap the whole in square brackets?  E.g.

```
return "[" + ("\n,".join(f"{info}" for info in self.infos) + "]"
```

tfenne (commented):
Mostly 👍 but with a few thoughts/questions.

fgpyo/fasta/sequence_dictionary.py
nh13, line 301:
```suggestion
        list of sequences from `pysam.AlignmentHeader#to_dict()["SQ"]`."""
```

fgpyo/fasta/sequence_dictionary.py
nh13: Given I'd like to port over fgbio faithfullly, this would be breaking behavior [relative to fgbio](https://github.com/fulcrumgenomics/fgbio/blob/fc8e1c6a442a7609e7727c46b3fc377e0de05c41/src/main/scala/com/fulcrumgenomics/fasta/SequenceDictionary.scala#L159).  Skipping for now.

fgpyo/fasta/sequence_dictionary.py
nh13: Done

fgpyo/fasta/sequence_dictionary.py
nh13: I think it does.  The `isinstance` checks if it's a `pysam.AlignmentHeader`, and if not, it assumes it's a `header.to_dict()["SQ"]`...  I'll update the docs

fgpyo/fasta/sequence_dictionary.py
nh13: done

fgpyo/fasta/sequence_dictionary.py
nh13: Nah, it's nice this way since it prints a valid SAM header text.  I don't prefer to have square brackets surrounding that.

fgpyo/fasta/sequence_dictionary.py
tfenne: I think this is the only sticking point - on reflection I'd say the fgbio impl is wrong.  Without reference to _implementation_ why would you expect iteration and other methods to _only_ work for optional attributes?  Including well-defined ones like `MD`, but not `LN`?

nh13: I'll fix it here and in fgprimer so they both agree.

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,480 @@
...
+        attributes = copy.deepcopy(meta)
+        del attributes[Keys.SEQUENCE_NAME]
+        del attributes[Keys.SEQUENCE_LENGTH]
+        return SequenceMetadata(name=name, length=length, index=index, attributes=attributes)
+
+    def __getitem__(self, key: Union[Keys, str]) -> Any:
+        if key == Keys.SEQUENCE_NAME.value:
+            return self.name
+        elif key == Keys.SEQUENCE_LENGTH.value:
+            return f"{self.length}"
```
nh13, line 338: @tfenne I'm not in love with this, but it is consistent with the pattern that accessing via named properties returns an appropriately typed value, whereas using this get method always returns a string.

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,502 @@
...
+    @property
+    def description(self) -> Optional[str]:
+        return self.get(Keys.DESCRIPTION)
+
+    @property
+    def topology(self) -> Optional[Topology]:
+        value = self.get(Keys.TOPOLOGY)
+        if value is None:
+            return None
+        return Topology[value]
```
tfenne, line 309:
```suggestion
        return None if value is None else Topology[value]
```

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,502 @@
...
+            raise ValueError(f"'{Keys.SEQUENCE_NAME}' should not given in the list of attributes")
+        if Keys.SEQUENCE_LENGTH in self.attributes:
+            raise ValueError(f"'{Keys.SEQUENCE_LENGTH}' should not given in the list of attributes")
+
+    @property
+    def aliases(self) -> List[str]:
+        """The aliases (not including the primary) name"""
+        if Keys.ALIASES not in self.attributes:
+            return []
+        return self.attributes[Keys.ALIASES].split(",")
```
tfenne, line 259:
```suggestion
        aliases = self.attributes.get(Keys.ALIASES)
        return [] if aliases is None else aliases.split(",")
```

To match the form used for `topology`

fgpyo/fasta/sequence_dictionary.py
```diff
@@ -0,0 +1,502 @@
...
+            raise KeyError(f"Cannot set '{key}' on SequenceMetadata with name '{self.name}'")
+        self.attributes[key] = value
+
+    def __delitem__(self, key: Union[Keys, str]) -> None:
+        if key == Keys.SEQUENCE_NAME or key == Keys.SEQUENCE_LENGTH:
+            raise KeyError(f"Cannot delete '{key}' on SequenceMetadata with name '{self.name}'")
+        del self.attributes[key]
+
+    def __iter__(self) -> Iterator[Union[Keys, str]]:
+        return iter([Keys.SEQUENCE_NAME, Keys.SEQUENCE_LENGTH] + list(self.attributes))
```
tfenne, line 374: I think this is preferable:

```python
import itertools
itertools.chain(iter((ys.SEQUENCE_NAME, Keys.SEQUENCE_LENGTH)), iter(self.attributes))
```

tfenne (approved):
A few minor comment, but good to go.

PR #162, opened 2024-07-17
feat: add methods for computing run lengths of repeats in a sequence

nh13:
The `longest_dinucleotide_run_length` method is optimized for dinucleotides, whereas the `longest_multinucleotide_run_length` method is general for any repeat of length N.

### DIFF
--- a/fgpyo/sequence.py
+++ b/fgpyo/sequence.py
@@ -160,29 +160,29 @@ def longest_hp_length(bases: str) -> int:
     return longest_homopolymer_length(bases=bases)
 
 
-def longest_homopolymer_length(bases: str, min_length: int = 0) -> int:
+def longest_homopolymer_length(bases: str) -> int:
     """Calculates the length of the longest homopolymer in the input sequence.
 
     Args:
         bases: the bases over which to compute
-        min_length: do not search for homopolymers of this length or shorter
 
     Return:
-        the length of the longest homopolymer, or zero if not at least the minimum length
+        the length of the longest homopolymer
     """
+    cur_length: int = 0
     i = 0
     # NB: if we have found a homopolymer of length `min_hp`, then we do not need
     # to examine the last `min_hp` bases since we'll never find a longer one.
     bases_len = len(bases)
-    while i < bases_len - min_length:
+    while i < bases_len - cur_length:
         base = bases[i].upper()
         j = i + 1
         while j < bases_len and bases[j] == base:
             j += 1
-        min_length = max(min_length, j - i)
+        cur_length = max(cur_length, j - i)
         # skip over all the bases in the current homopolymer
         i = j
-    return min_length
+    return cur_length
 
 
 def longest_dinucleotide_run_length(bases: str) -> int:
--- a/fgpyo/sequence.py
+++ b/fgpyo/sequence.py
@@ -148,29 +148,41 @@ def levenshtein(string1: str, string2: str) -> int:
     return matrix[0][0]
 
 
-def longest_homopolymer_length(bases: str, min_hp: int = 0) -> int:
+def longest_hp_length(bases: str) -> int:
     """Calculates the length of the longest homopolymer in the input sequence.
 
     Args:
         bases: the bases over which to compute
-        min_hp: do not search for homopolymers of this length or shorter
 
     Return:
         the length of the longest homopolymer
+    """
+    return longest_homopolymer_length(bases=bases)
+
+
+def longest_homopolymer_length(bases: str, min_length: int = 0) -> int:
+    """Calculates the length of the longest homopolymer in the input sequence.
+
+    Args:
+        bases: the bases over which to compute
+        min_length: do not search for homopolymers of this length or shorter
+
+    Return:
+        the length of the longest homopolymer, or zero if not at least the minimum length
     """
     i = 0
     # NB: if we have found a homopolymer of length `min_hp`, then we do not need
     # to examine the last `min_hp` bases since we'll never find a longer one.
     bases_len = len(bases)
-    while i < bases_len - min_hp:
+    while i < bases_len - min_length:
         base = bases[i].upper()
         j = i + 1
         while j < bases_len and bases[j] == base:
             j += 1
-        min_hp = max(min_hp, j - i)
+        min_length = max(min_length, j - i)
         # skip over all the bases in the current homopolymer
         i = j
-    return min_hp
+    return min_length
 
 
 def longest_dinucleotide_run_length(bases: str) -> int:
--- a/fgpyo/sequence.py
+++ b/fgpyo/sequence.py
@@ -76,29 +76,11 @@ def reverse_complement(bases: str) -> str:
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
 
 
-def longest_hp_length(bases: str) -> int:
-    """Calculates the length of the longest homopolymer in the input sequence."""
-    max_hp = 0
-    i = 0
-    # NB: if we have found a homopolymer of length `max_hp`, then we do not need
-    # to examine the last `max_hp` bases since we'll never find a longer one.
-    bases_len = len(bases)
-    while i < bases_len - max_hp:
-        base = bases[i]
-        j = i + 1
-        while j < bases_len and bases[j] == base:
-            j += 1
-        max_hp = max(max_hp, j - i)
-        # skip over all the bases in the current homopolymer
-        i = j
-    return max_hp
-
-
 def gc_content(bases: str) -> float:
     """Calculates the fraction of G and C bases in a sequence."""
     if len(bases) == 0:
         return 0
-    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
+    gc_count = sum(1 for base in bases if base in "CGcg")
     return gc_count / len(bases)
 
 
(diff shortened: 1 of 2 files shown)


fgpyo/sequence.py
```diff
@@ -70,9 +71,19 @@ def reverse_complement(bases: str) -> str:
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
 
 
-def longest_hp_length(bases: str) -> int:
+def gc_content(bases: str) -> float:
```
nh13, line 74: moved around the order of the methods...

msto (changes requested):
Could you also please include `Args`/`Returns`/`Raises` blocks in the docstring of each public function?

tests/fgpyo/test_sequence.py
```diff
@@ -54,3 +54,37 @@ def test_homopolymer(bases: str, expected_hp_len: int) -> None:
...
+        ("TAGTAGTAG", 3, 9),
+        ("TATAGTAGTAGTA", 3, 9),
+        ("TACGTACGTACTG", 4, 8),
+        ("TAGTAGTAGTAG", 3, 12),
+        ("TAGTAGTAGTAG", 6, 12),
+        ("TAGTAGTAGTAG", 12, 12),
+    ],
+)
+def test_longest_multinucleotide_run_length(bases: str, n: int, expected_length: int) -> None:
+    assert expected_length == 0 or expected_length % n == 0
```
msto, line 89:
```suggestion
    assert expected_length % n == 0
```

`0 mod n` is zero 🙂

tests/fgpyo/test_sequence.py
```diff
@@ -54,3 +54,37 @@ def test_homopolymer(bases: str, expected_hp_len: int) -> None:
...
+        ("A", 1, 1),
+        ("A", 2, 0),
+        ("ACGT", 2, 2),
+        ("AACCGGTT",2,  2),
+        ("TTTTCCCGGA",2,  4),
+        ("ACCGGGTTTT",2,  4),
+        ("GTGTGTAAAA", 2, 6),
+        ("GTGTGTGAAAA",2,  6),
+        ("GTGTGTACACACAC",2,  8),
+        ("GTGTGTGACACACAC",2,  8),
```
msto, line 73:
```suggestion
        ("AACCGGTT", 2, 2),
        ("TTTTCCCGGA", 2, 4),
        ("ACCGGGTTTT", 2, 4),
        ("GTGTGTAAAA", 2, 6),
        ("GTGTGTGAAAA", 2, 6),
        ("GTGTGTACACACAC", 2, 8),
        ("GTGTGTGACACACAC", 2, 8),
```

Have you run `ruff`?

tests/fgpyo/test_sequence.py
```diff
@@ -54,3 +54,37 @@ def test_homopolymer(bases: str, expected_hp_len: int) -> None:
...
+        ("TGTATATATA", 2, 8),
+        ("TGTATATATAC",2,  8),
+        ("TGATATATATC",2,  8),
+        ("TGATATATAC", 2, 6),
+        ("TAGTAGTAG", 3, 9),
+        ("TATAGTAGTAGTA", 3, 9),
+        ("TACGTACGTACTG", 4, 8),
+        ("TAGTAGTAGTAG", 3, 12),
+        ("TAGTAGTAGTAG", 6, 12),
+        ("TAGTAGTAGTAG", 12, 12),
```
msto, line 85:
```suggestion
        ("AACCGGTT", 1, 8),
        ("AACCGGTT", 3, 1),
        ("TAGTAGTAGTAG", 5, 5),
        ("TAGTAGTAGTAG", 7, 7),
```

A few more edge cases?

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
...
+    return best_length
+
+
+def longest_multinucleotide_run_length(bases: str, n: int) -> int:
+    """Number of bases in the longest multi-nucleotide run.
+
+    A multi-nucleotide run is when N nucleotides are repeated in tandem. For example,
+    TTGG (length = 4, N=2) or TAGTAGTAG (length = 9, N = 3). If there are no such runs,
+    returns 0."""
+
```
msto, line 133: Should this function raise an error when `n` is non-positive?

msto (changes requested):
Looped through the tests and a few more thoughts came to mind 🙂

tests/fgpyo/test_sequence.py
msto: Our actions (and `check.sh`, presumably) don't run `ruff` on the `tests/` directory. I'll open a quick PR to fix

msto: #163

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
...
+    if len(bases) < n:
         return 0
-    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
-    return gc_count / len(bases)
+    elif len(bases) == n:
+        return len(bases)
+    elif n == 1:
+        return longest_hp_length(bases=bases)
+    elif n == 2:
+        return longest_dinucleotide_run_length(bases=bases)
```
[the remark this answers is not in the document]
msto: ignore me, this is in the PR description 🙃

fgpyo/sequence.py
```diff
@@ -70,9 +71,19 @@ def reverse_complement(bases: str) -> str:
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
 
 
-def longest_hp_length(bases: str) -> int:
+def gc_content(bases: str) -> float:
+    """Calculates the fraction of G and C bases in a sequence."""
+    if len(bases) == 0:
+        return 0
+    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
```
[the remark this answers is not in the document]
msto: I've updated my suggestions above to use `base in "CG"` instead of `base in ["C", "G"]`

`base in "CG"` will be about twice as performant as `base in ["C", "G"]`, and roughly as performant as `base in {"C", "G"}` while being (imo) more readable

```py
In [1]: %timeit "G" in "CG"
12.5 ns ± 0.0861 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [2]: %timeit "G" in ["C", "G"]
19.8 ns ± 0.136 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit "G" in {"C", "G"}
11.2 ns ± 0.143 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```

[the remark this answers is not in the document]
tfenne: @msto You didn't include a timing for multiple equality checks ... so I took a look myself, and at least for me, multiple equality checks is 3-4 times _slower_ than using `ch in "ACGT"`.  So +1 for using `in` over multiple equality checks.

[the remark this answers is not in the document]
tfenne: I think however, that calling `.upper()` first is probably less desirable since this might get called on long sequences or called frequently, and checking `in "CGcg"` is probably faster than calling `.upper()` on all inputs?

fgpyo/sequence.py
```diff
@@ -70,9 +71,19 @@ def reverse_complement(bases: str) -> str:
...
+    return gc_count / len(bases)
+
+
+def longest_hp_length(bases: str, _max_hp: Optional[int] = None) -> int:
     """Calculates the length of the longest homopolymer in the input sequence."""
-    max_hp = 0
+    # NB: _max_hp is a private parameter used by longest_multinucleotide_run_length when we want
+    # to search for homopolymer run lengths of at least _max_hp
+    max_hp = 0 if _max_hp is None else _max_hp
```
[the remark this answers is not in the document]
tfenne: I tend to agree with Matt here (especially on naming).  If you really want this to be private, you could just define two functions?

```python 
def longest_hp_length(bases: str) -> int:
    return _longest_hp_length(bases, 1)

def _longest_hp_length(bases: str, _min_hp: int) -> int:
    ...
```

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
     return max_hp
 
 
-def gc_content(bases: str) -> float:
-    """Calculates the fraction of G and C bases in a sequence."""
-    if len(bases) == 0:
+def longest_dinucleotide_run_length(bases: str) -> int:
```
tfenne, line 102: I think this function (and longest multi-nuc run) need to be clear on how they handle case, and ideally should be case-insensitive.  E.g. if I pulled sequence from a FASTA file and it's `ACGACCATATatatatatatGC` I would want it to count all of `ATATatatatatat`.

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
...
+    A dinucleotide run is when two nucleotides are repeated in tandem. For example,
+    TTGG (length = 4) or AACCAACCAA (length = 10). If there are no such runs, returns 0."""
+
+    best_length: int = 0
+
+    start = 0  # the start index of the current dinucleotide run
+    while start < len(bases) - 1:
+        # get the dinuc bases
+        first_base = bases[start]
+        second_base = bases[start + 1]
```
tfenne, line 114: Wouldn't it be easier to slice here?  I.e. `dinuc = bases[start:start+2].upper()`, and then your comparisons below could be simplified to `bases[end:end+2].upper() == dinuc`?

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
...
 
 
-def gc_content(bases: str) -> float:
-    """Calculates the fraction of G and C bases in a sequence."""
-    if len(bases) == 0:
+def longest_dinucleotide_run_length(bases: str) -> int:
+    """Number of bases in the longest dinucleotide run in a primer.
+
+    A dinucleotide run is when two nucleotides are repeated in tandem. For example,
+    TTGG (length = 4) or AACCAACCAA (length = 10). If there are no such runs, returns 0."""
```
tfenne, line 106: Can you add a `Returns` in the docstring please and be even more explicit that it returns the _sequence_ length not the number of repeat units?  E.g.

`returns the number of bases in the longest dinuc repeat (NOT the number of repeat units).`

fgpyo/sequence.py
```diff
@@ -88,9 +99,55 @@ def longest_hp_length(bases: str) -> int:
...
+    A multi-nucleotide run is when N nucleotides are repeated in tandem. For example,
+    TTGG (length = 4, N=2) or TAGTAGTAG (length = 9, N = 3). If there are no such runs,
+    returns 0."""
+
+    if len(bases) < n:
         return 0
-    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
-    return gc_count / len(bases)
+    elif len(bases) == n:
+        return len(bases)
```
tfenne, line 137: Or just `return n`?

tfenne (changes requested):
I find the implementation of `longest_multinucleotide_run_length` hard to reason about - it took me several reads to figure out that you're generating a series of slices N times over.  And then also passing `list[str]` to `longest_hp_length` where it takes a `str`.

I think I would prefer to:

1. take the method from `longest_dinuc` and change that to use slices (which I get may be slightly slower), and then generalize that over kmer-size and call that `longest_multinucleotide_run_length`
2. Just have `longest_dinuc` call `longest_multinucleotide_run_length(s, 2)`

fgpyo/sequence.py
```diff
@@ -76,29 +76,11 @@ def reverse_complement(bases: str) -> str:
...
-        i = j
-    return max_hp
-
-
 def gc_content(bases: str) -> float:
     """Calculates the fraction of G and C bases in a sequence."""
     if len(bases) == 0:
         return 0
-    gc_count = sum(1 for base in bases if base == "C" or base == "G" or base == "c" or base == "g")
+    gc_count = sum(1 for base in bases if base in "CGcg")
```
nh13, line 83: not part of the original target of this PR, but changed due to popular demand

fgpyo/sequence.py
```diff
@@ -164,3 +146,87 @@ def levenshtein(string1: str, string2: str) -> int:
...
+    bases_len = len(bases)
+    while i < bases_len - min_hp:
+        base = bases[i].upper()
+        j = i + 1
+        while j < bases_len and bases[j] == base:
+            j += 1
+        min_hp = max(min_hp, j - i)
+        # skip over all the bases in the current homopolymer
+        i = j
+    return min_hp
```
tfenne, line 173: I find it a little weird that with this change, the following is true:

```
longest_homopolymer_length("ACGTACGTACGT", min_hp=3) == 3
```

The problem in general with having the `min_hp` parameter is that it becomes challenging to define what the behavior should be if no homopolymer is found of at least that length.  I.e. should this return `Optional[int]` now?

fgpyo/sequence.py
```diff
@@ -76,29 +76,11 @@ def reverse_complement(bases: str) -> str:
     return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
 
 
-def longest_hp_length(bases: str) -> int:
```
tfenne, line 79: Should this function remain, be deprecated, and vector to the new one?

fgpyo/sequence.py
```diff
@@ -164,3 +146,99 @@ def levenshtein(string1: str, string2: str) -> int:
...
+
+def longest_homopolymer_length(bases: str) -> int:
+    """Calculates the length of the longest homopolymer in the input sequence.
+
+    Args:
+        bases: the bases over which to compute
+
+    Return:
+        the length of the longest homopolymer
+    """
```
nh13, line 171: @tfenne the `min_hp` was removed as it's no longer needed

tfenne (approved):

PR #163, opened 2024-07-17
feat: Lint tests directory

msto:
The `tests/` directory should be included in formatting, linting, and type-checking.

### DIFF
--- a/ci/check.sh
+++ b/ci/check.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 function banner() {
     echo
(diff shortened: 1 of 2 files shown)


ci/check.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
msto, line 1: Some of us use homebrew-installed bash 🙂

ci/check.sh
```diff
@@ -37,12 +37,11 @@ function run() {
 
 parent=$(cd $(dirname $0) && pwd -P)
 
-banner "Executing in conda environment ${CONDA_DEFAULT_ENV} in directory fgpyo"
```
msto, line 40: `CONDA_DEFAULT_ENV` is no longer defined in the script

ci/check.sh
```diff
@@ -37,12 +37,11 @@ function run() {
...
-run "Style Checking" "ruff format fgpyo"
-run "Linting"        "ruff check --fix fgpyo"
-run "Type Checking"  "mypy -p fgpyo --config $parent/mypy.ini"
+run "Style Checking" "ruff format fgpyo tests"
+run "Linting"        "ruff check --fix fgpyo tests"
+run "Type Checking"  "mypy fgpyo tests --config $parent/mypy.ini"
 run "Unit Tests"     "python -m pytest -vv -r sx tests"
-run "Make docs"      "poetry run mkdocs build --strict"
+run "Make docs"      "mkdocs build --strict"
```
msto, line 44: I have aliased `poetry` to an installation in a dedicated environment, per [Poetry's official recommendation/requirement](https://python-poetry.org/docs/#installing-with-pipx), and `run` does not expand the alias. 

>Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system. It should in no case be installed in the environment of the project that is to be managed by Poetry. This ensures that Poetry’s own dependencies will not be accidentally upgraded or uninstalled. (Each of the following installation methods ensures that Poetry is installed into an isolated environment.) In addition, the isolated virtual environment in which poetry is installed should not be activated for running poetry commands.

i.e. `alias poetry=/Users/msto/.miniforge-x86/envs/poetry/bin/poetry`

The other lints are not run via poetry, so this is also more consistent with the other lints.

nh13: TIL

nh13 (approved):
Lgtm

PR #166, opened 2024-07-17
feat: move docs out of extras into their own group

### LINKED ISSUE #165: `mkdocs` dependencies should be declared as a dependency group, not extras
Per [past discussion](https://fulcrumgenomics.slack.com/archives/C04BQ9RMJ2G/p1718200843607199?thread_ts=1718200462.992179&cid=C04BQ9RMJ2G), I think it would make more sense to declare `mkdocs` and friends as a dependency group, rather than as extras. 

Extras are made available via pip's bracketed syntax (e.g. `pip install fgpyo[pysam]`) and imo should be limited to optional dependencies _for the end-user_.

Dependency groups are only available via `poetry install` (e.g. `poetry install --with docs`) and imo should be used for optional or required developer dependencies. (Required dependencies should be declared in the `dev` group, and optionals in other groups, as appropriate.)

For `mkdocs`, these dependencies could be declared in a `docs` group. However, if our `ci/check.sh` script requires `mkdocs build` to pass, I would suggest including these in the `dev` group.

nh13:
Fixes #165

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,17 +28,17 @@ attrs = ">=19.3.0"
 pysam = ">=0.22.1"
 pytest = ">=7.4.0"
 typing_extensions = { version = ">=3.7.4", python = "<3.12" }
-mkdocs-autorefs = { version = ">=0.5.0", optional = true }
-mkdocs-include-markdown-plugin = { version = ">=6.0.1", optional = true }
-mkdocs-material = { version = ">=9.2.8", optional = true }
-mkdocs-table-reader-plugin = { version = ">=2.0.1", optional = true }
-mkdocs = { version = ">=1.5.2", optional = true }
-mkdocs-gen-files = { version = ">=0.5.0", optional = true }
-mkdocs-literate-nav = { version = ">=0.6.1", optional = true }
-mkdocs-section-index = { version = ">=0.3.9", optional = true }
-mkdocstrings-python = { version = ">=1.6.2", optional = true }
-mkdocstrings = { version = ">=0.23.0", optional = true }
-black = {version = ">=24.4.2", optional = true}
+mkdocs-autorefs = { version = ">=0.5.0" }
+mkdocs-include-markdown-plugin = { version = ">=6.0.1" }
+mkdocs-material = { version = ">=9.2.8" }
+mkdocs-table-reader-plugin = { version = ">=2.0.1" }
+mkdocs = { version = ">=1.5.2" }
+mkdocs-gen-files = { version = ">=0.5.0" }
+mkdocs-literate-nav = { version = ">=0.6.1" }
+mkdocs-section-index = { version = ">=0.3.9" }
+mkdocstrings-python = { version = ">=1.6.2" }
+mkdocstrings = { version = ">=0.23.0" }
+black = {version = ">=24.4.2"}
 numpy = [
   {version = "^1.26.4", python = ">=3.12"},
   {version = "^1.25.2", python = ">=3.9,<3.12"},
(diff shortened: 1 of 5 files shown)


nh13:
Using a separate group for better granularity, versus lumping it all into the dev group

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
...
+mkdocs-autorefs = { version = ">=0.5.0" }
+mkdocs-include-markdown-plugin = { version = ">=6.0.1" }
+mkdocs-material = { version = ">=9.2.8" }
+mkdocs-table-reader-plugin = { version = ">=2.0.1" }
+mkdocs = { version = ">=1.5.2" }
+mkdocs-gen-files = { version = ">=0.5.0" }
+mkdocs-literate-nav = { version = ">=0.6.1" }
+mkdocs-section-index = { version = ">=0.3.9" }
+mkdocstrings-python = { version = ">=1.6.2" }
+mkdocstrings = { version = ">=0.23.0" }
```
msto, line 40: I think these can be removed from the package dependencies now, if they're provided via the `docs` group

```suggestion
```

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
...
+black = {version = ">=24.4.2"}
```
msto, line 41:
```suggestion
```

Similarly I think `black` can be removed as a package dependency and made a `dev` dependency.

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
...
 
 [tool.poetry.dev-dependencies]
```
msto, line 65:
```suggestion
[tool.poetry.group.dev.dependencies]
```

`dev-dependencies` is [deprecated](https://python-poetry.org/docs/managing-dependencies/)

> Poetry will slowly transition away from the dev-dependencies notation which will soon be deprecated, so it’s advised to migrate your existing development dependencies to the new group notation.

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
 pysam = ">=0.22.1"
 pytest = ">=7.4.0"
```
msto, line 29:
```suggestion
```

I think `pytest` should also be a dev dependency, not a package dependency

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
...
+mkdocs-include-markdown-plugin = { version = ">=6.0.1" }
+mkdocs-material = { version = ">=9.2.8" }
+mkdocs-table-reader-plugin = { version = ">=2.0.1" }
+mkdocs = { version = ">=1.5.2" }
+mkdocs-gen-files = { version = ">=0.5.0" }
+mkdocs-literate-nav = { version = ">=0.6.1" }
+mkdocs-section-index = { version = ">=0.3.9" }
+mkdocstrings-python = { version = ">=1.6.2" }
+mkdocstrings = { version = ">=0.23.0" }
+black = {version = ">=24.4.2"}
```
msto, line 63:
```suggestion
```

I think this should be included in `dev` instead of `docs`

pyproject.toml
```diff
@@ -28,38 +28,39 @@ attrs = ">=19.3.0"
...
 
 [tool.poetry.dev-dependencies]
 ruff = "0.4.8"
```
msto, line 66: Could you please add `black` to the dev dependencies?

msto (changes requested):
@nh13 If the `docs` dependencies are not required dev dependencies, would you be open to removing `mkdocs build` from the `ci/check.sh` script? I think that script should be runnable with only dev dependencies.

pyproject.toml
nh13: `mkdocstrings` and friends need `black` for some reason!

msto (approved):
lgtm!

tfenne (approved):
LGTM

PR #167, opened 2024-07-18
feat: add hamming and levenshtein distance functions

### LINKED ISSUE #93: Feature request: support sequence distance functions
Could be helpful to have Hamming and Levenshtein distance available

znorgaard:
Closes #93.

This PR adds functions for calculating Hamming and Levenshtein distances between two strings.

These are included for convenience and are not the optimal solution for performing many distance calculations. In those situations you should use a C-based implementation (ex. https://pypi.org/project/Distance/).


fgpyo/util/distance.py
```diff
@@ -0,0 +1,70 @@
+"""
```
nh13, line 1: I would put these methods in `sequence.py`, as I think we'll want to slowly move away from having a `util` sub-module where things just get shoved.

fgpyo/util/distance.py
```diff
@@ -0,0 +1,70 @@
...
+If you are performing many distance calculations, using a C based method is preferable.
+ex. https://pypi.org/project/Distance/
+"""
+
+from typing import List
+
+
+def hamming(string1: str, string2: str) -> int:
+    """
+    Calculates hamming distance between two strings. Strings must be of equal lengths.
```
nh13, line 12: Mention case sensitivity

fgpyo/util/distance.py
```diff
@@ -0,0 +1,70 @@
...
+        raise ValueError(
+            "Hamming distance requires two strings of equal lengths."
+            f"Received {string1} and {string2}."
+        )
+    return sum([string1[i] != string2[i] for i in range(len(string1))])
+
+
+def levenshtein(string1: str, string2: str) -> int:
+    """
+    Calculates levenshtein distance between two strings.
```
nh13, line 31: ditto, case sensitivity

fgpyo/util/distance.py
```diff
@@ -0,0 +1,70 @@
+"""
+These functions are included for convenience.
+If you are performing many distance calculations, using a C based method is preferable.
+ex. https://pypi.org/project/Distance/
```
nh13, line 4: Add to docs in sequence.py and name specific methods (hamming and levenshtein)

nh13 (changes requested):
One request to move the methods to sequence.py and a few other things, but other than that, looks good.

nh13 (commented):
Any reason not to support swaps, and sub/ins/del/swap penalities?  If not, perhaps we add an issue?  

See: https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala#L57

znorgaard:
> Any reason not to support swaps, and sub/ins/del/swap penalities? If not, perhaps we add an issue?
> 
> See: https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala#L57

Swaps aren't part of the classic Levenshtein distance. I'd argue for an additional `damerau_levenshtein` function or a boolean `demareu` option to support swapping.

No real reason for not including penalties. that would be an easy addition in this PR if it's desirable.

tfenne:
> Any reason not to support swaps, and sub/ins/del/swap penalities? If not, perhaps we add an issue?

Let me turn that around and ask .... who has an immediate use for that?  I'd prefer not to overcomplicate things.  As @znorgaard said in the PR - these are really here for "I need a quick and simple distance function" and one should use optimized versions for other work, so I think it's probably not worth investing effort making these more feature rich.

nh13:
Just thinking about what we have in commons:
https://github.com/fulcrumgenomics/commons/blob/9372c5585084a1d7486c1b6036d0e06824b2c8d4/src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala#L57

As suggested, we can make a PR later.

znorgaard:
Created an issue to track additional functionality requests, #170.

nh13 (approved):
Thank-you, LGTM

PR #168, opened 2024-07-19
feat: support PEP604 syntax `str | None`

### LINKED ISSUE #89: Metric.read should support PEP 604 syntax for `Optional` types
Currently, annotating an attribute as `Optional[str]` works, but annotating the same attribute as `str | None` results in the following error when trying to read an empty field

```py
AssertionError: Do not know how to convert string to str | None for value:
```

nh13:
Fixes #89

### DIFF
--- a/fgpyo/util/types.py
+++ b/fgpyo/util/types.py
@@ -166,7 +166,7 @@ def _make_union_parser_worker(
     raise ValueError(f"{value} could not be parsed as any of {union}")
 
 
-def make_union_parser(union: Type[UnionType], parsers: Iterable[Callable[[str], type]]) -> partial:
+def make_union_parser(union: Type[UnionType], parsers: Iterable[Callable[[str], UnionType]]) -> partial:
     """Generates a parser function for a union type object and set of parsers for the possible
     parsers to that union type object
     """
--- a/fgpyo/util/types.py
+++ b/fgpyo/util/types.py
@@ -112,6 +112,9 @@ def _is_optional(dtype: TypeAnnotation) -> bool:
     An optional type may be declared using three syntaxes: `Optional[T]`, `Union[T, None]`, or `T |
     None`. All of these syntaxes are supported by this function.
 
+    Type annotations may be any of `type`, `UnionType`, `types.GenericAlias`,
+    or `typing._GenericAlias`.
+
     Args:
         dtype: A type.
 
@@ -121,8 +124,6 @@ def _is_optional(dtype: TypeAnnotation) -> bool:
 
     Raises:
         TypeError: If the input is not a valid `TypeAnnotation` type.
-        Type annotations may be any of `type`, `types.UnionType`, `types.GenericAlias`,
-        or `typing._GenericAlias`.
     """
     if not isinstance(dtype, TYPE_ANNOTATION_TYPES):
         raise TypeError(f"Expected type annotation, got {type(dtype)}: {dtype}")
--- a/fgpyo/util/types.py
+++ b/fgpyo/util/types.py
@@ -1,5 +1,6 @@
 import collections
 import inspect
+import sys
 import typing
 from enum import Enum
 from functools import partial
@@ -11,7 +12,18 @@ from typing import TypeVar
 from typing import Union
 from typing import cast
 
-UnionType = TypeVar("UnionType", bound="Union")
+from typing_extensions import TypeAlias
+
+if sys.version_info >= (3, 10):
+    from types import UnionType
+else:
+    # NB: `types.UnionType`, available since Python 3.10, is **not** a `type`, but is a class.
+    # We declare an empty class here to use in the instance checks below.
+    class UnionType:
+        pass
+
+
+# UnionType = TypeVar("UnionType", bound="Union")
 EnumType = TypeVar("EnumType", bound="Enum")
 # conceptually bound to "Literal" but that's not valid in the spec
 # see: https://peps.python.org/pep-0586/#illegal-parameters-for-literal-at-type-check-time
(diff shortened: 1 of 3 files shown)


fgpyo/util/types.py
```diff
@@ -11,7 +12,18 @@
...
+if sys.version_info >= (3, 10):
+    from types import UnionType
+else:
+    # NB: `types.UnionType`, available since Python 3.10, is **not** a `type`, but is a class.
+    # We declare an empty class here to use in the instance checks below.
+    class UnionType:
+        pass
+
+
+# UnionType = TypeVar("UnionType", bound="Union")
```
nh13, line 26:
```suggestion
```

tests/fgpyo/util/test_types.py
```diff
@@ -1,11 +1,29 @@
...
 
 def test_is_listlike() -> None:
     assert types.is_list_like(List[str])
     assert types.is_list_like(Iterable[str])
     assert types.is_list_like(Sequence[str])
+    assert not types.is_list_like(str)
+
+
+def test_is_optional() -> None:
+    print(type((Union[str, NoneType])))
```
nh13, line 20:
```suggestion
```

tfenne:
I don't think I know enough to review this well, so will rely on @msto.

nh13:
@msto bump

fgpyo/util/types.py
```diff
@@ -73,9 +84,52 @@ def is_constructible_from_str(type_: type) -> bool:
     return False
 
 
-def _is_optional(type_: type) -> bool:
-    """Returns true if type_ is optional"""
-    return typing.get_origin(type_) is Union and type(None) in typing.get_args(type_)
+# NB: since `_GenericAlias` is a private attribute of the `typing` module, mypy doesn't find it
+TypeAnnotation: TypeAlias = Union[type, typing._GenericAlias, UnionType]  # type: ignore[name-defined]
```
msto, line 88: **issue** `GenericAlias` should be included as a valid type annotation type. Also add `from types import GenericAlias` above

```suggestion
TypeAnnotation: TypeAlias = Union[type, typing._GenericAlias, UnionType, 
types.GenericAlias]  # type: ignore[name-defined]
```

fgpyo/util/types.py
```diff
@@ -73,9 +84,52 @@ def is_constructible_from_str(type_: type) -> bool:
...
+    An optional type may be declared using three syntaxes: `Optional[T]`, `Union[T, None]`, or `T |
+    None`. All of these syntaxes is supported by this function.
+    Args:
+        dtype: A type.
+    Returns:
+        True if the type is a union type with exactly two elements, one of which is `None`.
+        False otherwise.
+    Raises:
+        TypeError: If the input is not a valid `TypeAnnotation` type (see above).
+    """
```
msto, line 118:
```suggestion
    """
    Check if a type is `Optional`.
    
    An optional type may be declared using three syntaxes: `Optional[T]`, `Union[T, None]`, or `T |
    None`. All of these syntaxes are supported by this function.
    
    Args:
        dtype: A type.
        
    Returns:
        True if the type is a union type with exactly two elements, one of which is `None`.
        False otherwise.
        
    Raises:
        TypeError: If the input is not a valid `TypeAnnotation` type. 
        Type annotations may be any of `type`, `types.UnionType`, `types.GenericAlias`, or `typing._GenericAlias`.
    """
```

fgpyo/util/types.py
```diff
@@ -73,9 +84,52 @@ def is_constructible_from_str(type_: type) -> bool:
...
+    4) `types.GenericAlias`, when declaring generic collection types using PEP 585 syntax (e.g.
+       `list[int]`)
+`types.GenericAlias` is a subclass of `type`, but `typing._GenericAlias` and `types.UnionType` are
+not and must be considered explicitly.
+"""
+
+# TODO When dropping support for Python 3.9, deprecate this in favor of performing instance checks
+# directly on the `TypeAnnotation` union type.
+# NB: since `_GenericAlias` is a private attribute of the `typing` module, mypy doesn't find it
+TYPE_ANNOTATION_TYPES = (type, typing._GenericAlias, UnionType)  # type: ignore[attr-defined]
```
msto, line 104: **issue** as above, we should include `types.GenericAlias`

```suggestion
TYPE_ANNOTATION_TYPES = (type, typing._GenericAlias, UnionType, GenericAlias)  # type: ignore[attr-defined]
```

fgpyo/util/types.py
```diff
@@ -73,9 +84,52 @@ def is_constructible_from_str(type_: type) -> bool:
...
+        True if the type is a union type with exactly two elements, one of which is `None`.
+        False otherwise.
+    Raises:
+        TypeError: If the input is not a valid `TypeAnnotation` type (see above).
+    """
+    if not isinstance(dtype, TYPE_ANNOTATION_TYPES):
+        raise TypeError(f"Expected type annotation, got {type(dtype)}: {dtype}")
+
+    origin = typing.get_origin(dtype)
+    args = typing.get_args(dtype)
```
msto, line 123: **nit** I think it's more in line with our convention to import these functions instead of referencing them as module attributes

```suggestion
    origin = get_origin(dtype)
    args = get_args(dtype)
```

tests/fgpyo/util/test_types.py
```diff
@@ -1,11 +1,28 @@
...
+def test_is_optional() -> None:
+    assert types._is_optional(Union[str, NoneType])
+    assert types._is_optional(Optional[str])
+    assert not types._is_optional(str)
+
+
+if sys.version_info >= (3, 10):
+
+    def test_is_optional_python_310() -> None:
+        assert types._is_optional(str | None)
```
msto, line 28: **suggestion**  Save a couple lines?

```suggestion
    assert not types._is_optional(str)

    if sys.version_info >= (3, 10):
        assert types._is_optional(str | None)
```

tests/fgpyo/util/test_types.py
```diff
@@ -1,11 +1,28 @@
...
     assert types.is_list_like(List[str])
     assert types.is_list_like(Iterable[str])
     assert types.is_list_like(Sequence[str])
+    assert not types.is_list_like(str)
+
+
+def test_is_optional() -> None:
+    assert types._is_optional(Union[str, NoneType])
+    assert types._is_optional(Optional[str])
+    assert not types._is_optional(str)
```
msto, line 22: **suggestion** Test a couple union types that don't meet our criteria

```suggestion
    assert types._is_optional(Optional[str])
    
    assert not types._is_optional(Union[str, int])
    assert not types._is_optional(Union[str, int, None])
    assert not types._is_optional(str)
```

msto (changes requested):

tests/fgpyo/util/test_types.py
msto: NB: for brevity (and to catch multiple failures in a single `pytest` invocation), you could use `pytest.mark.parameterize` over the tested types

PR #175, opened 2024-08-07
fix: fgpyo.fasta.sequence_dictionary incorrect StrEnum import

nh13:
Was getting:
    ModuleNotFoundError: No module named 'strenum'


tfenne (approved):
Maybe we should have an fgpyo function like:

```
is_py_version_at_least("3.11")
```

that can have tests, and then use it here?

PR #178, opened 2024-08-14
feat: support metric custom column ordering and sub-setting

### LINKED ISSUE #177: Metric.write should support alternate ordering of fields
Ordering is determined by the order of attributes in the class definition, but in cases where a `Metric`-derived class has been sub-classed, sometimes that ordering could be improved by interleaving some of the sub-class fields with the parent class fields, or having sub-class fields appear prior to the parent class fields.

nh13:
Fixes #177

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -111,6 +111,32 @@ Formatting and parsing the values for custom types is supported by overriding th
    >>> Person(name=Name(first='john', last='doe'), age=42, address=None).formatted_values()
    ["first last", "42"]
 ```
+
+Re-ordering and sub-setting the columns when writing is supported by overriding the `_header()`
+method.  In the example below, the `weight` field is not written, but is optional to support reading
+the metric back in.  Also, the `name` and `age` columns are written in reverse order.
+
+```python
+   >>> from fgpyo.util.metric import Metric
+   >>> import attr
+   >>> @attr.s(auto_attribs=True, frozen=True)
+   ... class Person(Metric["Person"]):
+   ...     name: str
+   ...     age: int
+   ...     weight: Optional[int]
+   ...     @classmethod
+   ...     def _header(cls, field_types: Optional[List[FieldType]] = None) -> List[str]:
+   ...         return ["age", "name"]
+   >>> person = Person(name="john", age=42, weight=180)
+   >>> person.header()
+   ["age", "name"]
+   >>> list(person.values())
+   [42, "john"]
+   >>> Person.write(Path("/path/to/metrics.txt"), person)
+   >>> list(Person.read(Path("/path/to/metrics.txt")))
+   [Person(name="john", age=42, weight=None)]
+```
+
 """
 
 from abc import ABC
@@ -122,10 +148,13 @@ from typing import Dict
 from typing import Generic
 from typing import Iterator
 from typing import List
+from typing import Optional
 from typing import TypeVar
+from typing import final
 
 from fgpyo import io
 from fgpyo.util import inspect
+from fgpyo.util.inspect import FieldType
 
 MetricType = TypeVar("MetricType", bound="Metric")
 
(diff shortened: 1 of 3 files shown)


ameynert (changes requested):
Additional test requested

fgpyo/util/metric.py
```diff
@@ -245,21 +276,41 @@ def write(cls, path: Path, *values: MetricType) -> None:
...
+        field_types = list(inspect.get_fields(cls))  # type: ignore[arg-type]
+        field_names = {field.name for field in field_types}
+        header = cls._header(field_types=field_types)
+        extra_fields = [h for h in header if h not in field_names]
+        if len(extra_fields) > 0:
+            raise ValueError("header() returned extra fields: " + ", ".join(extra_fields))
+        return header
+
+    @classmethod
+    def _header(cls, field_types: Optional[List[FieldType]] = None) -> List[str]:
```
tfenne, line 305: Could I suggest calling this something else - e.g. `_fields_to_write()`?  My reasoning:

1. It's a little confusing to have `header()` and `_header()`
2. More importantly, I _might_ think this would allow me to provide a custom header that doesn't match field names (e.g. `[f.replace("_", " ") for f in in self.header()]`

tfenne: Why is `field_types` optional?  Given this is a function called by the `Metric` class itself, it feels like it should either always be provided, or be removed.  How does the user who overrides this know whether or not they'll get a value?

tfenne (changes requested):

nh13 (commented):
@tfenne:
1. I've changed `_header` to `_fields_to_write`
2. This PR _does not_ allow a customer header that doesn't match the field names, as we use the names returned from the `header()` method to look up attributes on the class (see the `yield getattr(self, name)` in the `values()` method)
3.  I've made `field_types` not optional.  Good catch, as this was a left-over from a previous attempt.

ameynert (commented):
Looks good!

ameynert (approved):
tick!

fgpyo/util/metric.py
```diff
@@ -245,21 +277,39 @@ def write(cls, path: Path, *values: MetricType) -> None:
...
+        field_types = list(inspect.get_fields(cls))  # type: ignore[arg-type]
+        field_names = {field.name for field in field_types}
+        header = cls._fields_to_write(field_types=field_types)
+        extra_fields = [h for h in header if h not in field_names]
+        if len(extra_fields) > 0:
+            raise ValueError("header() returned extra fields: " + ", ".join(extra_fields))
+        return header
+
+    @classmethod
+    def _fields_to_write(cls, field_types: List[FieldType]) -> List[str]:
```
nh13, line 306: todo: move this to a parameter in the `read` method

clintval (changes requested):
Last time we talked, we discussed a few different design ideas that might be preferable.

The key issue was whether to attach this functionality to the `Metric` class itself at class definition time, or whether it is more appropriate to define these behaviors at serialization time.

I think we agreed that serialization time (`MetricWriter` or `Metric.write()`) was a better approach because if you make this a part of the class definition, then you must control or change the behavior with more class inheritance. More simply put, this is a serialization issue, not a class definition/instantiation issue. 

Handling this at serialization time is similar to how the standard library works ([see `fieldnames` in `csv.DictWriter()`](https://docs.python.org/3/library/csv.html)). Upon thinking about this a bit more, I realize the only downside is that validation errors will occur at serialization time and not class instantiation time such as with a `__post_init__()` check. If we are OK with that, then I think refactoring this solution to work with `Metric.write()` and `MetricWriter()` is preferred. A simple `fieldnames` argument could be optionally provided to allow subsetting and re-ordering of fields. A `KeyError` exception could be raised if a key is provided that doesn't exist on the dataclass (or any of it's parents).

nh13:
Superseded by #269 which adds both class-level (`_fields_to_write()`) and call-level (`include_fields`/`exclude_fields`) approaches per @clintval's feedback.

Issue #83, opened 2023-12-22
Add a method on Template to fixmate information 


nh13:
Add a method `fixmate` on `Template` that will set mate flags and appropriate SAM tags.  This is useful when we modify the records (e.g. clipping) in a template and want to update the tags.

For a code example, see [fgbio's `fixMateInfo`](https://github.com/fulcrumgenomics/fgbio/blob/9c475fdb06e4f05a2a5eabe9006e0facb31726b5/src/main/scala/com/fulcrumgenomics/bam/Bams.scala#L109C7-L109C18) and how it uses [htsjdk's `setMateInformationOnSupplementalAlignment`](https://github.com/samtools/htsjdk/blob/b16416c078e2df3e3a21486207d0c0c31b293cb9/src/main/java/htsjdk/samtools/SamPairUtil.java#L308C32-L308C32).  Note: we already have a method [`set_pair_info` in fgpyo](https://github.com/fulcrumgenomics/fgpyo/blob/0c5d6ffbbd3a539567dd385c32c245dbef368b03/fgpyo/sam/__init__.py#L589) that will reset mate info for _primary_ records in a pair where both records are mapped.

One can also validate this by running both/either `fgbio SetMateInformation` or `samtools fixmate`.  Probably a good idea to do both.

Tests should include templates with:
- only primary records
- one or both ends unmapped
- secondary records (are not updated)
- supplementary records (are updated)

Ideally the `fixmate` method on `Template` routes to `set_pair_info` and creates a new method `set_pair_info_on_supplementary`, as these methods are useful in their own right.

PR #182, opened 2024-09-03
feat: added template.fixmate() (#83)

TimD1:
- added fixmate() method
- added set_mate_info_on_supplementary() helper function
- moved builder._set_mate_info() to helper function set_mate_info()
- moved test_template_iterator.py to test_template.py (contained more than just iterator)
- added template_test_fixmate() tests

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -4,7 +4,7 @@
 This module contains utility classes for working with SAM/BAM files and the data contained
 within them.  This includes i) utilities for opening SAM/BAM files for reading and writing,
 ii) functions for manipulating supplementary alignments, iii) classes and functions for
-maniuplating CIGAR strings, and iv) a class for building sam records and files for testing.
+manipulating CIGAR strings, and iv) a class for building sam records and files for testing.
 
 ## Motivation for Reader and Writer methods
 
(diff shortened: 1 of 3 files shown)


fgpyo/sam/__init__.py
```diff
@@ -691,6 +691,100 @@ def set_pair_info(r1: AlignedSegment, r2: AlignedSegment, proper_pair: bool = Tr
...
+            rec.set_tag("MC", other.cigarstring)
+
+        if r1.reference_id == r2.reference_id:
+            r1p = r1.reference_end if r1.is_reverse else r1.reference_start
+            r2p = r2.reference_end if r2.is_reverse else r2.reference_start
+            r1.template_length = r2p - r1p
+            r2.template_length = r1p - r2p
+
+            # Arbitrarily set proper pair if the we have an FR pair with isize <= 1000
+            if r1.is_reverse != r2.is_reverse and abs(r1.template_length) <= 1000:
```
msto, line 758: **issue** This will break many structural variant callers. 

Is there a reason to overwrite the proper pair status reported by the aligner?

msto (changes requested):

fgpyo/sam/__init__.py
TimD1: I'm not sure what initially motivated this; I transferred this `_set_mate_info()` helper function out of `fgpyo/sam/builder.py`. Are you aware of a reason for it to be present there?

tfenne: @msto what's the alternative?  Invoking this function is tantamount to saying "I've changed some of the values on one or both of these reads, please make sure everything is in sync".  I understand the concern around `proper_pair` - and it's true that this implementation made sense when it was in SamBuilder, and less so here...

But if someone has changed alignment values .. the value from the aligner is also questionable here.

clintval: @coderabbitai what do you think?

tfenne: @TimD1 - @msto and I chatted and we think the right solution here is a little complicated.  We suggest doing something like:

```python
def set_mate_info(
  r1: AlignedSegment,
  r2: AlignedSegment,
  is_proper_pair: Callable[[AlignedSegment, AlignedSegment], bool] = lambda a, b: a.is_proper_pair and b.is_proper_pair
) -> Unit
    ...
```

I.e. the default is to leave it `True` if it's already set on R1 and R1, but set it to false if already false, or there is disagreement.  Then a user (e.g. SamBuilder) can provide it's own lambda.

PR #186, opened 2024-09-12
feat: Add `assert_fasta_indexed`

msto:
I think it would be helpful to have a utility function to verify the existence and readability of a FASTA and its associated indexes.


fgpyo/io/__init__.py
```diff
@@ -263,3 +263,39 @@ def suppress_stderr() -> Generator[None, None, None]:
     """A context manager that redirects output of stderr to /dev/null"""
     with redirect_to_dev_null(file_num=sys.stderr.fileno()):
         yield
+
+
+def assert_fasta_is_indexed(
+    fasta: Path,
+    faidx: bool = True,
+    dictionary: bool = True,
+    bwa: bool = True,
```
clintval, line 272: Maybe we need a poll, I'd assume "fasta_is_indexed" to default to checking if it has a FASTA index.

The dictionary and bwa index would be extras if someone wanted to check for those too.

```suggestion
    dictionary: bool = False,
    bwa: bool = False,
```

fgpyo/io/__init__.py
```diff
@@ -263,3 +263,39 @@ def suppress_stderr() -> Generator[None, None, None]:
...
+        assert_path_is_readable(fai_index)
+
+    if dictionary:
+        dict_index = Path(f"{fasta}.dict")
+        assert_path_is_readable(dict_index)
+
+    if bwa:
+        suffixes = [".amb", ".ann", ".bwt", ".pac", ".sa"]
+        for suffix in suffixes:
+            bwa_index = Path(f"{fasta}.{suffix}")
```
clintval, line 300: Too many `.`s.

```suggestion
        suffixes = ["amb", "ann", "bwt", "pac", "sa"]
        for suffix in suffixes:
            bwa_index = Path(f"{fasta}.{suffix}")
```

fgpyo/io/__init__.py
tfenne: I tend to agree ... given the name I would assume a) the fasta is known to exist, and b) we're checking that the FAI file exists.

I would agree with Clint's suggestion to change some defaults to `False`, and might also go so far as to change the function name to `assert_fasta_and_indices_present()` or something along those lines to make it clear it's checking the FASTA and multiple indices.

tfenne: I'm also torn between the utility of including BWA indices (we use that aligner so much more than any other), and the slight strangeness of singling out one aligner here.

tfenne (approved):

fgpyo/io/__init__.py
clintval: I like this and I'm OK with `assert_fasta_is_indexed()`.

Could you make all the booleans keyword only for safety sake?

```python
def assert_fasta_is_indexed(
    fasta: Path,
    /,
    dictionary: bool = False,
    bwa: bool = False,
) -> None:
```

msto: Done and done

PR #188, opened 2024-09-19
feat: add cigar utility method get_alignment_offsets 

yfarjoun:
- add cigar utility method get_alignment_offsets to find the offsets of the aligned segment in the read
- tests included

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -548,10 +548,9 @@ class Cigar:
         return sum([elem.length_on_target for elem in self.elements])
 
     def get_query_alignment_offsets(self, reverse: bool = False) -> Optional[range]:
-        """
-        Get the 0-based, end-exclusive positions of the first and last aligned base in the query.
-        The resulting range will contain the range of positions in the SEQ string for the bases t
-        hat are aligned. If no bases are aligned, the return value will be None.
+        """Gets the 0-based, end-exclusive positions of the first and last aligned base in the query.
+        The resulting range will contain the range of positions in the SEQ string for the bases
+        that are aligned. If no bases are aligned, the return value will be None.
 
         Args:
             reverse: If True, count from the end of the query. i.e. find the offsets
@@ -560,12 +559,11 @@ class Cigar:
 
         Returns:
             A range, defining the start and stop offsets of the aligned part
-            of the query. These offsets are 0-based and open-ended, with respect to the
-            beginning of the query. (If 'reverse' is True, the offsets are with
-            respect to the reversed query.)
-            If no bases are aligned, the return value will be None.
+                of the query. These offsets are 0-based and open-ended, with respect to the
+                beginning of the query. (If 'reverse' is True, the offsets are with
+                respect to the reversed query.)
+                If no bases are aligned, the return value will be None.
         """
-        # breaking the build
         start_offset: int = 0
         end_offset: int = 0
         element: CigarElement
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -547,22 +547,22 @@ class Cigar:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
-    def query_alignment_offsets(self) -> tuple:
+    def query_alignment_offsets(self) -> Tuple[int, int]:
         """Gets the 0-based, end-exclusive positions of the first and last aligned base in the
         query.
 
         The resulting range will contain the range of positions in the SEQ string for
-        the bases that are aligned. If no bases are aligned, the return value will be None.
+        the bases that are aligned.
         If counting from the end of the query is desired, use
         `cigar.reversed().query_alignment_offsets()`
 
         Returns:
             A tuple (start, stop) containing the start and stop positions
                 of the aligned part of the query. These offsets are 0-based and open-ended, with
-                respect to the beginning of the query. If no bases are aligned.
+                respect to the beginning of the query.
 
-        Throws:
-            A ValueError exception if according to the cigar, there are no aligned query bases.
+        Raises:
+            ValueError: If according to the cigar, there are no aligned query bases.
         """
         start_offset: int = 0
         end_offset: int = 0
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -549,39 +549,43 @@ class Cigar:
 
     def get_alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
         """
-        Get the starting and ending offsets for the alignment based on the CIGAR string.
+            Get the starting and ending offsets for the alignment based on the CIGAR string.
 
-        Args:
-            reverse: If True, count from the end of the read sequence.
-                Otherwise (default behavior), count from the beginning of the read sequence.
+            Args:
+                reverse: If True, count from the end of the read sequence.
+                    Otherwise (default behavior), count from the beginning of the read sequence.
 
-        Returns: 
-            A tuple (start, end), defining the start and end offsets of the _aligned part_ of the read.
-            These offsets are 0-based and open-ended, with respect to the beginning of the read sequence.
-            (If 'reverse' is True, the offsets are with respect to the end of the read sequence.)
+            Returns:
+                A tuple (start, end), defining the start and end offsets of the aligned part
+                of the read. These offsets are 0-based and open-ended, with respect to the
+                beginning of the read sequence. (If 'reverse' is True, the offsets are with
+                respect to the end of the read sequence.)
+                If the Cigar contains no alignment operators that consume sequence bases, or
+                only clipping operators, the start and end offsets will be the same value
+                (indicating an empty region). This shared value will be the offset to the first
+                base consumed by a non-clipping operator or the length of the read sequence if
+                there is no such base.
 
-        # TODO: FIGURE OUT WHY the following three lines causes mkdocs to fail
-        # If the Cigar contains no alignment operators that consume sequence bases, or
-        # only clipping operators, the start and end offsets will be the same value (indicating
-        # an empty region).
-        """
+        #
+        """  # TODO: figure out how to remove the '#' from the documentation above without
+        # breaking the build
         start_offset: int = 0
         end_offset: int = 0
-        cig_el: CigarElement
+        element: CigarElement
         alignment_began = False
         elements = self.elements if not reverse else reversed(self.elements)
-        for cig_el in elements:
-            if cig_el.operator.is_clipping and not alignment_began:
+        for element in elements:
+            if element.operator.is_clipping and not alignment_began:
                 # We are in the clipping operators preceding the alignment
-                start_offset += cig_el.length_on_query
-                end_offset += cig_el.length_on_query
-            elif cig_el.operator.is_clipping:
-                # We have exited the alignment and are in the clipping operators after the alignment
-                break
-            else:
+                start_offset += element.length_on_query
+                end_offset += element.length_on_query
+            elif not element.operator.is_clipping:
                 # We are within the alignment
                 alignment_began = True
-                end_offset += cig_el.length_on_query
+                end_offset += element.length_on_query
+            else:
+                # We have exited the alignment and are in the clipping operators after the alignment
+                break
 
         return start_offset, end_offset
 
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -158,7 +158,6 @@ and slices):
 import enum
 import io
 import sys
-from dataclasses import dataclass
 from pathlib import Path
 from typing import IO
 from typing import Any
@@ -361,46 +360,6 @@ class _CigarOpUtil:
     }
 
 
-@dataclass(frozen=True)
-class RangeOfBases:
-    """A simple data class for holding offsets into a range of bases in a read.
-
-    Attributes:
-        start (int): The starting offset (0-based) into the range.
-        stop (int): The ending (excluded) offset into the range.
-
-    Properties:
-        slice: A slice of the range. can be used rto extract the aligned bases from a string.
-        range: The range of bases represented by this object. Can be used to obtain the indexes
-            into the aligned bases in the read.
-        __len__: The length of the range.
-        __iter__: Enables unpacking of start and stop into a tuple by return the iterator of
-            (start, stop)
-
-    """
-
-    start: int
-    stop: int
-
-    @property
-    def slice(self) -> slice:
-        """A slice of the range"""
-        return slice(self.start, self.stop)
-
-    @property
-    def range(self) -> range:
-        """The range of bases represented by this object"""
-        return range(self.start, self.stop)
-
-    def __len__(self) -> int:
-        """The length of the range"""
-        return self.stop - self.start
-
-    def __iter__(self) -> Iterable[int]:
-        """enables unpacking of start and stop into a tuple"""
-        return (self.start, self.stop).__iter__()
-
-
 @enum.unique
 class CigarOp(enum.Enum):
     """Enumeration of operators that can appear in a Cigar string.
@@ -588,28 +547,28 @@ class Cigar:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
-    def query_alignment_offsets(self, reverse: bool = False) -> Optional[RangeOfBases]:
+    def query_alignment_offsets(self) -> tuple:
         """Gets the 0-based, end-exclusive positions of the first and last aligned base in the
-        query. The resulting range will contain the range of positions in the SEQ string for
-        the bases that are aligned. If no bases are aligned, the return value will be None.
+        query.
 
-        Args:
-            reverse: If True, count from the end of the query. i.e. find the offsets
-                using the reversed elements of the cigar.
+        The resulting range will contain the range of positions in the SEQ string for
+        the bases that are aligned. If no bases are aligned, the return value will be None.
+        If counting from the end of the query is desired, use
+        `cigar.reversed().query_alignment_offsets()`
 
         Returns:
-            A RangeOfBases object containing the start and stop positions (0-based, end-exclusive)
+            A tuple (start, stop) containing the start and stop positions
                 of the aligned part of the query. These offsets are 0-based and open-ended, with
-                respect to the beginning of the query. (If 'reverse' is True, the offsets are with
-                respect to the reversed query.) If no bases are aligned, the return value will be
-                None.
+                respect to the beginning of the query. If no bases are aligned.
+
+        Throws:
+            An
         """
         start_offset: int = 0
         end_offset: int = 0
         element: CigarElement
         alignment_began = False
-        elements = self.elements if not reverse else reversed(self.elements)
-        for element in elements:
+        for element in self.elements:
             if element.operator.is_clipping and not alignment_began:
                 # We are in the clipping operators preceding the alignment
                 # Note: hardclips have length-on-query=0
@@ -623,9 +582,9 @@ class Cigar:
                 # We have exited the alignment and are in the clipping operators after the alignment
                 break
 
-        ret = RangeOfBases(start_offset, end_offset)
-        if not alignment_began or len(ret) == 0:
-            return None
+        ret = (start_offset, end_offset)
+        if start_offset == end_offset:
+            raise ValueError("Cigar has no aligned bases")
         return ret
 
     def __getitem__(
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -548,9 +548,9 @@ class Cigar:
         return sum([elem.length_on_target for elem in self.elements])
 
     def query_alignment_offsets(self, reverse: bool = False) -> Optional[range]:
-        """Gets the 0-based, end-exclusive positions of the first and last aligned base in the query.
-        The resulting range will contain the range of positions in the SEQ string for the bases
-        that are aligned. If no bases are aligned, the return value will be None.
+        """Gets the 0-based, end-exclusive positions of the first and last aligned base in the
+        query. The resulting range will contain the range of positions in the SEQ string for
+        the bases that are aligned. If no bases are aligned, the return value will be None.
 
         Args:
             reverse: If True, count from the end of the query. i.e. find the offsets
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -549,15 +549,16 @@ class Cigar:
 
     def get_alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
         """
-                Get the starting and ending offset for the alignment based on the CIGAR string.
+        Get the starting and ending offsets for the alignment based on the CIGAR string.
 
-                Args:
-                    reverse: Whether to count from the end of the read sequence.
-                        Default is False, which counts from the beginning of the read sequence.
+        Args:
+            reverse: If True, count from the end of the read sequence.
+                Otherwise (default behavior), count from the beginning of the read sequence.
 
-                Returns: Tuple[int, int] The start and end of the _aligned part_ of the read.
-                    0-based, open-ended offsets (to the read sequence.)
-                    If 'reverse' is True, the offsets count from the end of the read sequence.
+        Returns: 
+            A tuple (start, end), defining the start and end offsets of the _aligned part_ of the read.
+            These offsets are 0-based and open-ended, with respect to the beginning of the read sequence.
+            (If 'reverse' is True, the offsets are with respect to the end of the read sequence.)
 
         # TODO: FIGURE OUT WHY the following three lines causes mkdocs to fail
         # If the Cigar contains no alignment operators that consume sequence bases, or
@@ -571,11 +572,14 @@ class Cigar:
         elements = self.elements if not reverse else reversed(self.elements)
         for cig_el in elements:
             if cig_el.operator.is_clipping and not alignment_began:
+                # We are in the clipping operators preceding the alignment
                 start_offset += cig_el.length_on_query
                 end_offset += cig_el.length_on_query
             elif cig_el.operator.is_clipping:
+                # We have exited the alignment and are in the clipping operators after the alignment
                 break
             else:
+                # We are within the alignment
                 alignment_began = True
                 end_offset += cig_el.length_on_query
 
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -368,7 +368,7 @@ class CigarOp(enum.Enum):
         code (int): The `~pysam` cigar operator code.
         character (int): The single character cigar operator.
         consumes_query (bool): True if this operator consumes query bases, False otherwise.
-        consumes_reference (bool): True if this operator consumes reference bases, False otherwise.
+        consumes_target (bool): True if this operator consumes target bases, False otherwise.
     """
 
     M = (0, "M", True, True)  #: Match or Mismatch the reference
@@ -547,27 +547,24 @@ class Cigar:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
-    def get_alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
+    def get_query_alignment_offsets(self, reverse: bool = False) -> Optional[range]:
         """
-            Get the starting and ending offsets for the alignment based on the CIGAR string.
+        Get the 0-based, end-exclusive positions of the first and last aligned base in the query.
+        The resulting range will contain the range of positions in the SEQ string for the bases t
+        hat are aligned. If no bases are aligned, the return value will be None.
 
-            Args:
-                reverse: If True, count from the end of the read sequence.
-                    Otherwise (default behavior), count from the beginning of the read sequence.
+        Args:
+            reverse: If True, count from the end of the query. i.e. find the offsets
+                using the reversed elements of the cigar.
 
-            Returns:
-                A tuple (start, end), defining the start and end offsets of the aligned part
-                of the read. These offsets are 0-based and open-ended, with respect to the
-                beginning of the read sequence. (If 'reverse' is True, the offsets are with
-                respect to the end of the read sequence.)
-                If the Cigar contains no alignment operators that consume sequence bases, or
-                only clipping operators, the start and end offsets will be the same value
-                (indicating an empty region). This shared value will be the offset to the first
-                base consumed by a non-clipping operator or the length of the read sequence if
-                there is no such base.
 
-        #
-        """  # TODO: figure out how to remove the '#' from the documentation above without
+        Returns:
+            A range, defining the start and stop offsets of the aligned part
+            of the query. These offsets are 0-based and open-ended, with respect to the
+            beginning of the query. (If 'reverse' is True, the offsets are with
+            respect to the reversed query.)
+            If no bases are aligned, the return value will be None.
+        """
         # breaking the build
         start_offset: int = 0
         end_offset: int = 0
@@ -577,6 +574,7 @@ class Cigar:
         for element in elements:
             if element.operator.is_clipping and not alignment_began:
                 # We are in the clipping operators preceding the alignment
+                # Note: hardclips have length-on-query=0
                 start_offset += element.length_on_query
                 end_offset += element.length_on_query
             elif not element.operator.is_clipping:
@@ -587,7 +585,10 @@ class Cigar:
                 # We have exited the alignment and are in the clipping operators after the alignment
                 break
 
-        return start_offset, end_offset
+        ret = range(start_offset, end_offset)
+        if not alignment_began or len(ret) == 0:
+            return None
+        return ret
 
     def __getitem__(
         self, index: Union[int, slice]
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -547,7 +547,7 @@ class Cigar:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
-    def get_query_alignment_offsets(self, reverse: bool = False) -> Optional[range]:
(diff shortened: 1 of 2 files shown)


TimD1 (approved):
Looks good to me, once the ruff/mypy CI stuff passes

yfarjoun:
not sure what's going on the mkdocs. I had to comment out some lines to get it to compile.... if someone knows how to fix this please let me know...

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,40 @@ def length_on_target(self) -> int:
...
+        """
+                Get the starting and ending offset for the alignment based on the CIGAR string.
+
+                Args:
+                    reverse: Whether to count from the end of the read sequence.
+                        Default is False, which counts from the beginning of the read sequence.
+
+                Returns: Tuple[int, int] The start and end of the _aligned part_ of the read.
+                    0-based, open-ended offsets (to the read sequence.)
+                    If 'reverse' is True, the offsets count from the end of the read sequence.
```
msto, line 560:
```suggestion
        Get the starting and ending offsets for the alignment based on the CIGAR string.

        Args:
            reverse: If True, count from the end of the read sequence.
                Otherwise (default behavior), count from the beginning of the read sequence.

        Returns: 
            A tuple (start, end), defining the start and end offsets of the _aligned part_ of the read.
            These offsets are 0-based and open-ended, with respect to the beginning of the read sequence.
            (If 'reverse' is True, the offsets are with respect to the end of the read sequence.)
```

**nit** please adhere to google style when documenting a tuple return value

> Do not imitate older ‘NumPy style’ ([example](https://numpy.org/doc/1.24/reference/generated/numpy.linalg.qr.html)), which frequently documented a tuple return value as if it were multiple return values with individual names (never mentioning the tuple). Instead, describe such a return value as: “Returns: A tuple (mat_a, mat_b), where mat_a is …, and …”. The auxiliary names in the docstring need not necessarily correspond to any internal names used in the function body (as those are not part of the API)

https://google.github.io/styleguide/pyguide.html

**nit** when documenting a boolean parameter, I always try to be explicit about the behavior when the bool is set to True (or False), rather than a phrase beginning with "Whether", which can be ambiguous

**nit** also, please dedent so the text is inline with the triple-quote

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,40 @@ def length_on_target(self) -> int:
...
+
+        # TODO: FIGURE OUT WHY the following three lines causes mkdocs to fail
```
msto, line 562: **FIXME** please fix and remove the TODO before merging

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,40 @@ def length_on_target(self) -> int:
...
+        # If the Cigar contains no alignment operators that consume sequence bases, or
+        # only clipping operators, the start and end offsets will be the same value (indicating
+        # an empty region).
+        """
+        start_offset: int = 0
+        end_offset: int = 0
+        cig_el: CigarElement
+        alignment_began = False
+        elements = self.elements if not reverse else reversed(self.elements)
+        for cig_el in elements:
```
msto, line 572:
```suggestion
        for element in elements:
```

**nit** it's a bit more conventional to have the loop variable be the singular form of the object being iterated over

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,40 @@ def length_on_target(self) -> int:
...
+                        Default is False, which counts from the beginning of the read sequence.
+
+                Returns: Tuple[int, int] The start and end of the _aligned part_ of the read.
+                    0-based, open-ended offsets (to the read sequence.)
+                    If 'reverse' is True, the offsets count from the end of the read sequence.
+
+        # TODO: FIGURE OUT WHY the following three lines causes mkdocs to fail
+        # If the Cigar contains no alignment operators that consume sequence bases, or
+        # only clipping operators, the start and end offsets will be the same value (indicating
+        # an empty region).
```
msto, line 565: **question** What will that value be? Will it always be zero, or the length of the read sequence, or something else?

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,40 @@ def length_on_target(self) -> int:
...
+        elements = self.elements if not reverse else reversed(self.elements)
+        for cig_el in elements:
+            if cig_el.operator.is_clipping and not alignment_began:
+                start_offset += cig_el.length_on_query
+                end_offset += cig_el.length_on_query
+            elif cig_el.operator.is_clipping:
+                break
+            else:
+                alignment_began = True
+                end_offset += cig_el.length_on_query
```
msto, line 580:
```suggestion
            if cig_el.operator.is_clipping and not alignment_began:
                # We are in the clipping operators preceding the alignment
                start_offset += cig_el.length_on_query
                end_offset += cig_el.length_on_query
            elif cig_el.operator.is_clipping:
                # We have exited the alignment and are in the clipping operators after the alignment
                break
            else:
                # We are within the alignment
                alignment_began = True
                end_offset += cig_el.length_on_query
```

**suggestion** clarify what each branch of the conditional maps to?

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,51 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10M", 0, 10),
+        ("10M10I", 0, 20),
+        ("10X10I", 0, 20),
+        ("10X10D", 0, 10),
+        ("10=10D", 0, 10),
+        ("10S10M", 10, 20),
+        ("10H10M", 0, 10),
+        ("10H10S10M", 10, 20),
+        ("10H10S10M5S", 10, 20),
+        ("10H10S10M5S10H", 10, 20),
```
msto, line 48: **issue** the start and end in these three cases should be 20, 30

In each case there are 20 bases of clipping - 10 hard and 10 soft - followed by 10 bases of matched sequence.

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,51 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10H", 0, 0),
+        ("10S", 10, 10),
+        ("10S10H", 10, 10),
+        ("5H10S10H", 10, 10),
```
msto, line 52: **issue** none of these CIGAR strings have aligned sequence, so the variability in the start and end offsets is confusing to me.

(See also my question above - why are the start and end 0 in some cases, the length of the CIGAR in others, and somewhere in the middle in still others?)

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,51 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10M", 0, 10),
+        ("10M10I", 0, 20),
+        ("10X10I", 0, 20),
+        ("10X10D", 0, 10),
+        ("10=10D", 0, 10),
+        ("10S10M", 0, 10),
+        ("10H10M", 0, 10),
+        ("10H10S10M", 0, 10),
+        ("10H10S10M5S", 5, 15),
+        ("10H10S10M5S10H", 5, 15),
```
msto, line 72: **issue** the start and end here should be 15, 25

(There are 15 bases of clipping - 10 hard and 5 soft - followed by 10 bases of matched sequence)

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,51 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10H", 0, 0),
+        ("10S", 10, 10),
+        ("10S10H", 10, 10),
+        ("5H10S10H", 10, 10),
+    },
+)
+def test_get_alignments_reversed(cigar_string: str, start: int, end: int) -> None:
+    cig = Cigar.from_cigarstring(cigar_string)
+    assert Cigar.get_alignment_offsets(cig, True) == (start, end)
```
msto, line 81:
```suggestion
    assert Cigar.get_alignment_offsets(cig, reverse=True) == (start, end)
```

**nit** Nils would like me to spread the good news of keyword arguments 🙂

msto (changes requested):
This will be a good addition to the package, thanks for adding it!

I think the documentation could be a bit clearer, and it looks like there's a bug when you have adjacent hard/soft clips

tests/fgpyo/sam/test_cigar.py
yfarjoun: The bases from hard clips are not in the sequence anymore...that's what make the clip _hard_...so 10,20 is correct.

fgpyo/sam/__init__.py
yfarjoun: I put it like this because I was unable to resolve...when I uncomment mkdocs barfs....

tests/fgpyo/sam/test_cigar.py
yfarjoun: see above

fgpyo/sam/__init__.py
yfarjoun:
```
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/docstrings/google.py", line 474, in _read_returns_section
    annotation = annotation.slice.elements[index]
                 ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```

fgpyo/sam/__init__.py
yfarjoun: thanks. I expanded the (commented out for now) documentation to clarify.

tests/fgpyo/sam/test_cigar.py
yfarjoun: image

tests/fgpyo/sam/test_cigar.py
yfarjoun: explained in the updated docs

yfarjoun:
Thanks for the detailed review @msto , much appreciated!

mkdocs is giving me grief and I was unable to make it play nice.... if someone knows why it barfs when I change the indentation or remove the final "comment" line... please let me know. this is the best (passing) state I could put it into....

fgpyo/sam/__init__.py
```diff
@@ -368,7 +368,7 @@ class CigarOp(enum.Enum):
         code (int): The `~pysam` cigar operator code.
         character (int): The single character cigar operator.
         consumes_query (bool): True if this operator consumes query bases, False otherwise.
-        consumes_target (bool): True if this operator consumes target bases, False otherwise.
+        consumes_reference (bool): True if this operator consumes reference bases, False otherwise.
```
yfarjoun, line 371: This is an unrelated fix. if deemed imporatant I could extract it to a separate PR.

tfenne: @yfarjoun This uses `target` intentionally - not all alignments are between a read and a reference - what if you align two reads vs. each other, or two contigs vs. each other?

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def get_alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
+        """
+            Get the starting and ending offsets for the alignment based on the CIGAR string.
```
tfenne, line 552: The whole docstring needs to be out-dented to start at the same column as the `"""`.

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def get_alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
```
tfenne, line 550: The name is quite ambiguous given that we're dealing with an alignment between a query and a target.  How about one of the following:

`get_query_alignment_offsets`
`get_query_alignment_bounds`
`get_query_alignment_range`

Also: consider returning a `range` object?

fgpyo/sam/__init__.py
tfenne, line 552:
```suggestion
        Get the 0-based, end-exclusive positions of the first and last aligned base in the query.
```

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
...
+        """
+            Get the starting and ending offsets for the alignment based on the CIGAR string.
+
+            Args:
+                reverse: If True, count from the end of the read sequence.
+                    Otherwise (default behavior), count from the beginning of the read sequence.
```
tfenne, line 556: This is unclear to me - does it _just_ count from the end?  Or does it also swap the order of "start" vs. "end"?  I think you're saying that:

```
cigar.get_alignment_offsets(reverse=True) == cigar.reverse.get_alignment_offsets()
```
?

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
...
+
+            Returns:
+                A tuple (start, end), defining the start and end offsets of the aligned part
+                of the read. These offsets are 0-based and open-ended, with respect to the
```
tfenne, line 560:
```suggestion
                of the query. These offsets are 0-based and open-ended, with respect to the
```

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
...
+                beginning of the read sequence. (If 'reverse' is True, the offsets are with
+                respect to the end of the read sequence.)
```
tfenne, line 562:
```suggestion
                respect to the end of the query sequence.)
```

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
...
+                If the Cigar contains no alignment operators that consume sequence bases, or
+                only clipping operators, the start and end offsets will be the same value
+                (indicating an empty region). This shared value will be the offset to the first
+                base consumed by a non-clipping operator or the length of the read sequence if
+                there is no such base.
```
tfenne, line 567: I think that either a) this function should return `Optional[...]` and return None when invoked on a cigar without alignment operations, _or_ raise an error in that situation.  There is no aligned start/end if there are no alignment operations.

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,48 @@ def length_on_target(self) -> int:
...
+        elements = self.elements if not reverse else reversed(self.elements)
+        for element in elements:
+            if element.operator.is_clipping and not alignment_began:
+                # We are in the clipping operators preceding the alignment
+                start_offset += element.length_on_query
+                end_offset += element.length_on_query
+            elif not element.operator.is_clipping:
+                # We are within the alignment
+                alignment_began = True
+                end_offset += element.length_on_query
```
tfenne, line 585: So ... I think you have to consider edge-case cigars here.  What does this function report for:

- `76D`
- `76I`
- `10P76S`
- `50S1000N50S`

tests/fgpyo/sam/test_cigar.py
tfenne: This is perhaps a signal that the docstring needs to explicltly explain this - i.e. the offsets are into the read sequence _present on the record_ and as such hard-clips are largely ignored.

tests/fgpyo/sam/test_cigar.py
tfenne: I do not think any of these should return values, as noted above.

tfenne (changes requested):

fgpyo/sam/__init__.py
msto: I'd argue in favor of `Optional` vs raising an error, but would be comfortable either way

tests/fgpyo/sam/test_cigar.py
msto: 🤦 

I was thinking in the context of the conditional block above, where the offset is advanced if the operator `is_clipping`

I don't know if the docstring needs to be updated (though wouldn't hurt), but maybe add a note in the respective branch that hard-clipping has zero length on query?

fgpyo/sam/__init__.py
yfarjoun: I thought I was correcting a broken link.... if this is intentional, I'll bring it back and let's haved the discussion in a different PR (if at all)

fgpyo/sam/__init__.py
yfarjoun: I tried this and mkdocs responds: 🤮 

```
^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/children.html.jinja", line 1, in top-level template code
    {% extends "_base/children.html.jinja" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/children.html.jinja", line 73, in top-level template code
    {% include class|get_template with context %}
^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/class.html.jinja", line 1, in top-level template code
    {% extends "_base/class.html.jinja" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/class.html.jinja", line 105, in top-level template code
    {% block contents scoped %}
^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/class.html.jinja", line 179, in block 'contents'
    {% block children scoped %}
^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/class.html.jinja", line 186, in block 'children'
    {% include "children"|get_template with context %}
^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/children.html.jinja", line 1, in top-level template code
    {% extends "_base/children.html.jinja" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/children.html.jinja", line 94, in top-level template code
    {% include function|get_template with context %}
^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/function.html.jinja", line 1, in top-level template code
    {% extends "_base/function.html.jinja" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/function.html.jinja", line 105, in top-level template code
    {% block contents scoped %}
^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/function.html.jinja", line 112, in block 'contents'
    {% block docstring scoped %}
^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/mkdocstrings_handlers/python/templates/material/_base/function.html.jinja", line 112, in block 'docstring'
    {% block docstring scoped %}
^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/jinja2/environment.py", line 487, in getattr
    return getattr(obj, attribute)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/functools.py", line 993, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/dataclasses.py", line 119, in parsed
    return self.parse()
           ^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/dataclasses.py", line 136, in parse
    return parse(self, parser or self.parser, **(options or self.parser_options))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/docstrings/parsers.py", line 41, in parse
    return parsers[parser](docstring, **options)  # type: ignore[operator]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/docstrings/google.py", line 809, in parse
    section, offset = reader(docstring, offset=offset + 1, **options)  # type: ignore[operator]
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yossifarjoun/micromamba/envs/fgpyo/lib/python3.12/site-packages/griffe/docstrings/google.py", line 474, in _read_returns_section
    annotation = annotation.slice.elements[index]
                 ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
IndexError: list index out of range
```

fgpyo/sam/__init__.py
yfarjoun: you mean a python range ?

yfarjoun: neat idea!

clintval: Functions prefixed with `get_` are usually reserved for true property [getters](https://docs.python.org/3/library/functions.html#property.getter).

Can we drop the prefix too?


```suggestion
    def alignment_offsets(self, reverse: bool = False) -> Tuple[int, int]:
```

https://google.github.io/styleguide/pyguide.html#315-getters-and-setters

fgpyo/sam/__init__.py
yfarjoun: yes.

fgpyo/sam/__init__.py
yfarjoun: fixed.

fgpyo/sam/__init__.py
tfenne: 100% on different PR.  Though I'll also argue there for maintaining "target".

Little known fact is that CIGAR actually predates SAM/BAM and Google believes it was first documented as part of [exonerate](https://www.ebi.ac.uk/about/vertebrate-genomics/software/exonerate-manual) which uses "query" and "target" 😛

msto (changes requested):

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,47 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def query_alignment_offsets(self, reverse: bool = False) -> Optional[range]:
```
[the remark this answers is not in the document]
yfarjoun: (almost stop!=end)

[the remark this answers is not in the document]
msto: but you can't unpack the `range`

i.e.

```python
# this is valid
from typing import NamedTuple

class Offsets(NamedTuple):
    start: int
    end: int
    
offsets = Offsets(start=1, end=200)

start, end = offsets

# this is invalid
offsets = range(1, 200)
start, end = offsets
```

So when calling the method, you can either accept the tuple or unpack it immediately, depending on which makes more sense in your context

e.g. I imagine most use cases will look like the following
```py
start, end = cigar.alignment_offsets()
aligned_sequence = query[start:end]
```

fgpyo/sam/__init__.py
```diff
@@ -360,6 +361,46 @@ class _CigarOpUtil:
     }
 
 
+@dataclass(frozen=True)
+class RangeOfBases:
```
msto, line 365: This is clever but I still think returning a `tuple[int, int]` is simpler and easier for the user

fgpyo/sam/__init__.py
```diff
@@ -547,6 +588,46 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def query_alignment_offsets(self, reverse: bool = False) -> Optional[RangeOfBases]:
+        """Gets the 0-based, end-exclusive positions of the first and last aligned base in the
+        query. The resulting range will contain the range of positions in the SEQ string for
+        the bases that are aligned. If no bases are aligned, the return value will be None.
```
tfenne, line 594:
```suggestion
        """
        Gets the 0-based, end-exclusive positions of the first and last aligned base in the query.
        
        The resulting range will contain the range of positions in the SEQ string for
        the bases that are aligned. If no bases are aligned, the return value will be None.
```

tfenne: Suggested change is to adhere to Google style guidelines of a single summary sentence separated from the longer description.

tfenne (changes requested):
@yfarjoun: @msto and I caught up in person and would request the following changes to this PR:

1. Could you please remove the `reverse` argument, and in the docstring mention that if you have a CIGAR from a negative strand read you might want to `cigar.reversed().query_alignment_offsets()`?
2. Can we please stick to just a `Tuple[int, int]` for the return type?
3. Instead of returning `Optional` .. could it just raise an exception at the end if it was called on a pathological CIGAR for which we can't generate this?  That seems exceptional since it would be an empty cigar or a cigar that is all-clipping.

fgpyo/sam/__init__.py
yfarjoun: done.

fgpyo/sam/__init__.py
yfarjoun: done.

yfarjoun:
Thanks for the comments. done.

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,46 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def query_alignment_offsets(self) -> tuple:
```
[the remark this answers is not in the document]
yfarjoun: query here doesn't mean "get"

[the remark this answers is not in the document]
yfarjoun: when I do this I get an error:

```
fgpyo/sam/__init__.py:550: error: "tuple" is not subscriptable, use "typing.Tuple" instead  [misc]
Found 1 error in 1 file (checked 52 source files)
```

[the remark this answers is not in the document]
yfarjoun: so I switched to `Tuple[int, int]`

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,46 @@ def length_on_target(self) -> int:
...
+                alignment_began = True
+                end_offset += element.length_on_query
+            else:
+                # We have exited the alignment and are in the clipping operators after the alignment
+                break
+
+        ret = (start_offset, end_offset)
+        if start_offset == end_offset:
+            raise ValueError(f"Cigar {self.__str__()} has no aligned bases")
+        return ret
```
tfenne, line 588: Why not just?

```suggestion
        if not alignment_began:
            raise ValueError(f"Cigar {self.__str__()} has no aligned bases")
        return start_offset, end_offset
```

tfenne (approved):

fgpyo/sam/__init__.py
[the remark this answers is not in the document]
msto: That's fine. Generic type hints for the standard library collections were only added in 3.9[^1], and `fgpyo` still supports 3.8. (See #187)

The main point was to document the shape and type of the tuple's contents.

[^1]: https://peps.python.org/pep-0585/

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,68 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10=10D", (0, 10)),
+        ("10S10M", (10, 20)),
+        ("10H10M", (0, 10)),
+        ("10H10S10M", (10, 20)),
+        ("10H10S10M5S", (10, 20)),
+        ("10H10S10M5S10H", (10, 20)),
+        ("76I", (0, 76)),
+    ],
+)
+def test_query_alignment_offsets(cigar_string: str, expected_range: tuple) -> None:
```
msto, line 52: Please fix the type hints here, as well. Please try to avoid generic collection type hints without a parametrized type.

```suggestion
def test_query_alignment_offsets(cigar_string: str, expected_range: Tuple[int, int]) -> None:
```

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +31,68 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10X10D", (0, 10)),
+        ("10=10D", (0, 10)),
+        ("10S10M", (0, 10)),
+        ("10H10M", (0, 10)),
+        ("10H10S10M", (0, 10)),
+        ("10H10S10M5S", (5, 15)),
+        ("10H10S10M5S10H", (5, 15)),
+    ],
+)
+def test_query_alignment_offsets_reversed(cigar_string: str, expected_range: tuple) -> None:
```
msto, line 94: ditto 

```suggestion
def test_query_alignment_offsets_reversed(cigar_string: str, expected_range: Tuple[int, int]) -> None:
```

msto (changes requested):

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,45 @@ def length_on_target(self) -> int:
         """Returns the length of the alignment on the target sequence."""
         return sum([elem.length_on_target for elem in self.elements])
 
+    def query_alignment_offsets(self) -> Tuple[int, int]:
+        """Gets the 0-based, end-exclusive positions of the first and last aligned base in the
+        query.
```
msto, line 552: I think this will fit on one line if you don't start it on the same line as the triple quote.

You could also omit the part about the positions being zero-based exclusive, and defer that detail to the body or the `Returns` section.

```suggestion
        """
        Gets the 0-based, end-exclusive positions of the first and last aligned base in the query.
```

fgpyo/sam/__init__.py
```diff
@@ -547,6 +547,45 @@ def length_on_target(self) -> int:
...
+            elif not element.operator.is_clipping:
+                # We are within the alignment
+                alignment_began = True
+                end_offset += element.length_on_query
+            else:
+                # We have exited the alignment and are in the clipping operators after the alignment
+                break
+
+        if start_offset == end_offset:
+            raise ValueError(f"Cigar {self.__str__()} has no aligned bases")
```
msto, line 586: **nit** Avoid calling the `__str__` dunder method directly; instead call `str(x)`. 

In the context of an f-string, you don't even need the latter - the object will automatically be converted to its string representation.

```suggestion
            raise ValueError(f"Cigar {self} has no aligned bases")
```

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +33,70 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10=10D", (0, 10)),
+        ("10S10M", (10, 20)),
+        ("10H10M", (0, 10)),
+        ("10H10S10M", (10, 20)),
+        ("10H10S10M5S", (10, 20)),
+        ("10H10S10M5S10H", (10, 20)),
+        ("76I", (0, 76)),
+    ],
+)
+def test_query_alignment_offsets(cigar_string: str, expected_range: Tuple[int, int]) -> None:
```
msto, line 54:
```suggestion
def test_query_alignment_offsets(cigar_string: str, expected_range: Tuple[int, int]) -> None:
    """query_alignment_offsets() should return the expected start and stop positions."""
```

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +33,70 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10S10M", (0, 10)),
+        ("10H10M", (0, 10)),
+        ("10H10S10M", (0, 10)),
+        ("10H10S10M5S", (5, 15)),
+        ("10H10S10M5S10H", (5, 15)),
+    ],
+)
+def test_query_alignment_offsets_reversed(
+    cigar_string: str, expected_range: Tuple[int, int]
+) -> None:
```
msto, line 98:
```suggestion
) -> None:
    """query_alignment_offsets() should return the expected start and stop positions on a reversed CIGAR."""
```

tests/fgpyo/sam/test_cigar.py
```diff
@@ -31,3 +33,70 @@ def test_bad_index_raises_index_error(index: int) -> None:
...
+        ("10H"),
+        ("10S"),
+        ("10S10H"),
+        ("5H10S10H"),
+        ("76D"),
+        ("10P76S"),
+        ("50S1000N50S"),
+    ],
+)
+def test_query_alignment_offsets_failures(cigar_string: str) -> None:
```
msto, line 72:
```suggestion
def test_query_alignment_offsets_failures(cigar_string: str) -> None:
    """query_alignment_offsets() should raise a ValueError if the CIGAR has no aligned positions."""
```

msto (approved):

PR #190, opened 2024-10-17
chore: Require error codes on `type: ignore` pragmas

msto:
This PR adds the mypy config introduced into our template in https://github.com/fulcrumgenomics/python-template/pull/28.

Error codes are now required on any `type: ignore` pragmas. This PR also adds error codes to any existing `ignore`s without a code.


msto:
I think it's fine to ignore codecov on this one

tfenne (approved):

PR #194, opened 2024-11-14
feat: Add sorted Cartesian product

msto:
This PR implements a generic version of the algorithm Jason outlined to efficiently produce the sorted Cartesian product of two sorted input collections.

Original implementation: https://gist.github.com/theJasonFan/b84f8036855bf4e0eff15a1c654283c5

### DIFF
--- a/fgpyo/util/itertools.py
+++ b/fgpyo/util/itertools.py
@@ -1,15 +1,91 @@
+import heapq
 from dataclasses import dataclass
-from typing import TypeVar, Callable, Iterable, Iterator
+from dataclasses import field
+from typing import Callable
+from typing import Collection
+from typing import Generic
+from typing import Iterable
+from typing import List
+from typing import Optional
+from typing import Tuple
+from typing import TypeVar
 
-T = TypeVar('T')
+T = TypeVar("T")
+
+
+@dataclass(order=True)
+class Coord:
+    x: int
+    y: int
+
+    def incr_x(self) -> "Coord":
+        return Coord(self.x + 1, self.y)
+
+    def incr_y(self) -> "Coord":
+        return Coord(self.x, self.y + 1)
+
+
+@dataclass(order=True)
+class HeapElem(Generic[T]):
+    _priority: float
+    val: Tuple[T, T] = field(compare=False)
+    coord: Coord  # NB: sorting must also be stable / deterministic on the coordinates
+
+    @classmethod
+    def new(cls, x: T, y: T, coord: Coord, priority: Callable[[T], float]) -> "HeapElem":
+        return cls(_priority=priority(x) + priority(y), val=(x, y), coord=coord)
+
+
+class ProductSumTable(Generic[T]):
+    xs: List[T]
+    ys: List[T]
+    priority: Callable[[T], float]
+
+    def __init__(self, xs: Iterable[T], ys: Iterable[T], priority: Callable[[T], float]) -> None:
+        self.xs = list(xs)
+        self.ys = list(ys)
+        self.priority = priority
+
+    def has_right(self, elem: HeapElem) -> bool:
+        return (elem.coord.x + 1) < len(self.xs)
+
+    def has_down(self, elem: HeapElem) -> bool:
+        return (elem.coord.y + 1) < len(self.ys)
+
+    def is_last(self, elem: HeapElem) -> bool:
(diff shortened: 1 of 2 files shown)


fgpyo/util/itertools.py
```diff
@@ -0,0 +1,195 @@
...
+    curr: HeapElem[T] = table.get_elem(Coord(x=0, y=0))
+
+    # Each heap element corresponds to a single pairing in the product.
+    # `heap` contains the current heap, and `result` contains the sorted pairings generated so far.
+    heap: List[HeapElem[T]] = []
+    result: List[HeapElem[T]] = []
+
+    # If no limit is specified, generate all elements.
+    if limit is None:
+        limit = len(xs) * len(ys)
```
msto, line 139: **FIXME** I think we can relax the requirement that the inputs be `Collection`s and instead permit any `Iterable`, if we update this to avoid the length checks on the inputs. 

Maybe update the loop condition to `(limit is not None and len(result) < limit)` or similar

tests/fgpyo/util/test_itertools.py
```diff
@@ -0,0 +1,24 @@
...
+    xs: List[int] = [25, 10, 10, 1]
+    ys: List[int] = [1, 2, 10]
+
+    Pairing: TypeAlias = Tuple[int, int]
+    actual_product: List[Pairing] = list(sorted_product(xs, ys, priority=lambda x: float(x)))
+
+    # We should get the same result as using `itertools.product` and then sorting, the
+    # implementation is just more efficient (in most cases).
+    expected_product: List[Pairing] = sorted(
+        product(xs, ys), key=lambda tup: tup[0] + tup[1], reverse=True
```
msto, line 21: **FIXME** the original implementation returns the product in descending order, I'd like the default behavior to return the items in ascending order, to be consistent with the standard library's `sorted()` 

We should likely include a `reverse: bool = False` parameter on `sorted_product()` as well.

tests/fgpyo/util/test_itertools.py
msto, line 1: **FIXME** I've only written a test for the `sorted_product()` function, but we should likely include some tests over the helper classes

fgpyo/util/itertools.py
```diff
@@ -0,0 +1,195 @@
...
+
+    Args:
+        xs: A sorted list of elements.
+        ys: A sorted list of elements.
+        priority: A function that assigns a priority to each element.
+        limit: If specified, only the first `limit` elements of the product will be generated.
+
+    Yields:
+        Each tuple (x, y) from the Cartesian product of xs and ys, sorted by the sum of the priority
+        of x and y.
```
msto, line 126: **FIXME** the function currently returns a `List` of pairings, but ideally would be an `Iterator`

tests/fgpyo/util/test_itertools.py
```diff
@@ -0,0 +1,24 @@
...
+
+from typing_extensions import TypeAlias
+
+from fgpyo.util.itertools import sorted_product
+
+
+def test_sorted_product() -> None:
+    """sorted_product should return a sorted cartesian product of the input lists."""
+    xs: List[int] = [25, 10, 10, 1]
+    ys: List[int] = [1, 2, 10]
```
theJasonFan, line 13:
```suggestion
    ys: List[int] = [10, 2, 1]
```

msto: Doh

leaving myself a reminder to verify the inputs are sorted (and in the correct order)

fgpyo/util/itertools.py
```diff
@@ -0,0 +1,195 @@
...
+            x=x, y=y, coord=Coord(len(self.xs) - 1, len(self.ys) - 1), priority=self.priority
+        )
+
+
+def sorted_product(
+    xs: Collection[T],
+    ys: Collection[T],
+    priority: Callable[[T], float],
+    limit: Optional[int] = None,
+) -> List[Tuple[T, T]]:
```
msto, line 114: **FIXME** we should parametrize the algorithm used by the function, to enable the user to fall back to the naive approach (`sorted(itertools.product(xs, ys))`) for small inputs

also, someone sufficiently motivated could benchmark the two methods and identify a reasonable break even point

msto:
Closing because the anticipated use case (primer pairing) is better (or at least more easily) served by the naive approach

PR #196, opened 2024-12-03
ci: migrate ubuntu-latest to ubuntu-24.04

nh13:
Removes CI checks for Python 3.8 (see https://github.com/fulcrumgenomics/fgpyo/issues/187)

PR #197, opened 2024-12-05
feat: remove python 3.8 support

nh13:


pyproject.toml
```diff
@@ -73,7 +73,7 @@ unfixable = ["B"]
 force-single-line = true
 
 [tool.mypy]
-python_version              = 3.8
+python_version              = 3.9
```
clintval, line 76: We had misunderstood what this did (it locks MyPy to this interpreter version) instead of using the current interpreter.


```suggestion
```

I suggest deleting as we did in the Python template:

https://github.com/fulcrumgenomics/python-template/pull/19

clintval (approved):
If you have the time to comment I'm curious why the bump and pin to the Ubuntu versions in all our GitHub runners? Is there a reason not to use latest any longer?

.github/workflows/pythonpackage.yml
```diff
@@ -9,10 +9,10 @@ permissions:
 
 jobs:
   testing:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
nh13, line 12: See: https://fulcrumgenomics.slack.com/archives/C089VCMJR/p1732639852653859

nh13: See: #196

PR #198, opened 2024-12-06
ci: add CI support for python 3.13

nh13:


nh13:
Added in https://github.com/fulcrumgenomics/fgpyo/pull/223

PR #199, opened 2024-12-23
Add a method for progress logging multiple alignments

clintval:
Often I am iterating through a BAM by query name and find I'm having to manually loop over all records to increment the progress logger. This new method would make logging all the records in a template more convenient.

### DIFF
--- a/fgpyo/util/logging.py
+++ b/fgpyo/util/logging.py
@@ -45,8 +45,6 @@ from typing import Union
 
 from pysam import AlignedSegment
 
-from fgpyo.sam import Template
-
 # Global that is set to True once logging initialization is run to prevent running > once.
 __FGPYO_LOGGING_SETUP: bool = False
 
--- a/fgpyo/util/logging.py
+++ b/fgpyo/util/logging.py
@@ -44,6 +44,8 @@ from typing import Union
 
 from pysam import AlignedSegment
 
+from fgpyo.sam import Template
+
 # Global that is set to True once logging initialization is run to prevent running > once.
 __FGPYO_LOGGING_SETUP: bool = False
 
@@ -162,6 +164,23 @@ class ProgressLogger(AbstractContextManager):
             return self.record(None, None)
         else:
             return self.record(rec.reference_name, rec.reference_start + 1)
+
+    def record_template(
+        self,
+        template: Template,
+    ) -> bool:
+        """Correctly record all records in a fgpyo.sam.Template (zero-based coordinates).
+
+        Args:
+            template: fgpyo.sam.Template objects
+
+        Returns:
+            true if a message was logged, false otherwise
+        """
+        logged_message: bool = False
+        for rec in template.all_recs():
+            logged_message = self.record_alignment(rec) or logged_message
+        return logged_message
 
     def _log(
         self,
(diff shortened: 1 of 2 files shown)


tfenne:
I don't object to this, but I wonder if a slightly different approach would work for you and be more general:

```python
def record_alignments(
        self,
        alns: Iterable[AlignedSegment],
    ) -> bool:
        """Correctly record all records in an iterable of alignments.
        Args:
            alns: an iterable of zero or more alignments
        Returns:
            true if a message was logged, false otherwise
        """
        logged_message: bool = False
        for rec in alns:
            logged_message = self.record_alignment(rec) or logged_message
        return logged_message
```

That could then be called with `progress.record(template.all_recs())`.

clintval:
@tfenne that also works. If I went that approach I might suggest making a small API break since we already have a `record_alignment()` and supporting variadic args:

```diff
diff --git a/fgpyo/util/logging.py b/fgpyo/util/logging.py
--- a/fgpyo/util/logging.py
+++ b/fgpyo/util/logging.py
@@ -150,20 +150,23 @@ class ProgressLogger(AbstractContextManager):

     def record_alignment(
         self,
-        rec: AlignedSegment,
+        *recs: AlignedSegment,
     ) -> bool:
         """Correctly record pysam.AlignedSegments (zero-based coordinates).
```

tfenne:
> Let me know which option you like the most and I can fixup.

I'm not sure I like the variadic arg version without changing the name (e.g. to `record_alignment*s*`), which is a bigger breaking change.  And without that, I think most people would miss it.

But I'd be fine to add a `record_alignments` _and_ a `record_template` that just vectors to the former.

clintval:
@tfenne I thought about this more and chose to ditch the idea of recording "templates". Simply because I didn't want a user to think the logger was incrementing `1` for every template instead of `n` for all the records _in_ the template. So I've changed the title of the PR and added a new method called `record_alignments()` to help out with collections of alignments including templates.

fgpyo/util/logging.py
```diff
@@ -163,6 +164,23 @@ def record_alignment(
         else:
             return self.record(rec.reference_name, rec.reference_start + 1)
 
+    def record_alignments(
+        self,
+        recs: Iterable[AlignedSegment],
```
msto, line 169: I think this is a sensible and helpful addition to the library.

My only suggestion is to type this more strictly as a `Collection`, so the user does not inadvertently pass a generator which would be consumed by the function.

```suggestion
        recs: Collection[AlignedSegment],
```

clintval: @msto I do like that idea. What's awkward is almost everything on template returns an iterator even when it doesn't need to (it's just returning things it already has in memory unaltered). So, if I upgrade to `Collection` I'd have to do this:

```python
progress.record_alignments(recs=list(template.all_recs()))
```

Not the worst, but a little awkward.

clintval: I thought a bit more about this and I left the typing as the most generic thing necessary for the function to work, which is `Iterable`. The function is named "record_alignments" so it should be unsurprising that it will advance/consume/iterate-over whatever you supply it.

tfenne (approved):

PR #200, opened 2024-12-23
Add all_r1 and all_r2 methods on Template

clintval:
Having a view into the R1 alignments or R2 alignments without having to manually chain the iterables each time would be helpful for dealing with complex alignment scenarios. Would these two methods be a welcome addition?


tfenne (approved):
LGTM, but could we pluralize the functions please (i.e. `all_r1s` instead of `all_r1`) to match `all_recs`?

PR #201, opened 2024-12-24
Add a function for determining read pair orientation

clintval:
I have a project where I am re-building a complex alignment situation for a template and it involves an assessment of read pair orientation and whether the new pair is "proper" or not.

The functions in this PR allow for an optional R2 such as when you are iterating through a BAM in coordinate order and don't have easy access to the R2. When this happens, R1 is required to have enough mate information to derive the pair status and whether or not the read pair is "proper".

The implementation honors what is implemented in HTSJDK/fgbio but with:

1. Fewer exceptions by using `None` instead of an exception, when possible
2. Proper pair status also considers template length / insert size

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -622,45 +622,45 @@ class PairOrientation(enum.Enum):
 
     @classmethod
     def build(
-        cls, r1: AlignedSegment, r2: Optional[AlignedSegment] = None
+        cls, rec1: AlignedSegment, rec2: Optional[AlignedSegment] = None
     ) -> Optional["PairOrientation"]:
         """Returns the pair orientation if both reads are mapped to the same reference sequence.
 
         Args:
-            r1: The first read in the template.
-            r2: The second read in the template. If undefined, mate data set upon R1 will be used.
+            rec1: The first record in the pair.
+            rec2: The second record in the pair. If None, then mate info on `rec1` will be used.
 
         See:
             [`htsjdk.samtools.SamPairUtil.getPairOrientation()`](https://github.com/samtools/htsjdk/blob/c31bc92c24bc4e9552b2a913e52286edf8f8ab96/src/main/java/htsjdk/samtools/SamPairUtil.java#L71-L102)
         """
 
-        if r2 is None:
-            r2_is_unmapped = r1.mate_is_unmapped
-            r2_reference_id = r1.next_reference_id
+        if rec2 is None:
+            rec2_is_unmapped = rec1.mate_is_unmapped
+            rec2_reference_id = rec1.next_reference_id
         else:
-            r2_is_unmapped = r2.is_unmapped
-            r2_reference_id = r2.reference_id
+            rec2_is_unmapped = rec2.is_unmapped
+            rec2_reference_id = rec2.reference_id
 
-        if r1.is_unmapped or r2_is_unmapped or r1.reference_id != r2_reference_id:
+        if rec1.is_unmapped or rec2_is_unmapped or rec1.reference_id != rec2_reference_id:
             return None
 
-        if r2 is None:
-            if not r1.has_tag("MC"):
+        if rec2 is None:
+            if not rec1.has_tag("MC"):
                 raise ValueError('Cannot determine proper pair status without R2\'s cigar ("MC")!')
-            r2_cigar = Cigar.from_cigarstring(str(r1.get_tag("MC")))
-            r2_is_forward = r1.mate_is_forward
-            r2_reference_start = r1.next_reference_start
-            r2_reference_end = r1.next_reference_start + r2_cigar.length_on_target()
+            rec2_cigar = Cigar.from_cigarstring(str(rec1.get_tag("MC")))
+            rec2_is_forward = rec1.mate_is_forward
+            rec2_reference_start = rec1.next_reference_start
+            rec2_reference_end = rec1.next_reference_start + rec2_cigar.length_on_target()
         else:
-            r2_is_forward = r2.is_forward
-            r2_reference_start = r2.reference_start
-            r2_reference_end = r2.reference_end
+            rec2_is_forward = rec2.is_forward
+            rec2_reference_start = rec2.reference_start
+            rec2_reference_end = rec2.reference_end
 
-        if r1.is_forward is r2_is_forward:
+        if rec1.is_forward is rec2_is_forward:
             return PairOrientation.TANDEM
-        elif r1.is_forward and r1.reference_start < r2_reference_end:
+        elif rec1.is_forward and rec1.reference_start < rec2_reference_end:
             return PairOrientation.FR
-        elif r1.is_reverse and r2_reference_start < r1.reference_end:
+        elif rec1.is_reverse and rec2_reference_start < rec1.reference_end:
             return PairOrientation.FR
         else:
             return PairOrientation.RF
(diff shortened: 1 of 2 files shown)


fgpyo/sam/__init__.py
```diff
@@ -265,7 +265,8 @@ def _pysam_open(
             assert file_type is not None, "Must specify file_type when writing to standard output"
             path = sys.stdout
         else:
-            file_type = file_type or SamFileType.from_path(path)
+            if file_type is None:
+                file_type = SamFileType.from_path(path)
```
clintval, line 269: I must have a newer MyPy in my virtual environment because it complained about this pattern.

fgpyo/sam/__init__.py
```diff
@@ -706,9 +811,8 @@
         r1: read 1
         r2: read 2 with the same queryname as r1
     """
-    assert not r1.is_unmapped, f"Cannot process unmapped mate {r1.query_name}/1"
-    assert not r2.is_unmapped, f"Cannot process unmapped mate {r2.query_name}/2"
-    assert r1.query_name == r2.query_name, "Attempting to pair reads with different qnames."
+    if r1.query_name != r2.query_name:
+        raise ValueError("Cannot set pair info on reads with different query names!")
```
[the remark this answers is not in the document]
clintval: I know... but this is on a now-deprecated function in the following PR, so I'm not interested in implementing more tests.

- https://github.com/fulcrumgenomics/fgpyo/pull/202

fgpyo/sam/__init__.py
```diff
@@ -606,6 +607,110 @@ def __getitem__(
...
+
+    RF = "RF"
+    """A pair orientation for reverse-forward reads ("outie")."""
+
+    TANDEM = "TANDEM"
+    """A pair orientation for tandem (forward-forward or reverse-reverse) reads."""
+
+    @classmethod
+    def build(
+        cls, r1: AlignedSegment, r2: Optional[AlignedSegment] = None
```
tfenne, line 625: I see this pattern of `r1` and optional `r2` in several places.  I think it would be better to use different naming (e.g. `read` and `mate`) - because I don't think you actually care that the first read is `first_of_pair`, and I think this function would work called with a single `second_of_pair` and no mate.

fgpyo/sam/__init__.py
```diff
@@ -606,6 +607,110 @@ def __getitem__(
...
+    """A pair orientation for forward-reverse reads ("innie")."""
+
+    RF = "RF"
+    """A pair orientation for reverse-forward reads ("outie")."""
+
+    TANDEM = "TANDEM"
+    """A pair orientation for tandem (forward-forward or reverse-reverse) reads."""
+
+    @classmethod
+    def build(
```
tfenne, line 624: How would you feel about calling this `from()` or similar?  It feels wrong to me to have a `build()` on an enum.

fgpyo/sam/__init__.py
```diff
@@ -606,6 +607,110 @@ def __getitem__(
...
+        else:
+            r2_is_unmapped = r2.is_unmapped
+            r2_reference_id = r2.reference_id
+
+        if r1.is_unmapped or r2_is_unmapped or r1.reference_id != r2_reference_id:
+            return None
+
+        if r2 is None:
+            if not r1.has_tag("MC"):
+                raise ValueError('Cannot determine proper pair status without R2\'s cigar ("MC")!')
```
tfenne, line 649: I think we could probably narrow the cases where we raise this error a bit more.  I.e. if the "r2" start is greater than "r1" end, then we don't technically need the mate cigar.

tfenne: Likewise if r2 is positive strand, we won't need the cigar.

fgpyo/sam/__init__.py
```diff
@@ -606,6 +607,110 @@ def __getitem__(
...
+            return PairOrientation.RF
+
+
+DefaultProperlyPairedOrientations = {PairOrientation.FR}
+"""The default orientations for properly paired reads."""
+
+
+def is_proper_pair(
+    r1: AlignedSegment,
+    r2: Optional[AlignedSegment] = None,
```
tfenne, line 675: Same comment on naming

fgpyo/sam/__init__.py
```diff
@@ -606,6 +607,110 @@ def __getitem__(
...
+    max_insert_size: int = 1000,
+    orientations: set[PairOrientation] = DefaultProperlyPairedOrientations,
```
tfenne, line 677: How would you feel about making this `Iterable` and using a `tuple` for the default?

tfenne (approved):
Approved so long as you agree with my comments, otherwise let's discuss any differences!

fgpyo/sam/__init__.py
clintval: `from` is a reserved keyword in Python so that wouldn't work syntactically. I saw a `build()` on `Template` so copied that word:

https://github.com/fulcrumgenomics/fgpyo/blob/3680b4fe1fe3d8686172e05dbf9c5043fa51ae1a/fgpyo/sam/__init__.py#L858

I could do:

- `of(rec1, rec2)`
- `from_recs(rec1, rec2)`

Do you like any of those better?

fgpyo/sam/__init__.py
clintval: I agree!

fgpyo/sam/__init__.py
clintval: I like making the arg `Iterable` as it it is more generic. But a `set` default value seems to capture the intent more than a non-set (why would you allow duplicates?).

What I think I need to do is use a `Container` type to support a set default value, but other types of containers too:

https://docs.python.org/3/library/collections.abc.html#collections.abc.Container

fgpyo/sam/__init__.py
clintval: Good observation!

fgpyo/sam/__init__.py
clintval: https://github.com/fulcrumgenomics/fgpyo/pull/201/commits/4e64644fc0176bc93bfa32e4ed6f7a9e61118b06

fgpyo/sam/__init__.py
clintval: https://github.com/fulcrumgenomics/fgpyo/pull/201/commits/6f3c514ed48c8cbd6dc4f69eafa50b06f871201c

fgpyo/sam/__init__.py
```diff
@@ -606,6 +608,118 @@ def __getitem__(
...
+    """A pair orientation for forward-reverse reads ("innie")."""
+
+    RF = "RF"
+    """A pair orientation for reverse-forward reads ("outie")."""
+
+    TANDEM = "TANDEM"
+    """A pair orientation for tandem (forward-forward or reverse-reverse) reads."""
+
+    @classmethod
+    def from_recs(  # noqa: C901  # `from_recs` is too complex (11 > 10)
```
clintval, line 625: I couldn't really think of a way of simplifying this function without making harder to grok in isolation private subfunctions... and 11 vs 10 didn't seem so bad.

clintval:
@tfenne if you have time to re-review:

- The function has more branches, but fetches the mate cigar as a last resort
- I renamed the classmethod to `from_recs(rec1: AlignedSegment, rec2: AlignedSegment | None)`
- I typed `orientations` as `Collection` instead of `set`
- Instead of `r1` and `r2` I'm calling the new parameters `rec1` and `rec2`
   - I did this because they don't even technically need to be mates

fgpyo/sam/__init__.py
tfenne: Either one works for me.  I just think `build()` is strange since it's _selecting_ an enum value, not building a new one.

fgpyo/sam/__init__.py
tfenne: Honestly, I don't like the complexity check.  Sometimes things _are_ complex, and breaking it out into functions arbitrarily just makes it harder to follow what's going onl

fgpyo/sam/__init__.py
```diff
@@ -607,6 +608,118 @@ def __getitem__(
...
+        if rec1.is_forward and rec1.reference_start <= rec2_reference_start:
+            return PairOrientation.FR
+        if rec1.is_reverse and rec2_reference_start < rec1.reference_end:
+            return PairOrientation.FR
+        if rec1.is_reverse and rec2_reference_start >= rec1.reference_end:
+            return PairOrientation.RF
+
+        if rec2 is None:
+            if not rec1.has_tag("MC"):
+                raise ValueError('Cannot determine proper pair status without a mate cigar ("MC")!')
```
tfenne, line 666:
```suggestion
                raise ValueError('Cannot determine pair orientation without a mate cigar ("MC")!')
```

tfenne (approved):
LGTM

fgpyo/sam/__init__.py
clintval: I went with `from_recs(rec1, rec2)` :)

geoffjentry: I tend to go with `.from_FOO()` for these sorts of things, FWIW

fgpyo/sam/__init__.py
clintval: 🤦 copy-paste...

PR #202, opened 2024-12-24
Deprecate set_pair_info and _set_mate_info for set_mate_info

clintval:
I notice we have two implementations for setting mate info on a pair of alignments.

I've refactored for the following reasons:

- [`set_pair_info()`](https://github.com/fulcrumgenomics/fgpyo/blob/3680b4fe1fe3d8686172e05dbf9c5043fa51ae1a/fgpyo/sam/__init__.py#L700-L731):
    - Unnecessarily disallows unmapped reads
    - Uses `assert` instead of `raise`, which can be turned off at runtime
    - Does not set mate quality tag ([`MQ`](https://samtools.github.io/hts-specs/SAMtags.pdf))
    - Does not set mate score tag ([`ms`](https://www.htslib.org/doc/samtools-fixmate.html))
    - Defaults proper pair status to `True` (it's easy to forget to change this)
    - Forces you to override the "properly paired" status with True/False
- [`_set_mate_info()`](https://github.com/fulcrumgenomics/fgpyo/blob/3680b4fe1fe3d8686172e05dbf9c5043fa51ae1a/fgpyo/sam/builder.py#L267-L329):
    - Looks like a direct translation of [HTSJDK's implementation](https://github.com/samtools/htsjdk/blob/c31bc92c24bc4e9552b2a913e52286edf8f8ab96/src/main/java/htsjdk/samtools/SamPairUtil.java#L206-L287) (it can be simpler with pysam)
    - Does not always cleanup mate cigar tag ([`MC`](https://samtools.github.io/hts-specs/SAMtags.pdf))
    - Does not set mate quality tag ([`MQ`](https://samtools.github.io/hts-specs/SAMtags.pdf))
    - Does not set mate score tag ([`ms`](https://www.htslib.org/doc/samtools-fixmate.html))
    - Forces a reset of properly paired status

Now we have:

- [`set_pair_info()`](https://github.com/fulcrumgenomics/fgpyo/blob/1e066132a0fdc9af8e3b4c5dbd1633881e9f4c08/fgpyo/sam/__init__.py#L872-L884): annotated as _**deprecated**_ to avoid an API break
- [`set_mate_info()`](https://github.com/fulcrumgenomics/fgpyo/blob/1e066132a0fdc9af8e3b4c5dbd1633881e9f4c08/fgpyo/sam/__init__.py#L807-L837): a more useful function for setting mate info on alignments
- ~[`set_as_pairs()`](https://github.com/fulcrumgenomics/fgpyo/blob/1e066132a0fdc9af8e3b4c5dbd1633881e9f4c08/fgpyo/sam/__init__.py#L840-L869): this behavior was in `set_pair_info()` so it is now a sole function~

Questions?

- ~Is it worth keeping `set_as_pairs()`? What was the motivation for this functionality?~
- Since we are still pre-v1, can we remove `set_pair_info()` without deprecating first?

Along the way I found a bug in `SamBuilder.add_pair()` and fixed that too.

~If initial code review is favorable, I can go and add more unit tests.~


fgpyo/sam/builder.py
```diff
@@ -444,8 +380,16 @@ def add_pair(
...
+
+        if start1 != sam.NO_REF_POS:
+            chrom1 = next(c for c in (chrom1, chrom) if c is not None)
+        else:
+            chrom1 = sam.NO_REF_NAME
+
+        if start2 != sam.NO_REF_POS:
+            chrom2 = next(c for c in (chrom2, chrom) if c is not None)
+        else:
+            chrom2 = sam.NO_REF_NAME
```
clintval, line 392: Without these edge cases, `SamBuilder` would build partially incorrect pairs with syntax like:

```python
r1, r2 = builder.add_pair(chrom="chr1", start1=100)
```

In the above, R1 would be mapped to `chr1:100`.

R2 would be unmapped, but would still have its reference sequence name set to `chr1` which is non-sensical.

fgpyo/sam/__init__.py
```diff
@@ -802,38 +804,104 @@ def isize(r1: AlignedSegment, r2: AlignedSegment) -> int:
...
+    template_length = isize(r1, r2)
+    r1.template_length = template_length
+    r2.template_length = -template_length
+
+    proper_pair = is_proper_pair(r1, r2)
+    r1.is_proper_pair = proper_pair
+    r2.is_proper_pair = proper_pair
+
+
+def set_as_pairs(
```
tfenne, line 860: This, to me, feels like it belongs elsewhere and not in `fgpyo.sam` - I'm not sure why this behavior exists, and I don't think it's general enough that it belongs here.  Perhaps we can find when/why it was introduced and push it upstream somewhere?

fgpyo/sam/__init__.py
```diff
@@ -802,38 +804,104 @@ def isize(r1: AlignedSegment, r2: AlignedSegment) -> int:
...
+        dest.next_reference_id = src.reference_id
+        dest.next_reference_name = src.reference_name
+        dest.next_reference_start = src.reference_start
+        dest.mate_is_forward = src.is_forward
+        dest.mate_is_mapped = src.is_mapped
+        dest.set_tag("MC", src.cigarstring)
+        dest.set_tag("MQ", src.mapping_quality)
+
+    r1.set_tag("ms", sum_of_base_qualities(r2))
+    r2.set_tag("ms", sum_of_base_qualities(r1))
```
tfenne, line 849: Do we recall why we do this?  Seems like an expensive operation to do, for what is probably some edge case use case?

fgpyo/sam/__init__.py
```diff
@@ -802,38 +804,104 @@ def isize(r1: AlignedSegment, r2: AlignedSegment) -> int:
...
+        raise ValueError("Cannot set mate info on alignments with different query names!")
+
+    for src, dest in [(r1, r2), (r2, r1)]:
+        dest.next_reference_id = src.reference_id
+        dest.next_reference_name = src.reference_name
+        dest.next_reference_start = src.reference_start
+        dest.mate_is_forward = src.is_forward
+        dest.mate_is_mapped = src.is_mapped
+        dest.set_tag("MC", src.cigarstring)
+        dest.set_tag("MQ", src.mapping_quality)
```
tfenne, line 846: Does this "just work" when the source read is unmapped?  I.e. does src.cigarstring return None, causing `MC` to get removed?  Or do we end up with `MC:Z:*`?

tfenne (commented):
I'm supportive!

fgpyo/sam/__init__.py
clintval: I'll triple check, but it should be the former.

If either field is `None` (such as when the read is unmapped) then the tag is cleared.

fgpyo/sam/__init__.py
clintval: I added this as a new feature.

It is a required tag needed in a typical samtools fixmate/markdup workflow:

https://www.htslib.org/doc/samtools-fixmate.html
https://www.htslib.org/doc/samtools-markdup.html

```console
-m   Add ms (mate score) tags. These are used by markdup to select the best reads to keep.
```

I think the time/compute-expense is probably worth it given how common the samtools/markdup workflow is (I use it by default vs the same pattern in Picard). The alternative would be to use fgpyo to fix up _some_ mate information but then you'd be required to call `samtools fixmate` again to regenerate the `ms` tag.

fgpyo/sam/__init__.py
clintval: This is a new function in this PR with just the functionality I didn't want retained in `set_mate_info()`. It was inherited from samwell: https://github.com/fulcrumgenomics/fgpyo/commit/314fe06fa24f789b8a724e088d7b8e50ed18ec74

I propose we don't make it a new function.

What should we do with the deprecated `set_pair_info()`? Retain it's function as-is and mark it deprecated, or delete it also?

fgpyo/sam/__init__.py
clintval: I confirm these tests pass when `src` is unmapped:

```python
assert not dest.has_tag("MC")

assert dest.has_tag("MQ")
assert dest.get_tag("MQ") == 0

assert dest.has_tag("ms")
assert dest.get_tag("ms") == 3000
```

Since a record's mapping quality is non-optional according to the SAM spec (§ 1.4 Table 1) it makes sense to set the mate's mapping quality to the default value of 0 for an unmapped read, which is what happens. The tag `ms` is the sum of >=15 base qualities as needed by a typical samtools markdup workflow and is not contingent on whether the read is mapped or not, so it also makes sense it would be set for an unmapped read.

fgpyo/sam/__init__.py
clintval: Deleted this solo function.

fgpyo/sam/__init__.py
tfenne: The ripple effects of a decision I made 12+ years ago.  Argh!

tfenne (approved):

PR #203, opened 2024-12-26
Template to treat 'secondary supplemental' as supplemental

clintval:
Currently, `Template` treats "secondary supplemental" alignments as both secondary alignments and as supplemental alignments. Before this PR, `Template` would _**add**_ the "secondary supplemental" alignments into both the secondary and supplementary lists, thereby duplicating them. This behavior makes it difficult to use `Template` in complex alignment scenarios since there are duplicated records across both the secondary and supplementary lists e.g. records become duplicated when writing to a BAM file with `template.write_to(writer)`.

It would be more intuitive to me that supplemental alignments of a secondary alignment belong in the "supplementary" bin instead of both the secondary and the supplementary bin. With this PR, the fields of Template become more specific:

- Primary: all primary non-supplementary alignments
- Secondary: all secondary non-supplementary alignments
- Supplementary: all supplementary alignments

For context, I have a situation where I am intentionally creating and comparing primary and secondary alignments, each having a chimeric alignment (e.g. an alignment across different reference sequences), and having to de-duplicate the supplemental records from both the secondary and supplementary bins is error-prone and a lot of work.

### DIFF
--- a/tests/fgpyo/sam/test_template_iterator.py
+++ b/tests/fgpyo/sam/test_template_iterator.py
@@ -131,6 +131,51 @@ def test_write_template(
         assert len([r for r in template.all_recs()]) == 2
 
 
+def test_template_treats_secondary_supplementary_as_supplementary() -> None:
+    """Test that Template treats "secondary supplementaries" as supplementary."""
+    builder = SamBuilder()
+
+    r1, r2 = builder.add_pair(name="x", chrom="chr1", start1=10, start2=30)
+    r1_secondary, r2_secondary = builder.add_pair(name="x", chrom="chr1", start1=2, start2=5)
+    r1_secondary.is_secondary = True
+    r2_secondary.is_secondary = True
+
+    r1_supp, r2_supp = builder.add_pair(name="x", chrom="chr1", start1=2, start2=3)
+    r1_supp.is_supplementary = True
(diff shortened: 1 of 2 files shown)


tests/fgpyo/sam/test_template_iterator.py
```diff
@@ -131,6 +131,51 @@ def test_write_template(
...
+    )
+    expected = Template(
+        name="x",
+        r1=r1,
+        r2=r2,
+        r1_secondaries=[r1_secondary],
+        r2_secondaries=[r2_secondary],
+        r1_supplementals=[r1_supp, r1_secondary_supp],
+        r2_supplementals=[r2_supp, r2_secondary_supp],
+    )
```
clintval, line 175: Previously, the template would look like:

```python
expected = Template(
    name="x",
    r1=r1,
    r2=r2,
    r1_secondaries=[r1_secondary, r1_secondary_supp],
    r2_secondaries=[r2_secondary, r2_secondary_supp],
    r1_supplementals=[r1_supp, r1_secondary_supp],
    r2_supplementals=[r2_supp, r2_secondary_supp],
)
```

tfenne (approved):
Could you please also update the class level docs (and/or attributes) to clearly explain this?

PR #204, opened 2024-12-27
Add methods to fix mate info on non-primaries and templates

clintval:
Closes https://github.com/fulcrumgenomics/fgpyo/issues/83
Closes https://github.com/fulcrumgenomics/fgpyo/issues/182

I need a set of functions to make it easier to set mate info on secondary, supplemental, "secondary supplemental", and templates. The work in this PR builds on the work in:

- https://github.com/fulcrumgenomics/fgpyo/pull/201
- https://github.com/fulcrumgenomics/fgpyo/pull/202
- https://github.com/fulcrumgenomics/fgpyo/pull/203
- https://github.com/fulcrumgenomics/fgpyo/pull/205

This PR introduces:

- [`set_mate_info_on_secondary`](https://github.com/fulcrumgenomics/fgpyo/blob/32024b9190aee61e02b96c89bc54c0ed6ac14cc1/fgpyo/sam/__init__.py#L879-L914)
- [`set_mate_info_on_supplementary`](https://github.com/fulcrumgenomics/fgpyo/blob/32024b9190aee61e02b96c89bc54c0ed6ac14cc1/fgpyo/sam/__init__.py#L917C5-L948)
- ~[`set_mate_info_for_template`](https://github.com/fulcrumgenomics/fgpyo/blob/32024b9190aee61e02b96c89bc54c0ed6ac14cc1/fgpyo/sam/__init__.py#L1294-L1316)~
- [`Template.set_mate_info()`](https://github.com/fulcrumgenomics/fgpyo/blob/32024b9190aee61e02b96c89bc54c0ed6ac14cc1/fgpyo/sam/__init__.py#L1232-L1234)

All "mate" fields are set using the primary of the opposite read ordinal.

However, I am seeking advice on how we should set:

- `is_proper_pair`
- `template_length`

###### For Secondary Alignments

For secondary alignments, I personally am curious about those values not as they are computed on the two primary alignments, but between the secondary alignment and the primary of the opposite ordinal (e.g. if this secondary _**was**_ primary, then what would it look like paired with the primary of the opposite ordinal).

###### For Supplementary Alignments

For supplementary alignments of the _**primary**_ alignment, it makes sense to have these two values point to the primary alignment of the mate. However, for _**secondary**_ supplemental alignments, I think these values are undefined without knowing more about the specific alignment situation, so I am choosing to leave them as-is.


clintval:
If early review is favorable, I'll go and finish the PR with unit tests!

fgpyo/sam/__init__.py
```diff
@@ -876,6 +876,78 @@ def set_mate_info(
     r2.is_proper_pair = proper_pair
 
 
+def set_mate_info_on_secondary(
+    primary: AlignedSegment,
+    secondary: AlignedSegment,
```
tfenne, line 881: I think I would call the first argument "mate_primary" or something to make it clear you're supposed to provide the primary of the mate, not the primary of the secondary.

fgpyo/sam/__init__.py
```diff
@@ -876,6 +876,78 @@ def set_mate_info(
...
+    is_proper_pair: Callable[[AlignedSegment, AlignedSegment], bool] = is_proper_pair,
+) -> None:
+    """Set mate info on a secondary alignment to the next read ordinal's primary alignment.
```
tfenne, line 884:
```suggestion
    """Set mate info on a secondary alignment to its mate's primary alignment.
```

fgpyo/sam/__init__.py
```diff
@@ -876,6 +876,78 @@ def set_mate_info(
...
+        raise ValueError("Cannot set mate info on alignments with different query names!")
+
+    secondary.next_reference_id = primary.reference_id
+    secondary.next_reference_name = primary.reference_name
+    secondary.next_reference_start = primary.reference_start
+    secondary.mate_is_forward = primary.is_forward
+    secondary.mate_is_mapped = primary.is_mapped
+    secondary.set_tag("MC", primary.cigarstring)
+    secondary.set_tag("MQ", primary.mapping_quality)
+    secondary.set_tag("ms", sum_of_base_qualities(primary))
```
tfenne, line 910: Could all of this be extracted into a `_transfer_mate_fields()` that is used by this function, `set_mate_info()` and also `set_mate_info_on_supplementary()`; i.e. put all the one-way transfer into a function and re-use that.

fgpyo/sam/__init__.py
```diff
@@ -876,6 +876,78 @@ def set_mate_info(
...
+
+    # NB: calculate isize and proper pair as if this secondary alignment was the primary alignment.
+    secondary.is_proper_pair = is_proper_pair(primary, secondary)
+    secondary.template_length = isize(primary, secondary)
+
+
+def set_mate_info_on_supplementary(primary: AlignedSegment, supp: AlignedSegment) -> None:
+    """Set mate info on a supplementary alignment to the next read ordinal's primary alignment.
```
tfenne, line 918:
```suggestion
    """Set mate info on a supplementary alignment to its mate's primary alignment.
```

tfenne (commented):
I'm honestly not sure what to make of this PR.  I think there is in fact a correct way to do all of this for a Template, but it's complicated and I don't know of an aligner that a) produces the necessary information and b) produces alignments that would require a full solution.

My 2c is that a fully correct solution:

- Would need to group records into "distinct template alignments", e.g. using the `HI` (hit index) tag.  One group would be "primary" and all subsequent groups would be secondary
- All processing then occurs within a group, including:
  - TLEN/isize being calculated between the two non-supplementary records within the group
  - is_proper_pair also on the same two records as TLEN
  - mate info is set, for each record, from the non-supplementary mate record within the group

clintval:
> I'm honestly not sure what to make of this PR. I think there is in fact a correct way to do all of this for a Template, but it's complicated and I don't know of an aligner that a) produces the necessary information and b) produces alignments that would require a full solution.
> 
> My 2c is that a fully correct solution:
> 
> * Would need to group records into "distinct template alignments", e.g. using the `HI` (hit index) tag.  One group would be "primary" and all subsequent groups would be secondary
> * All processing then occurs within a group, including:
>   
>   * TLEN/isize being calculated between the two non-supplementary records within the group
>   * is_proper_pair also on the same two records as TLEN
>   * mate info is set, for each record, from the non-supplementary mate record within the group

I agree with your approach on how to do it in the **_most correct way_**, but I think this PR is a step in the right direction and isn't technically incorrect. At least, it does as much as it can with the information it has on hand (_`bwa` doesn't emit the_ `HI` _tag so I would need more code to derive that_).

I think the special case I have for `set_mate_info_on_supplementary` is more technically correct since it correctly sets the mate info for the primary supplements to the primary of the mate:

```python
if not supp.is_secondary:
    supp.is_proper_pair = mate_primary.is_proper_pair
    supp.template_length = -mate_primary.template_length
```

That, I think we should keep.

How I'm setting the values for secondary alignments is a bit unique and could be removed. In many practical cases I tend to see only one of two reads multi-mapping, so I end up caring more about how the secondaries relate back to the single (and primary) alignment of the mate.

fgpyo/sam/__init__.py
```diff
@@ -863,6 +863,28 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
...
 
 
+def _set_common_mate_fields(dest: AlignedSegment, source: AlignedSegment) -> None:
+    """Set common mate info an alignment to its mate's primary alignment.
+
+    Args:
+        dest: The alignment to set the mate info upon.
+        source: The primary alignment to use as a mate reference.
+    """
+    if source.is_read1 is dest.is_read1 or source.is_secondary or source.is_supplementary:
```
tfenne, line 873: I think we could/should remove `source.is_secondary` from this `if`.

fgpyo/sam/__init__.py
```diff
@@ -1221,6 +1288,31 @@ def __next__(self) -> Template:
         return Template.build(recs, validate=False)
 
 
+def set_mate_info_for_template(
```
tfenne, line 1291: I think I'd prefer this to be an instance method on `Template`, what do you think?

tfenne (commented):
I agree that the code for setting values on supplementaries is a good addition.  I'm still a little wary of the choices for secondaries ... if anyone is using an aligner that is better than bwa at reporting secondary alignments, this might override more useful information with arguably incorrect values.

fgpyo/sam/__init__.py
clintval: Yes, good find!

fgpyo/sam/__init__.py
clintval: Currently I have it as both. But since the function always requires a template it doesn't make sense to have both forms of the same function since they don't vary in any other way. I'll add this only to the template for now. If someone wants a standalone function, they can make an argument and refactor later.

tests/fgpyo/sam/test_sam.py
```diff
@@ -804,3 +827,202 @@ def test_set_mate_info_both_mapped() -> None:
...
+    assert r1.reference_start == 200
+    assert r1.next_reference_start == 300
+    assert r2.reference_start == 300
+    assert r2.next_reference_start == 200
+    assert r1.template_length == 200
+    assert r2.template_length == -200
+    assert r1.is_forward is True
+    assert r2.is_reverse is True
+    assert r1.mate_is_reverse is True
+    assert r2.mate_is_forward is True
```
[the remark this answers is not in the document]
clintval: No thank you.

tfenne:
@nh13 can you take a look through this one and give your opinions please?

fgpyo/sam/__init__.py
```diff
@@ -863,40 +864,111 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
     return score
 
 
+def _set_common_mate_fields(dest: AlignedSegment, source: AlignedSegment) -> None:
+    """Set common mate info on a destination alignment from its mate's non-supplementary alignment.
```
nh13, line 868: Having an offline discussion in Slack.  I think you want the mate's non-secondary and non-supplementary alignment.

fgpyo/sam/__init__.py
```diff
@@ -863,40 +864,111 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
...
+        raise ValueError("Cannot set mate info on alignments with different query names!")
+
+    dest.next_reference_id = source.reference_id
+    dest.next_reference_name = source.reference_name
+    dest.next_reference_start = source.reference_start
+    dest.mate_is_forward = source.is_forward
+    dest.mate_is_mapped = source.is_mapped
+    dest.set_tag("MC", source.cigarstring)
+    dest.set_tag("MQ", source.mapping_quality)
+    dest.set_tag("ms", sum_of_base_qualities(source))
```
nh13, line 893: thank-you!

fgpyo/sam/__init__.py
```diff
@@ -863,40 +864,111 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
...
+
+
 def set_mate_info(
-    r1: AlignedSegment,
-    r2: AlignedSegment,
+    rec1: AlignedSegment,
+    rec2: AlignedSegment,
```
nh13, line 898: This breaks the API?

fgpyo/sam/__init__.py
```diff
@@ -863,40 +864,111 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
...
+        raise ValueError("The mate primary must not be secondary or supplementary!")
+    if not supp.is_supplementary:
+        raise ValueError("Cannot set mate info on an alignment not marked as supplementary!")
+
+    _set_common_mate_fields(dest=supp, source=mate_primary)
+
+    # NB: for a non-secondary supplemental alignment, set the following to the same as the primary.
+    if not supp.is_secondary:
+        supp.is_proper_pair = mate_primary.is_proper_pair
+        supp.template_length = -mate_primary.template_length
```
nh13, line 972: I don't think template-length is correct, since the `mate_primary.template_length` is calculated from the `primary` relative to `supp`.  Should we recalculate it, or should we set it to zero since it's not the primary-primary alignment?

fgpyo/sam/__init__.py
```diff
@@ -1164,6 +1236,31 @@ def all_recs(self) -> Iterator[AlignedSegment]:
...
+            for rec in self.r2_secondaries:
+                set_mate_info_on_secondary(secondary=rec, mate_primary=self.r1)
+            for rec in self.r2_supplementals:
+                set_mate_info_on_supplementary(supp=rec, mate_primary=self.r1)
+        if self.r2 is not None:
+            for rec in self.r1_secondaries:
+                set_mate_info_on_secondary(secondary=rec, mate_primary=self.r2)
+            for rec in self.r1_supplementals:
+                set_mate_info_on_supplementary(supp=rec, mate_primary=self.r2)
+        return self
```
nh13, line 1262: I was expecting this:
```suggestion
        if self.r1 is not None:
            for rec in self.all_r2s:
                _set_common_mate_fields(dest=rec, source=self.r1)
        if self.r2 is not None:
            for rec in self.all_r1s:
                _set_common_mate_fields(dest=rec, source=self.r2)
        return self
```

fgpyo/sam/__init__.py
clintval: Yes! You are right. I'm never using a secondary alignment for `source` so all I need to do here is refine the documentation and add another assert to make sure no one sends in a secondary to `source`.

fgpyo/sam/__init__.py
clintval: Hmm, I'm having a hard time seeing how this is wrong. Really wish we could whiteboard!

Here, "mate primary" is the primary alignment of the supplemental's mate. What we want to do is have the supplement's template length be equal to the primary of the supplement (which we unfortunately do not have access to). However, the implementation [in `fgpyo`](https://github.com/fulcrumgenomics/fgpyo/blob/b0b4227de8acff8a47f48d2afb06190c1b9850ad/fgpyo/sam/__init__.py#L894-L895) ([and elsewhere](https://github.com/samtools/htsjdk/blob/b16416c078e2df3e3a21486207d0c0c31b293cb9/src/main/java/htsjdk/samtools/SamPairUtil.java#L272-L273)) flips the sign of the template length for mates. So, if we flip the sign of the supplemental's mate primary, then we should have the value that would have been set upon the supplemental's primary.

For secondary supplementals (and for secondaries in general) I am not setting template length anywhere because I feel the specification leaves this ambiguous, and a smarter aligner (or custom code) might set specific template lengths on the secondaries and I don't want to override that info. I feel the spec. is ambiguous for secondaries because I don't see reference to `RNAME`/`RNEXT` in the definition of `TLEN`.

fgpyo/sam/__init__.py
clintval: With my last commit, I still have two separate functions for secondaries and supplementaries. They aren't that different from `_set_common_mate_fields` except for:

- Additional checking for if the record is secondary or supplementary for safety sake
- Proper pair and template length settings on supplementaries (but only for non-secondary supplementaries).

fgpyo/sam/__init__.py
nh13: I think you're right.

nh13 (approved):
Approved after discussing in person

PR #205, opened 2024-12-27
Allow insert size calculation to work on 1 read only

clintval:
I have a use case where I'd like to (re)-calculate the isize of an alignment but don't have access to the R2 alignment (such as in coordinate sorted alignments) and would like to use the mate information on R1 to provide the R2 information.


fgpyo/sam/__init__.py
```diff
@@ -668,6 +668,35 @@ def build(
             return PairOrientation.RF
 
 
+def isize(r1: AlignedSegment, r2: Optional[AlignedSegment] = None) -> int:
```
tfenne, line 671: Same comment about r1/r2 => read/mate.

fgpyo/sam/__init__.py
```diff
@@ -668,6 +668,35 @@ def build(
...
+    """Computes the insert size for a pair of records."""
```
tfenne, line 672: Could you update docs to explain that the second arg is optional and when it will fail if the second record isn't provided?

fgpyo/sam/__init__.py
```diff
@@ -668,6 +668,35 @@ def build(
...
+    else:
+        r2_is_unmapped = r2.is_unmapped
+        r2_reference_id = r2.reference_id
+
+    if r1.is_unmapped or r2_is_unmapped or r1.reference_id != r2_reference_id:
+        return 0
+
+    if r2 is None:
+        if not r1.has_tag("MC"):
+            raise ValueError('Cannot determine proper pair status without R2\'s cigar ("MC")!')
```
tfenne, line 685: Same though here as in another PR - we only need "MC" if a) the mate's start position is < the first read's end position _and_ b) the mate is on the negative strand

tfenne (approved):

PR #206, opened 2024-12-28
Add a class for parsing secondary alignments from XA/XB

clintval:
I have a project where I am "reconstituting" the secondary alignments from `bwa` into actual SAM records. I created a helper class to do this for data stored in the `XB` SAM tag and extended it to also parse the more common `XA` SAM tag.

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -205,6 +205,7 @@ import pysam
 from pysam import AlignedSegment
 from pysam import AlignmentFile as SamFile
 from pysam import AlignmentHeader as SamHeader
+from typing_extensions import Self
 from typing_extensions import deprecated
 
 import fgpyo.io
@@ -1328,7 +1329,7 @@ class SecondaryAlignment:
             raise ValueError(f"Mapping quality cannot be <0! Found: {self.mapq}")
 
     @classmethod
-    def from_tag_part(cls, part: str) -> "SecondaryAlignment":
+    def from_tag_part(cls, part: str) -> Self:
         """Build a secondary alignment from a single `XA` or `XB` tag part.
 
         Args:
@@ -1359,7 +1360,7 @@ class SecondaryAlignment:
         )
 
     @classmethod
-    def many_from_tag(cls, value: str) -> list["SecondaryAlignment"]:
+    def many_from_tag(cls, value: str) -> list[Self]:
         """Build many secondary alignments from a single `XA` or `XB` tag value.
 
         Args:
@@ -1368,13 +1369,13 @@ class SecondaryAlignment:
         return [cls.from_tag_part(part) for part in value.rstrip(";").split(";")]
 
     @classmethod
-    def many_from_rec(cls, rec: AlignedSegment) -> list["SecondaryAlignment"]:
+    def many_from_rec(cls, rec: AlignedSegment) -> list[Self]:
         """Build many secondary alignments from a single SAM record.
 
         Args:
             rec: The SAM record to generate secondary alignments from.
         """
-        secondaries: list["SecondaryAlignment"] = []
+        secondaries: list[Self] = []
         if rec.has_tag("XA"):
             secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XA"))))
         if rec.has_tag("XB"):
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -1271,7 +1271,7 @@ class SecondaryAlignment:
     Format of a single secondary alignment in the `XA` tag (`,`-delimited):
 
     ```text
-    chr:<orientation><position>,cigar,NM
+    chr,<orientation><position>,cigar,NM
     ```
 
     Full example of an `XA` tag value (`;`-delimited):
@@ -1283,7 +1283,7 @@ class SecondaryAlignment:
     Format of a single secondary alignment in the `XB` tag (`,`-delimited):
 
     ```text
-    chr:<orientation><position>,cigar,NM,AS,MapQ
+    chr,<orientation><position>,cigar,NM,AS,MapQ
     ```
 
     Full example of an `XB` tag value (`;`-delimited):
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -1319,13 +1319,13 @@ class SecondaryAlignment:
     def __post_init__(self) -> None:
         """Perform post-initialization validation on this dataclass."""
         if self.reference_start < 0:
-            raise ValueError(f"Start cannot be less zero! Found: {self.reference_start}")
+            raise ValueError(f"Start cannot be <0! Found: {self.reference_start}")
         if self.edit_distance < 0:
-            raise ValueError(f"Edit distance cannot be less zero! Found: {self.edit_distance}")
+            raise ValueError(f"Edit distance cannot be <0! Found: {self.edit_distance}")
         if self.alignment_score is not None and self.alignment_score < 0:
-            raise ValueError(f"Alignment score cannot be less zero! Found: {self.alignment_score}")
+            raise ValueError(f"Alignment score cannot be <0! Found: {self.alignment_score}")
         if self.mapq is not None and self.mapq < 0:
-            raise ValueError(f"Mapping quality cannot be less zero! Found: {self.mapq}")
+            raise ValueError(f"Mapping quality cannot be <0! Found: {self.mapq}")
 
     @classmethod
     def from_tag_part(cls, part: str) -> "SecondaryAlignment":
@@ -1365,7 +1365,7 @@ class SecondaryAlignment:
         Args:
             value: A single `XA` or `XB` tag value.
         """
-        return list(map(cls.from_tag_part, value.rstrip(";").split(";")))
+        return [cls.from_tag_part(part) for part in value.rstrip(";").split(";")]
 
     @classmethod
     def many_from_rec(cls, rec: AlignedSegment) -> list["SecondaryAlignment"]:
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -1272,7 +1272,7 @@ class SecondaryAlignment:
     Format of a single secondary alignment in the `XA` tag (`,`-delimited):
 
     ```text
-    chr,<orientation><position>,cigar,NM
+    chr,<orientation><1-based position>,cigar,NM
     ```
 
     Full example of an `XA` tag value (`;`-delimited):
@@ -1284,7 +1284,7 @@ class SecondaryAlignment:
     Format of a single secondary alignment in the `XB` tag (`,`-delimited):
 
     ```text
-    chr,<orientation><position>,cigar,NM,AS,MapQ
+    chr,<orientation><1-based position>,cigar,NM,AS,MapQ
     ```
 
     Full example of an `XB` tag value (`;`-delimited):
@@ -1295,7 +1295,9 @@ class SecondaryAlignment:
 
     Args:
         reference_name: The reference sequence name.
-        reference_start: The 0-based start position of the alignment.
+        reference_start: The 0-based start position of the alignment. Note that the start position
+            in the `XA` and `XB` tags text is 1-based; this dataclass contains a 0-based
+            representation consistent with Python and pysam convention.
         is_forward: If the alignment is in the forward orientation or not.
         cigar: The Cigar sequence representing the alignment.
         edit_distance: The number of mismatches between the query and the target.
@@ -1307,6 +1309,12 @@ class SecondaryAlignment:
         - [https://github.com/lh3/bwa/pull/292](https://github.com/lh3/bwa/pull/292)
         - [https://github.com/lh3/bwa/pull/293](https://github.com/lh3/bwa/pull/293)
         - [https://github.com/lh3/bwa/pull/355](https://github.com/lh3/bwa/pull/355)
+
+    Raises:
+        ValueError: If either reference_start is set to a value less than zero.
+        ValueError: If either edit_distance is set to a value less than zero.
+        ValueError: If either alignment_score is set to a value less than zero.
+        ValueError: If either mapq is set to a value less than zero.
     """
 
     reference_name: str
@@ -1319,14 +1327,17 @@ class SecondaryAlignment:
 
     def __post_init__(self) -> None:
         """Perform post-initialization validation on this dataclass."""
+        errors: list[str] = []
         if self.reference_start < 0:
-            raise ValueError(f"Start cannot be <0! Found: {self.reference_start}")
+            errors.append(f"Reference start cannot be <0! Found: {self.reference_start}")
         if self.edit_distance < 0:
-            raise ValueError(f"Edit distance cannot be <0! Found: {self.edit_distance}")
+            errors.append(f"Edit distance cannot be <0! Found: {self.edit_distance}")
         if self.alignment_score is not None and self.alignment_score < 0:
-            raise ValueError(f"Alignment score cannot be <0! Found: {self.alignment_score}")
+            errors.append(f"Alignment score cannot be <0! Found: {self.alignment_score}")
         if self.mapq is not None and self.mapq < 0:
-            raise ValueError(f"Mapping quality cannot be <0! Found: {self.mapq}")
+            errors.append(f"Mapping quality cannot be <0! Found: {self.mapq}")
+        if len(errors) > 0:
+            raise ValueError("\n".join(errors))
 
     @classmethod
     def from_tag_part(cls, part: str) -> Self:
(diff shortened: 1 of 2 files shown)


fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+
+@dataclass(frozen=True)
+class SecondaryAlignment:
+    """A secondary alignment as encoded in one part of the `XA` or `XB` tag value on a SAM record.
+
+    Format of a single secondary alignment in the `XA` tag (`,`-delimited):
+
+    ```text
+    chr,<orientation><position>,cigar,NM
+    ```
```
msto, line 1276: **question** `fgpyo` already provides a `SupplementaryAlignment` class for parsing the `XA` tag. I've run into a few cases where it'd be helpful to have an analogous `SecondaryAlignment` class, and I think this PR will be a welcome addition. What are the scenarios where it would make sense to use `SecondaryAlignment` instead of `SupplementaryAlignment` to parse the `XA` tag? 

I do find it sensible to have separate classes to represent the distinct tag formats and the distinct semantic meanings between the two types of alignment, but I'm wondering if there are use cases where it'd be helpful to process both types of alignment in tandem.

https://github.com/fulcrumgenomics/fgpyo/blob/bd3889807834f70078e35f78994acce54eefbe2b/fgpyo/sam/__init__.py#L768-L769

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+    Full example of an `XA` tag value (`;`-delimited):
+
+    ```text
+    XA:Z:chr9,-104599381,49M,4;chr3,+170653467,49M,4;chr12,+46991828,49M,5;
+    ```
+
+    Format of a single secondary alignment in the `XB` tag (`,`-delimited):
+
+    ```text
+    chr,<orientation><position>,cigar,NM,AS,MapQ
```
msto, line 1287:
```suggestion
    chr,<orientation><1-based position>,cigar,NM,AS,MapQ
```

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+
+    Full example of an `XB` tag value (`;`-delimited):
+
+    ```text
+    XB:Z:chr9,-104599381,49M,4,0,30;chr3,+170653467,49M,4,0,20;chr12,+46991828,49M,5,0,10;
+    ```
+
+    Args:
+        reference_name: The reference sequence name.
+        reference_start: The 0-based start position of the alignment.
```
msto, line 1298:
```suggestion
        reference_start: The **0-based** start position of the alignment. Note that the start position in the XA and XB tag text is 1-based; this dataclass contains a 0-based representation.
```

**suggestion** I think this behavior is appropriate and sensible, as it's consistent with how pysam represents alignment records (SAM is 1-based, alignment records are 0-based), but I think it'd be helpful to have a short note for clarity.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+    def __post_init__(self) -> None:
+        """Perform post-initialization validation on this dataclass."""
+        if self.reference_start < 0:
+            raise ValueError(f"Start cannot be <0! Found: {self.reference_start}")
+        if self.edit_distance < 0:
+            raise ValueError(f"Edit distance cannot be <0! Found: {self.edit_distance}")
+        if self.alignment_score is not None and self.alignment_score < 0:
+            raise ValueError(f"Alignment score cannot be <0! Found: {self.alignment_score}")
+        if self.mapq is not None and self.mapq < 0:
+            raise ValueError(f"Mapping quality cannot be <0! Found: {self.mapq}")
```
msto, line 1329: **suggestion** I think it'd be helpful to collect all possible error messages and raise at the end.

It'd also be helpful to include the query name - this info isn't in the scope of the dataclass constructor, so I'd wrap the constructions in the various classmethods below in a try-except to include the query name when constructing from AlignedSegment


```suggestion
        errs: list[str] = []
        if self.reference_start < 0:
            errs.append(f"Start cannot be <0! Found: {self.reference_start}")
        if self.edit_distance < 0:
            errs.append(f"Edit distance cannot be <0! Found: {self.edit_distance}")
        if self.alignment_score is not None and self.alignment_score < 0:
            errs.append(f"Alignment score cannot be <0! Found: {self.alignment_score}")
        if self.mapq is not None and self.mapq < 0:
            errs.append(f"Mapping quality cannot be <0! Found: {self.mapq}")
            
        if errs:
            raise ValueError("\n".join(errs))
```

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+
+    @classmethod
+    def from_tag_part(cls, part: str) -> Self:
+        """Build a secondary alignment from a single `XA` or `XB` tag part.
+
+        Args:
+            part: A single element in an `XA` or `XB` tag value.
```
msto, line 1336: **suggestion** Document the errors raised by the constructor during parsing and validation. 

```suggestion
            part: A single element in an `XA` or `XB` tag value.
            
        Raises:
            ValueError: If an `XA` tag part does not have four elements (reference name, position, CIGAR, and edit distance) or an `XB` tag part does not have six elements (reference name, position, CIGAR, edit distance, alignment score, and MAPQ)
            ValueError: If the tag part's position is not represented as a stranded integer (e.g. `+20` or `-150`)
```

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+    ```
+
+    Args:
+        reference_name: The reference sequence name.
+        reference_start: The 0-based start position of the alignment.
+        is_forward: If the alignment is in the forward orientation or not.
+        cigar: The Cigar sequence representing the alignment.
+        edit_distance: The number of mismatches between the query and the target.
+        alignment_score: The aligner-reported alignment score, if available.
+        mapq: The aligner-reported probability of an incorrect mapping, if available.
```
msto, line 1303: **suggestion** We should document the requirements of the post-init method

```suggestion
        mapq: The aligner-reported probability of an incorrect mapping, if available.
        
    Raises:
            ValueError: If the start position (encoded as 1-based in the tag part) is non-positive.
            ValueError: If the tag part's edit distance, alignment score, or MAPQ are negative.
```

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+        """Build many secondary alignments from a single SAM record.
+
+        Args:
+            rec: The SAM record to generate secondary alignments from.
+        """
+        secondaries: list[Self] = []
+        if rec.has_tag("XA"):
+            secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XA"))))
+        if rec.has_tag("XB"):
+            secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XB"))))
```
msto, line 1382: **suggestion** Where it's possible to do so, I think it's helpful to include the query name when a record is malformatted

```suggestion
        if rec.has_tag("XB"):
            try:
                secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XB"))))
            except ValueError as e:
                raise ValueError(f"Could not parse secondary alignments for read: {rec.query_name}\n{e}") from e
```

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,212 @@ class SamOrder(enum.Enum):
...
+            secondary.query_qualities = query_qualities
+            secondary.is_read1 = rec.is_read1
+            secondary.is_read2 = rec.is_read2
+            secondary.is_duplicate = rec.is_duplicate
+            secondary.is_paired = rec.is_paired
+            secondary.is_proper_pair = False
+            secondary.is_qcfail = rec.is_qcfail
+            secondary.is_forward = hit.is_forward
+            secondary.is_secondary = True
+            secondary.is_supplementary = False
```
msto, line 1440: **issue** In the current implementation, `cls.many_from_rec()` rehydrates both secondary alignments from the `XB` tag and supplementary alignments from the `XA` tag into the alignment records processed by this method. Supplementary alignments should be flagged appropriately.

msto (changes requested):
This is really nice and I'd love to see it merged in. I've run into a few cases where it'd be helpful to have a `SecondaryAlignment` class in addition to the existing `SupplementaryAlignment` class.

The only blocking issue I've identified is the handling of rehydrated secondary and supplementary alignments. I'm not certain that supplementary alignments (from the XA tag) _should_ be handled by this class, but if there are use cases where it would be helpful to do so, we should maintain the distinction (and correct SAM flags) between the two types of alignment when rehydrating.

I'd also suggest that the `SecondaryAlignment` and `SupplementaryAlignment` classes standardize on attribute and method names, where possible. (e.g. `SecondaryAlignment.many_from_rec()` and `SupplementaryAlignment.from_read()` have analogous signatures and behaviors - IMO they should have the same name.) I don't have a strong opinion on _what_ the standardized names are, just that they should match between the two classes.

fgpyo/sam/__init__.py
msto: Also on this topic - with our usual alignment settings, `bwa` will output alignment records for supplementary alignments. Do we need to add logic to ensure that the rehydration does not create redundant records?

fgpyo/sam/__init__.py
clintval: All good points! FYI the `SupplementaryAlignment` class is meant to parse the `SA` tag and the `SecondaryAlignment` classes parses the two "sibling" tags that bwa makes for secondaries (`XA` and `XB`).

fgpyo/sam/__init__.py
clintval: See https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1899623157

Supplementaries come from the `SA` tag.

Bwa can be run in a mode where it makes the `XA` tag for secondaries _**or**_ a mode where it makes the `XB` tag for secondaries. I didn't bother implementing logic distinguishing the two since it's highly unlikely a BAM will have both tags on the _**same record**_. But even if you did, I don't see that as a bug, just as an odd use of using both tag formats simultaneously.

clintval:
> I'd also suggest that the SecondaryAlignment and SupplementaryAlignment classes standardize on attribute and method names, where possible. (e.g. SecondaryAlignment.many_from_rec() and SupplementaryAlignment.from_read() have analogous signatures and behaviors - IMO they should have the same name.) I don't have a strong opinion on what the standardized names are, just that they should match between the two classes.

I agree, @tfenne I favor the new method names since it is more clear when "many" of a class are being built by a classmethod. I also like `rec` better than `read`, because I sometimes have SAM records which are not actually "reads" in the pedantic sense. Would you be OK if I added similarly named methods onto `SupplementaryAlignment` and deprecated the older methods with a `@deprecated()` annotation?

msto (approved):

tfenne:
What would you think about combining `SupplementaryAlignment` and `SecondaryAlignment` into a single class?  They are nearly identical (as is the parsing code for the various tags).  And then e.g.  having SupplementaryAlignment being a deprecated sublcass of the common class?

I feel like the hardest part would just be in naming the single class, some suggestions:

- `AuxiliaryAlignment`
- `AlignedSegmentLite`
- `SamAlignment`

clintval:
> What would you think about combining SupplementaryAlignment and SecondaryAlignment into a single class?

I think I like the idea. It would definitely help clean up the two interfaces.

I might need to try a refactor and see what it looks like which shouldn't be too hard. I'll sketch something out with a PR on top of this PR (maybe later in the week though).

clintval:
Draft alternate PR opened here (branched off this branch):

- https://github.com/fulcrumgenomics/fgpyo/pull/208

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+        """
+        if (
+            rec.is_unmapped
+            or rec.cigarstring is None
+            or rec.query_sequence is None
+            or rec.query_qualities is None
+        ):
+            return
+
+        for hit in cls.many_from_rec(rec):
```
nh13, line 1432: do you want to move the body of this loop into its own method.  I.e. would it be useful to have a `to_pysam(self): AlignedSegment` method?

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+            secondary.mate_is_reverse = rec.mate_is_reverse
+            secondary.set_tag("MC", rec.get_tag("MC") if rec.has_tag("MC") else None)
+            secondary.set_tag("MQ", rec.get_tag("MQ") if rec.has_tag("MQ") else None)
+            secondary.set_tag("ms", rec.get_tag("ms") if rec.has_tag("ms") else None)
+
+            # NB: set some optional but highly recommended SAM tags on the secondary alignment.
+            secondary.set_tag("AS", hit.alignment_score)
+            secondary.set_tag("NM", hit.edit_distance)
+            secondary.set_tag("RG", rec.get_tag("RG") if rec.has_tag("RG") else None)
+            secondary.set_tag("RX", rec.get_tag("RX") if rec.has_tag("RX") else None)
```
nh13, line 1479: I wish pysam allowed us to provide a default value (including `None`) to the `get_tag` method.  Perhaps an issue?

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+            secondary.set_tag("rh", True)
+
+            yield secondary
+
+    @classmethod
+    def add_to_template(cls, template: Template) -> Template:
+        """Rebuild a template by adding secondary alignments from all R1/R2 `XA` and `XB` tags."""
+        r1_secondaries = iter([]) if template.r1 is None else cls.many_sam_from_rec(template.r1)
+        r2_secondaries = iter([]) if template.r2 is None else cls.many_sam_from_rec(template.r2)
+        return Template.build(chain(template.all_recs(), r1_secondaries, r2_secondaries))
```
nh13, line 1491: nit: name the parameters to build?

nh13 (commented):
Agreed on renaming the `SupplementaryAlignment` class to `AlternativeAlignment`, then making an empty sub-class of `AlternativeAlignment` called `SupplementaryAlignment` and deprecating the latter.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+            errors.append(f"Edit distance cannot be <0! Found: {self.edit_distance}")
+        if self.alignment_score is not None and self.alignment_score < 0:
+            errors.append(f"Alignment score cannot be <0! Found: {self.alignment_score}")
+        if self.mapq is not None and self.mapq < 0:
+            errors.append(f"Mapping quality cannot be <0! Found: {self.mapq}")
+        if len(errors) > 0:
+            raise ValueError("\n".join(errors))
+
+    @classmethod
+    def from_tag_part(cls, part: str) -> Self:
```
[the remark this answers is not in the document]
clintval: I agree it should probably be private. It was [public in `SupplementaryAlignment`](https://github.com/fulcrumgenomics/fgpyo/blob/b0b4227de8acff8a47f48d2afb06190c1b9850ad/fgpyo/sam/__init__.py#L806-L819) so that's why it is public here.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+            reference_start=int(stranded_start[1:]) - 1,
+            is_forward=stranded_start[0] == "+",
+            cigar=Cigar.from_cigarstring(cigar),
+            edit_distance=int(edit_distance),
+            alignment_score=None if alignment_score is None else int(alignment_score),
+            mapq=None if mapq is None else int(mapq),
+        )
+
+    @classmethod
+    def many_from_tag(cls, value: str) -> list[Self]:
```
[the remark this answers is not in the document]
clintval: To me, `from_tag` reads like it will only build one instance of the object. But we are building many instances. We don't have it codified anywhere _yet_ but I really like the convention of using `from_` for class methods that build a single instance and `many_from_` for class methods that build a collection of the instance.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1263,234 @@ class SamOrder(enum.Enum):
...
+
+        Args:
+            rec: The SAM record to generate secondary alignments from.
+        """
+        secondaries: list[Self] = []
+        try:
+            if rec.has_tag("XA"):
+                secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XA"))))
+            if rec.has_tag("XB"):
+                secondaries.extend(cls.many_from_tag(cast(str, rec.get_tag("XB"))))
```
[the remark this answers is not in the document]
clintval: Do you mean like:

- `from_tag_string`
- `from_tag_int`
- etc. etc.

Or do you mean a generic function? Generic functions only got support in Python 3.12 so I doubt pysam will be quick to implement it. For example, this will soon be possible, but only if pysam drops old Python support.


```python
def get_tag[T](self, tag: str) -> T:
```

PR #208, opened 2024-12-31
DRAFT of AuxAlignment for parsing secondary/supplementary alignments

clintval:
This is a draft PR to explore a common hierarchy for "alignments from SAM tags".

Motivated by:

- https://github.com/fulcrumgenomics/fgpyo/pull/206#issuecomment-2565707447

The diff is ugly but there are only 3 classes. I suggest reviewing like:

- [`AuxAlignment`](https://github.com/fulcrumgenomics/fgpyo/blob/5277047d55987a1358dd8f6f1e03b3d4ce5d6425/fgpyo/sam/__init__.py#L1192-L1401): abstract parent with some common concrete shared functions
- [`SecondaryAlignment`](https://github.com/fulcrumgenomics/fgpyo/blob/5277047d55987a1358dd8f6f1e03b3d4ce5d6425/fgpyo/sam/__init__.py#L1404-L1510): functionality specific to secondary alignments
- [`SupplementaryAlignment`](https://github.com/fulcrumgenomics/fgpyo/blob/5277047d55987a1358dd8f6f1e03b3d4ce5d6425/fgpyo/sam/__init__.py#L1513-L1644): functionality specific to supplementary alignments

I was able to make almost everything backwards compatible except:

- `SupplementaryAlignment` is now a dataclass vs attrs
- `SupplementaryAlignment` had a few init fields renamed

@tfenne what do you think?

I'm looking for a quick 👍 👎 so I can abandon the idea, or merge it into my actual feature branch for additional polish and a real round of review.

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -153,17 +153,45 @@ and slices):
    >>> [str(sup.cigar) for sup in sups]
    ['50S100M', '75S75M']
 ```
+
+## Examples of parsing bwa's `XA` and `XB` tags into individual secondary alignments
+
+```python
+>>> from fgpyo.sam import SecondaryAlignment
+>>> xa = SecondaryAlignment.from_tag_item("chr9,-104599381,49M,4")
+>>> xa.reference_name
+'chr9'
+>>> xb = SecondaryAlignment.from_tag_item("chr9,-104599381,49M,4,0,30")
+>>> xb.reference_name
+'chr9'
+>>> xb.mapq
+30
+>>> xa.cigar == xb.cigar
+True
+>>> xb_tag = "chr9,-104599381,49M,4,0,30;chr3,+170653467,49M,4,0,20;"
+>>> xb1, xb2 = SecondaryAlignment.many_from_tag(xb_tag)
+>>> xb1.is_forward
+False
+>>> xb1.is_forward
+True
+>>> xb1.mapq, xb2.mapq
+('30', '20')
+```
+
 """
 
 import enum
 import io
 import sys
 from collections.abc import Collection
+from dataclasses import dataclass
+from functools import cached_property
 from itertools import chain
 from pathlib import Path
 from typing import IO
 from typing import Any
 from typing import Callable
+from typing import ClassVar
 from typing import Dict
 from typing import Iterable
 from typing import Iterator
@@ -178,10 +206,12 @@ import pysam
 from pysam import AlignedSegment
 from pysam import AlignmentFile as SamFile
 from pysam import AlignmentHeader as SamHeader
+from typing_extensions import Self
 from typing_extensions import deprecated
 
 import fgpyo.io
 from fgpyo.collections import PeekableIterator
+from fgpyo.sequence import reverse_complement
 
 SamPath = Union[IO[Any], Path, str]
 """The valid base classes for opening a SAM/BAM/CRAM file."""
@@ -194,6 +224,9 @@ NO_REF_NAME: str = "*"
 
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
+
+NO_QUERY_BASES: str = "*"
+"""The string to use for a SAM record with missing query bases."""
 
 _IOClasses = (io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase)
 """The classes that should be treated as file-like classes"""
@@ -765,6 +798,7 @@ def is_proper_pair(
     )
 
 
+@deprecated("SupplementaryAlignment is deprecated after fgpyo 0.8.0. Use AuxAlignment instead!")
 @attr.s(frozen=True, auto_attribs=True)
 class SupplementaryAlignment:
     """Stores a supplementary alignment record produced by BWA and stored in the SA SAM tag.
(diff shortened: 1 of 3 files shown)


fgpyo/sam/__init__.py
```diff
@@ -1266,8 +1190,220 @@ class SamOrder(enum.Enum):
 
 
 @dataclass(frozen=True)
-class SecondaryAlignment:
-    """A secondary alignment as encoded in one part of the `XA` or `XB` tag value on a SAM record.
+class AuxAlignment(ABC, DataclassInstance):
```
nh13, line 1193: Yeah, this is a better name than `AlternateAlignment` since it's shorter and wont be confused as an alignment to an _alternate_ contig.

fgpyo/sam/__init__.py
```diff
@@ -1266,8 +1190,220 @@ class SamOrder(enum.Enum):
...
+    """An alignment parsed from the data stored in the SAM tags of a SAM record.
```
nh13, line 1194: This is typically where it's parsed from, but really, it's just some minimum information about an alternative alignment.  Would you be ope to changing the description?

fgpyo/sam/__init__.py
```diff
@@ -1266,8 +1190,220 @@ class SamOrder(enum.Enum):
...
+        alignment_score: The aligner-reported alignment score, if available.
+
+    Raises:
+        ValueError: If reference_start is set to a value less than zero.
+        ValueError: If mapq is set to a value less than zero.
+        ValueError: If edit_distance is set to a value less than zero.
+        ValueError: If alignment_score is set to a value less than zero.
+    """
+
+    sam_tags: ClassVar[Collection[str]]
```
nh13, line 1212: class constants should be upper snake case ([link](https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations))
```suggestion
    SAM_TAGS: ClassVar[Collection[str]]
```

fgpyo/sam/__init__.py
```diff
@@ -1266,8 +1190,220 @@ class SamOrder(enum.Enum):
...
+        """Perform post-initialization validation on this dataclass."""
+        errors: list[str] = []
+        if self.reference_start < 0:
+            errors.append(f"Reference start cannot be less than 0! Found: {self.reference_start}")
+        if self.mapq is not None and self.mapq < 0:
+            errors.append(f"Mapping quality cannot be less than 0! Found: {self.mapq}")
+        if self.edit_distance < 0:
+            errors.append(f"Edit distance cannot be less than 0! Found: {self.edit_distance}")
+        if self.alignment_score is not None and self.alignment_score < 0:
+            errors.append(f"Alignment score cannot be less than 0! Found: {self.alignment_score}")
```
nh13, line 1233: why not?  I think a negative alignment score is aligned, especially in a global alignment.

nh13 (commented):
I did not duplicate my comments from #206, so please incorporate those too as appropriate.

Instead of two sub-classes of `AuxAlignment`, what about storing an optional enum for if the alignment is secondary or supplementary (or `None` if neither or unknown)?  Then for the `from_tag(value: str)` method, you could condition on if it's 4 values (XA), or 6 values with strand as the third item (SA), or 6 values with the cigar as the third item (XB).  Alternatively, could you explain the motivation for the three classes beyond how they parse differently (which I think is resolved above)?  Do we want to match on the class/type?

fgpyo/sam/__init__.py
clintval: I'd like to learn more about this concern! I haven't seen a score below zero for any output from `bwa` so I figured it would be a sign that data was corrupt. Couple questions:

1. Is there a way to run `bwa` in which the score can go below zero?
2. Knowing that this code will probably parse `bwa` tags for the forseeable future, do you think keeping the strict check is better for safety-sake or would hinder someone in the future with a popular or custom global aligner putting auxiliary alignments in SAM tags?

clintval:
Comments from Nils in the other PR:

- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#pullrequestreview-2541497602
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909778418
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909779254
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909779438
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909780019
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909781386
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909783463
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909785485
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/206#discussion_r1909785891

fgpyo/sam/__init__.py
clintval: I see what you mean, but the implementation is heavily hardcoded around the idea of building auxiliary alignments from SAM tags. For example, the fields are a rough subset of what you have for a pysam alignment, but only those that exist in SAM tags. I'm leaving the description as-is for now and we can revisit once I finish a refactor to see if rephrasing it makes more sense to me.

clintval:
Closing this and will open a clean PR with a refactor.

PR #209, opened 2025-01-10
Add AuxAlignment for parsing secondary/supplementary alignments in SAM tags

clintval:


fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1250,285 @@ class SamOrder(enum.Enum):
...
+                alignment_score=None if alignment_score is None else int(alignment_score),
+                is_secondary=True,
+                is_supplementary=False,
+            )
+
+        else:
+            raise ValueError(f"{tag} tag value has the incorrect number of fields: {value}")
+
+    @classmethod
+    def many_from_primary(cls, primary: AlignedSegment) -> list[Self]:
```
clintval, line 1417: I know we don't have it codified anywhere but I really like using `from_` for class methods that build 1 instance of the class, and `many_from_` for class methods that build multiple instances in a collection/iterable. I think it makes code more readable.

tfenne: Though I don't think it should be `from_primary` ... one could also call this on a supplementary record to inflate the `SA` tag if one so desired.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1250,285 @@ class SamOrder(enum.Enum):
...
+        """Build all auxiliary alignments for a given primary alignment.
+
+        Args:
+            primary: The primary alignment to build auxiliary alignments from.
+
+        Raises:
+            ValueError: If the input record is a secondary or supplementary alignment.
+        """
+        if primary.is_secondary or primary.is_supplementary:
+            raise ValueError("Cannot build auxiliary alignments from a non-primary alignment!")
```
tfenne, line 1427: But you can!

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1250,285 @@ class SamOrder(enum.Enum):
...
+            aux.set_tag("rh", True)
+
+            yield aux
+
+    @classmethod
+    def add_all_to_template(cls, template: Template) -> Template:
+        """Rebuild a template by adding auxiliary alignments from the primary alignment tags."""
+        r1_aux = iter([]) if template.r1 is None else cls.many_pysam_from_primary(template.r1)
+        r2_aux = iter([]) if template.r2 is None else cls.many_pysam_from_primary(template.r2)
+        return Template.build(recs=chain(template.all_recs(), r1_aux, r2_aux))
```
tfenne, line 1534: I think this would feel more at home on `Template`, and it's quite simple, so I don't see any harm adding e.g.:

```python
def with_all_aux_alignments() -> Template:
```

to Template.

tfenne (commented):
I like this direction!

nh13 (approved):
LGTM after @tfenne's comments look good and you're happy with it

fgpyo/sam/__init__.py
msto: Just to add to this, after Nils' comments in slack yesterday, I've found it helpful to disambiguate the use of "primary" with the following distinctions:

- **"primary"** refers to _"non-secondary"_, meaning the best (or arbitrarily selected) among multiple alignments .
- **"representative"** refers to _"non-supplementary"_, meaning the alignment record chosen to represent a chimeric alignment

msto: I can't speak to how secondary alignments are typically represented, but I believe that supplementary alignments do typically include the `SA` tag with aux-representations of the representative and any other supplementary alignments

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1256,268 @@ class SamOrder(enum.Enum):
...
+        errors: list[str] = []
+        if self.reference_start < 0:
+            errors.append(f"Reference start cannot be less than 0! Found: {self.reference_start}")
+        if self.mapping_quality is not None and self.mapping_quality < 0:
+            errors.append(f"Mapping quality cannot be less than 0! Found: {self.mapping_quality}")
+        if self.edit_distance is not None and self.edit_distance < 0:
+            errors.append(f"Edit distance cannot be less than 0! Found: {self.edit_distance}")
+        # TODO: Some aligners might allow for a score <0 but I'm not sure bwa does... Keep this?
+        # if self.alignment_score is not None and self.alignment_score < 0:
+        #    errors.append(f"Alignment score cannot be less than 0! Found: {self.alignment_score}")
```
clintval, line 1350: @tfenne @nh13 there is a concern about keeping this validation.

Do you know what the right thing is to do?

Source: https://github.com/fulcrumgenomics/fgpyo/pull/208#discussion_r1910434919

nh13: See the `bwa aln -T` parameter, which I have set negative to get more alignments.  I'd say remove it.

fgpyo/sam/__init__.py
```diff
@@ -1231,3 +1256,268 @@ class SamOrder(enum.Enum):
...
+            or rec.query_qualities is None
+        ):
+            return
+
+        for hit in cls.many_from_rec(rec):
+            # TODO: When the original record has hard clips we must set the bases and quals to "*".
+            #       It would be smarter to pad/clip the sequence to be compatible with new cigar...
+            if "H" in rec.cigarstring or rec.query_sequence == NO_QUERY_BASES:
+                query_sequence = NO_QUERY_BASES
+                query_qualities = None
```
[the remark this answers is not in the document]
clintval: Ugh. It's too hard to solve right now... no thank you!

fgpyo/sam/__init__.py
```diff
@@ -1203,6 +1227,12 @@ def set_tag(
         for rec in self.all_recs():
             rec.set_tag(tag, value)
 
+    def with_aux_alignments(self) -> "Template":
```
msto, line 1230:
```suggestion
    def with_aux_alignments(self, include_secondary: bool = True, include_supplementary: bool = False) -> "Template":
```

Are these sensible defaults?

By default, `bwa aln` will produce a _single_ record with all **secondary** alignments in the XA tag, and `bwa mem` will produce _multiple_ records, each with all other supplementary alignments in the SA tag

Alternatively (or in addition), we could confirm that rehydrating does not create any duplicate records on the template

clintval: Yes! This refactor may actually help me with another design decision I didn't love but couldn't solve easily.

- https://github.com/fulcrumgenomics/fgpyo/pull/209#discussion_r1910662934

PR #212, opened 2025-01-13
fix: `sum_of_base_qualities()` returns zero when a record has no base qualities

### LINKED ISSUE #210: `sum_of_base_qualities` raises `TypeError` if the received record has no base qualities
## Summary

`sum_of_base_qualities()` accesses the input record's `query_qualities` without first checking if the record has base qualities.

When the record does not have base qualities, this attribute is `None`, and the attempted comprehension fails with a `TypeError: 'NoneType' object is not iterable`.

https://github.com/fulcrumgenomics/fgpyo/blob/b0b4227de8acff8a47f48d2afb06190c1b9850ad/fgpyo/sam/__init__.py#L849-L863

## Suggested solution

After empirically testing the results of `samtools fixmate -m` on the real read pair that originated this error, I found that **samtools will assign a mate score of 0 when the read has no base qualities.** (h/t @clintval for the suggestion)

I think it's therefore most appropriate if the function returns 0 in that case. 

---

## Original suggestion

~Either of the following could be appropriate:~

~1. Changing the signature of the function to return `int | None`, and returning `None` when the input record has no base qualities~
~2. Raising `ValueError` with a more specific error message when the input record has no base qualities~

msto:
Closes #210

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -158,6 +158,7 @@ and slices):
 import enum
 import io
 import sys
+from array import array
 from collections.abc import Collection
 from itertools import chain
 from pathlib import Path
@@ -178,6 +179,7 @@ import pysam
 from pysam import AlignedSegment
 from pysam import AlignmentFile as SamFile
 from pysam import AlignmentHeader as SamHeader
+from pysam import qualitystring_to_array
 from typing_extensions import deprecated
 
 import fgpyo.io
@@ -194,6 +196,12 @@ NO_REF_NAME: str = "*"
 
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
+
+STRING_PLACEHOLDER: str = "*"
+"""The value to use when a string field's information is unavailable."""
+
+NO_QUERY_QUALITIES: array = qualitystring_to_array(STRING_PLACEHOLDER)
+"""The quality array corresponding to an unavailable query quality string ("*")."""
 
 _IOClasses = (io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase)
 """The classes that should be treated as file-like classes"""
@@ -864,7 +872,7 @@ def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> i
         [`calc_sum_of_base_qualities()`](https://github.com/samtools/samtools/blob/4f3a7397a1f841020074c0048c503a01a52d5fa2/bam_mate.c#L227-L238)
         [`MD_MIN_QUALITY`](https://github.com/samtools/samtools/blob/4f3a7397a1f841020074c0048c503a01a52d5fa2/bam_mate.c#L42)
     """
-    if rec.query_qualities is None:
+    if rec.query_qualities is None or rec.query_qualities == NO_QUERY_QUALITIES:
         return 0
 
     score: int = sum(qual for qual in rec.query_qualities if qual >= min_quality_score)
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -849,16 +849,24 @@ class SupplementaryAlignment:
 def sum_of_base_qualities(rec: AlignedSegment, min_quality_score: int = 15) -> int:
     """Calculate the sum of base qualities score for an alignment record.
 
-    This function is useful for calculating the "mate score" as implemented in samtools fixmate.
+    This function is useful for calculating the "mate score" as implemented in `samtools fixmate`.
+    Consistently with `samtools fixmate`, this function returns 0 if the record has no base
+    qualities.
 
     Args:
         rec: The alignment record to calculate the sum of base qualities from.
         min_quality_score: The minimum base quality score to use for summation.
 
+    Returns:
+        The sum of base qualities on the input record. 0 if the record has no base qualities.
+
     See:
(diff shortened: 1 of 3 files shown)


fgpyo/util/inspect.py
```diff
@@ -457,7 +457,7 @@ def attr_from(
             # True, because python, so we need to check for that explicitly
             if not set_value and attribute.type is not None and not attribute.type == bool:
                 try:
-                    return_value = attribute.type(str_value)  # type: ignore[operator]
+                    return_value = attribute.type(str_value)
```
msto, line 460: NB: this was causing a mypy error on my local (unused type ignore)

fgpyo/sam/__init__.py
```diff
@@ -849,16 +849,24 @@ def from_read(cls, read: pysam.AlignedSegment) -> List["SupplementaryAlignment"]
...
 
+    Returns:
+        The sum of base qualities on the input record. 0 if the record has no base qualities.
+
     See:
         [`calc_sum_of_base_qualities()`](https://github.com/samtools/samtools/blob/4f3a7397a1f841020074c0048c503a01a52d5fa2/bam_mate.c#L227-L238)
         [`MD_MIN_QUALITY`](https://github.com/samtools/samtools/blob/4f3a7397a1f841020074c0048c503a01a52d5fa2/bam_mate.c#L42)
     """
+    if rec.query_qualities is None:
+        return 0
```
clintval, line 868: For safety-sake, such as if someone manually created a SAM record with the `*`, I'd probably still want an extra guard here:

```suggestion
    if rec.query_qualities is None or rec.query_qualities == NO_QUERY_QUALS:
        return 0
```

msto: Additionally, after assigning to `record.query_sequence`, the `query_qualities` is automatically set to `None`

https://github.com/pysam-developers/pysam/blob/0eae5be21ac3ab3ac7aa770a3931e2977e37b909/pysam/libcalignedsegment.pyx#L1461-L1463

```py
In [10]: record.query_qualities

In [11]: record = builder.add_single()

In [12]: record.query_sequence = "*"

In [13]: record.query_qualities is None
Out[13]: True
```

nh13: It's deprecated, but you can use `single.qual = None` or `single.qual = ""` or `single.qual = "*"`.

nh13: https://github.com/pysam-developers/pysam/pull/1324

msto: Ah, I see. 

Then I think the correct thing to do is to check if the record's `query_qualities` is equal to `qualitystring_to_array(NO_QUERY_QUALS)`, not the string itself.

fgpyo/sam/__init__.py
```diff
@@ -195,6 +197,12 @@
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
 
+STRING_PLACEHOLDER: str = "*"
+"""The value to use when a string field's information is unavailable."""
+
+NO_QUERY_QUALITIES: array = qualitystring_to_array(STRING_PLACEHOLDER)
+"""The quality array corresponding to an unavailable query quality string ("*")."""
+
```
msto, line 205: @clintval @nh13 Are you okay with the addition of these module-level constants?

I thought it was more appropriate to include a generic `STRING_PLACEHOLDER` constant rather than a more specific "NO_QUERY_QUALITY" constant because the SAM spec indicates that `*` is the generic placeholder for unavailable string fields. 

> if the information represented by any of these fields is unavailable, that field’s value will be a
placeholder, either ‘0’ or ‘*’ as determined by the field’s type. 

(Technically, I think `NO_REF_NAME` should have been this originally, but I wanted to leave that in for backwards compatibility.)

And I included the separate `NO_QUERY_QUALITIES` constant because I thought it was more efficient to convert this string to an array once, rather than every time `sum_of_base_qualities()` is called.

fgpyo/sam/__init__.py
```diff
@@ -195,6 +197,12 @@
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
 
+STRING_PLACEHOLDER: str = "*"
+"""The value to use when a string field's information is unavailable."""
+
+NO_QUERY_QUALITIES: array = qualitystring_to_array(STRING_PLACEHOLDER)
```
msto, line 203: Please note that `array` is not a generic collection type and so we cannot make this annotation more specific.

fgpyo/sam/__init__.py
msto: P.S. @clintval your original suggestion makes more sense after I found `NO_QUERY_QUALS` defined in your unmerged PR 😅 

https://github.com/fulcrumgenomics/fgpyo/blob/e271e6ca9a13c6420ad2e5acdb6e321ed97107cb/fgpyo/sam/__init__.py#L218-L219

clintval (approved):

fgpyo/sam/__init__.py
```diff
@@ -195,6 +197,12 @@
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
 
+STRING_PLACEHOLDER: str = "*"
+"""The value to use when a string field's information is unavailable."""
```
clintval, line 201: If you place this above line 194, you can use it on the right-hand-side of `NO_REF_NAME`. I don't think we need this variable, but I also don't mind having it.

PR #214, opened 2025-01-17
Add tabix indexing of gzipped VCFs in VariantBuilder

tfenne:


fgpyo/vcf/builder.py
```diff
@@ -279,6 +280,9 @@ def to_path(self, path: Optional[Path] = None) -> Path:
             for variant in self.to_sorted_list():
                 writer.write(variant)
 
+        if str(path.suffix) == ".gz":
+            pysam.tabix_index(str(path), preset="vcf")
```
nh13, line 284: Nit because this is on the builder class and we control the file :P

Will this fail if the file is already indexed (idk)?  If it isn't block-compressed (probably)?

Can you add to the docstring that indexing will occur if the VCF is gz?

nh13 (approved):
I had to find one comment right?  Thank-you for adding this!

fgpyo/vcf/builder.py
tfenne: I added to the docstring.

We only partially control the path - a user can request it be written to anywhere, and may specifically ask for a non-gzipped path, so I think the test has to remain.

I added `force=True` which will cause it to override any existing index (e.g. if the user repeatedly write different VCFs to the same path).  It will fail if the file isn't block-compressed ... but samtools' VCF writing writes bgzip for `.gz` extensions, so yay?

PR #220, opened 2025-02-10
feat: allow lineterminator specification in MetricWriter

znorgaard:
I stumbled upon the fact that `csv.DictWriter` uses windows  new line endings by default. This PR lets you change that when using `MetricWriter`.

I'm arguing for the default to be `"\n"` instead of `os.linesep` because:
- [Python has universal new line support](https://peps.python.org/pep-0278/) (so it doesn't care what we do)
- If you're working on a shared analysis and checking in files to version control, you don't want to worry about if you're running on the same OS as your collaborator
- In my experience, more bioinformatics tools have problems with windows line endings than with unix line endings

### DIFF
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -450,7 +450,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
         delimiter: str = "\t",
         include_fields: Optional[List[str]] = None,
         exclude_fields: Optional[List[str]] = None,
-        **kwargs: Any,
+        lineterminator: str = "\n",
     ) -> None:
         """
         Args:
@@ -465,7 +465,8 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
             exclude_fields: If specified, any listed fieldnames will be excluded when writing
                 records to file.
                 May not be used together with `include_fields`.
-            kwargs: any keyword arguments to be passed to csv.DictWriter.
+            lineterminator: The string used to terminate lines produced by the MetricWriter.
+                Default = "\n".
 
         Raises:
             TypeError: If the provided metric class is not a dataclass- or attr-decorated
@@ -508,7 +509,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
             f=self._fout,
             fieldnames=self._fieldnames,
             delimiter=delimiter,
-            **kwargs,
+            lineterminator=lineterminator,
         )
 
         # If we aren't appending to an existing file, write the header before any rows
--- a/fgpyo/util/metric.py
+++ b/fgpyo/util/metric.py
@@ -450,6 +450,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
         delimiter: str = "\t",
         include_fields: Optional[List[str]] = None,
         exclude_fields: Optional[List[str]] = None,
+        **kwargs: Any,
     ) -> None:
         """
         Args:
@@ -464,6 +465,7 @@ class MetricWriter(Generic[MetricType], AbstractContextManager):
             exclude_fields: If specified, any listed fieldnames will be excluded when writing
                 records to file.
                 May not be used together with `include_fields`.
+            kwargs: any keyword arguments to be passed to csv.DictWriter.
 
         Raises:
             TypeError: If the provided metric class is not a dataclass- or attr-decorated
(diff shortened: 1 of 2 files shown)


fgpyo/util/metric.py
```diff
@@ -464,6 +465,7 @@ def __init__(
             exclude_fields: If specified, any listed fieldnames will be excluded when writing
                 records to file.
                 May not be used together with `include_fields`.
+            kwargs: any keyword arguments to be passed to csv.DictWriter.
```
nh13, line 468: Since we already bass in `f`, `fieldnames`, and `delimiter`.  Should we (a) document that those should not be given, and (b) throw an exception if they're present in `kwargs`?

tests/fgpyo/util/test_metric.py
```diff
@@ -613,11 +613,30 @@ def test_writer_from_str(tmp_path: Path) -> None:
 
     with fpath.open("r") as f:
         assert next(f) == "foo\tbar\n"
+        assert repr(f.newlines) == repr("\r\n")
```
nh13, line 616: nit: I see that you're using this test to also show that the _default_ newlines is `"\r\n"`, but this test is not for that purpose, and so this is a bit of a footgun if someone comes along in the future.  We should at least add a comment, since this would look out of place.  Better, would be to move this into `test_writer_with_kwargs` and parameterize the `lineterminator` in the test.

tests/fgpyo/util/test_metric.py
```diff
@@ -613,11 +613,30 @@ def test_writer_from_str(tmp_path: Path) -> None:
...
         assert next(f) == "abc\t1\n"
         with pytest.raises(StopIteration):
             next(f)
 
 
+def test_writer_with_kwargs(tmp_path: Path) -> None:
```
nh13, line 622: Add this and fix up the test:
```suggestion
@pytest.mark.parameterize("lineterminator", ["\n", "\r", "\r\n"])
def test_writer_with_kwargs(tmp_path: Path, lineterminator: str) -> None:
```

nh13 (changes requested):
Thank-you for adding this! I have no objections to this in principle, so a few minor suggestions, and then I can re-review.

tfenne:
I'm going to say that I don't love this - it exposes, at the API level, that we're using `csv.dictwriter` internally.  What if we change that?  How would you feel about instead a) a PR to allow passing in the preferred newline as a specific argument, and b) having it be `str | None` and if `None` we do a better job of defaulting it based on whether we're on windows or not-windows?

fgpyo/util/metric.py
```diff
@@ -506,6 +508,7 @@ def __init__(
             f=self._fout,
             fieldnames=self._fieldnames,
             delimiter=delimiter,
+            **kwargs,
```
msto, line 511: We should probably change the default behavior of this class to override `DictWriter` and use `"\n"` as the default lineterminator.

Maybe check to see if `lineterminator` is in the user-specified `kwargs` and assign `"\n"` if not

znorgaard:
I'm seeing a couple different perspectives on this one.

### To expose kwargs or not to expose kwargs
If we only expose a "lineterminator" arg at the API level, there's a chance we'll end up needing to expose other arguments supported by DictWriter. This is potentially more developer work in the long term, but keeps the implementation relatively easy to follow and allows us to swap out the writer without changing the API.

Exposing kwargs opens a can of worms around "what if someone used API level `delimiter` or `fieldnames` arg and passed in a dictionary for kwargs with those values defined again?" We probably want to add some checking/filtering code to make sure arguments are not doubled and document which one takes precedence. This could lead to a breaking API change in the future if we change the underlying writer, but we would immediately support every DictWriter arg someone might want to use.

Personally, I'm leaning towards just exposing lineterminator because I don't see many other options we'd want to expose at the API level, but that could just be personal perspective.

### What should the default newline character be?

Arguments for "\n" and platform aware. Python readers and writers are generally smart enough to auto-detect file endings without much trouble regardless of the current OS. Other software (nextflow) is not and assumes "\n". Excel, likely to be used with metric files is smart enough to use "\n" or "\n\r" line endings.

I think I'd prefer to default to "\n" because that's what will "just work" with most software we're likely to pass these files to.

### tldr

I vote we only expose `lineterminator` at the API level and default to "\n". Major objections?

nh13 (approved):

PR #221, opened 2025-02-12
chore: Move numpy to docs dependency group

msto:
## Summary

I'm running into a problem where after bumping a package from `fgpyo=0.3.0` to `fgpyo=0.8.0` , poetry downgrades `numpy` from `2.x` to `1.x` and mypy throws a bunch of errors as a consequence.

AFAICT the numpy dependency and version pin was introduced with [the change to mkdocs](https://github.com/fulcrumgenomics/fgpyo/pull/135), but we don't use numpy anywhere in the fgpyo codebase. 

I suggest moving numpy to the docs dependency group instead, so it is not required to install `fgpyo`.

## Changes
- Moved `numpy` to the `docs` dependency group, so it is no longer required to install `fgpyo`.
- Made changes to type hints and documentation as a consequence of `mypy` and `mkdocs` being updated in the new lock file.


nh13:
And the real offender is me when I didn't move `numpy` along with this change: https://github.com/fulcrumgenomics/fgpyo/pull/166

msto:
It's an odd dependency to have for documentation! I reviewed and approved that PR and didn't catch it either.

msto:
Outside of the `pyproject.toml` and lock file updates, the only changes are to type hints and docstrings, so I would be comfortable merging without codecov's approval.

nh13 (approved):
LGTM

PR #225, opened 2025-03-11
ci: use github config for git-cliff

nh13:


nh13:
Whatever we do here, we probably want to do for:
- https://github.com/fulcrumgenomics/prymer/pull/127
- https://github.com/fulcrumgenomics/pybwa/pull/59
- https://github.com/fulcrumgenomics/python-template
and others.

My motivation was to include PR links and attributions, which the `--config github` provides, but @clintval also notes we lose the order by conventional commits, which is an unintended and unwanted side effect, so I am making this a draft until we can have both.

PR #229, opened 2025-03-25
feat: calculate MD and NM tags

### LINKED ISSUE #84: Add a method to calculate MD and NM tags.
This is useful for when we perform clipping or other modifications to a SAM record that would affect these values.

For a code example, see [htsjdk's `calculateMdAndNmTags`](https://github.com/samtools/htsjdk/blob/b16416c078e2df3e3a21486207d0c0c31b293cb9/src/main/java/htsjdk/samtools/util/SequenceUtil.java#L964).  I think an additional parameter to make the output compatible with samtools makes sense (i.e. _do not treat N->N_ as a match).  See how this is used in fgbio both [here](https://github.com/fulcrumgenomics/fgbio/blob/9c475fdb06e4f05a2a5eabe9006e0facb31726b5/src/main/scala/com/fulcrumgenomics/bam/Bams.scala#L551) and [here](https://github.com/fulcrumgenomics/fgbio/blob/9c475fdb06e4f05a2a5eabe9006e0facb31726b5/src/main/scala/com/fulcrumgenomics/bam/Bams.scala#L468), which handle unmapped reads.

Bonus points for adding another method that takes a iterator over `Template`s (and/or `pysam.AlignedSegment`s) and a `pysam.FastaFile` and updates the records.

emmcauley:
Closes #84. Adds an `md` attribute to the `ReadEditInfo` class and an `n_as_match` boolean flag to `calculate_edit_info()`.  `calculate_edit_info()` is now a little more complex than we might like (11 > 10). I tried not to change any pre-existing code more than was necessary.

Some of the reads for testing were inherited from `htsjdk`.

⚠️ Once we converge on the style and the functionality in this PR, I can open another PR to address this part of the Github issue: `Bonus points for adding another method that takes a iterator over Templates (and/or pysam.AlignedSegments) and a pysam.FastaFile and updates the records.`

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -1048,7 +1048,7 @@ def calculate_edit_info(  # noqa: C901 (11 > 10)
     reference_sequence: str,
     n_as_match: bool = True,
     reference_offset: Optional[int] = None,
-) -> ReadEditInfo | None:
+) -> Optional[ReadEditInfo]:
     """
     Constructs a `ReadEditInfo` instance giving summary stats about how the read aligns to the
     reference.  Computes the number of mismatches, indels, indel bases as well as the
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -1046,31 +1046,42 @@ class ReadEditInfo:
 def calculate_edit_info(  # noqa: C901 (11 > 10)
     rec: AlignedSegment,
     reference_sequence: str,
-    n_as_match: bool,
+    n_as_match: bool = True,
     reference_offset: Optional[int] = None,
-) -> ReadEditInfo:
+) -> ReadEditInfo | None:
     """
     Constructs a `ReadEditInfo` instance giving summary stats about how the read aligns to the
     reference.  Computes the number of mismatches, indels, indel bases as well as the
     SAM NM and MD tags.
 
-    The read must be aligned.
+    Calculation of NM and MD tags is based off of htsjdk:
+    https://github.com/samtools/htsjdk/blob/7034b33636b4cb9fec300a2136588e7c12c7ccd5/src/main/java/htsjdk/samtools/util/SequenceUtil.java#L964:L1029
+
+    Per the SAM specification (https://samtools.github.io/hts-specs/SAMtags.pdf), the NM tag
+    encapsulates the number of differences between the query read and reference sequence, counting
+    only A, C, G and T bases (case-insensitive). Everything else should be considered a mismatch
+    (e.g., ambiguity codes like R and N). We set the default of `n_as_match` to True to be
+    concordant with `htsjdk`, which treats an N->N as a match.
+
+    If the read is unmapped or the query sequence contains missing bases (`*`), returns None, as it
+    is not possible to recalculate the MD and NM tags without access to the query sequence and
+    reference sequence.
 
     Args:
         rec: the read/record for which to calculate values
         reference_sequence: the reference sequence (or fragment thereof) to which the read is
             aligned
         n_as_match: if True, mirror htsjdk `calculateMdAndNmTags` and treat N->N as a match. If
-            False, N->N is treated as a mismatch.
+            False, N->N is treated as a mismatch. (default: True)
         reference_offset: if provided, assume that reference_sequence[reference_offset] is the
             first base aligned to in reference_sequence, otherwise use r.reference_start
 
     Returns:
         a ReadEditInfo with information about how the read differs from the reference
     """
-    assert (
-        not rec.is_unmapped and rec.query_sequence != NO_QUERY_BASES
-    ), f"Cannot calculate edit info for unmapped and/or empty read: {rec}"
+    if rec.is_unmapped or rec.query_sequence is None or rec.query_sequence == NO_QUERY_BASES:
+        return None
+
     query_offset: int = 0
     target_offset: int = reference_offset if reference_offset is not None else rec.reference_start
     cigar: Cigar = Cigar.from_cigartuples(rec.cigartuples)
@@ -1084,7 +1095,7 @@ def calculate_edit_info(  # noqa: C901 (11 > 10)
         # TODO: use match-case statements after we drop Python 3.9 support
         if op == CigarOp.N:  # skipped region from ref, consumes ref
             target_offset += elem.length
-        elif op in [CigarOp.I, CigarOp.S]:  # consumes query
+        elif op in (CigarOp.I, CigarOp.S):  # consumes query
             query_offset += elem.length
             if op == CigarOp.I:
                 insertions += 1
@@ -1098,7 +1109,7 @@ def calculate_edit_info(  # noqa: C901 (11 > 10)
             current_match_count = 0
             md_edits.append(f"^{reference_sequence[target_offset:target_offset + elem.length]}")
             target_offset += elem.length
-        elif op in [CigarOp.M, CigarOp.X, CigarOp.EQ]:
+        elif op in (CigarOp.M, CigarOp.X, CigarOp.EQ):
             for in_block_offset in range(0, elem.length):
                 if (target_offset + in_block_offset) >= len(reference_sequence):
                     break  # out of bounds
@@ -1113,9 +1124,10 @@ def calculate_edit_info(  # noqa: C901 (11 > 10)
                 else:  # match
                     matches += 1
                     current_match_count += 1
-
             query_offset += elem.length_on_query
             target_offset += elem.length_on_target
+        elif op not in (CigarOp.H, CigarOp.P):  # if we come across an unexpected CIGAR operation
+            raise ValueError(f"Invalid CIGAR operation: {op}")
 
     md_edits.append(str(current_match_count))
 
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -201,6 +201,9 @@ NO_REF_NAME: str = STRING_PLACEHOLDER
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
 
+NO_QUERY_BASES: str = "*"
+"""The string to use for a SAM record with missing query bases."""
+
 NO_QUERY_QUALITIES: array = qualitystring_to_array(STRING_PLACEHOLDER)
 """The quality array corresponding to an unavailable query quality string ("*")."""
 
@@ -1027,6 +1030,7 @@ class ReadEditInfo:
             the reference but not in the read).
         nm: the computed value of the SAM NM tag, calculated as mismatches + inserted_bases +
             deleted_bases
+        md: the computed value of the SAM MD tag
     """
 
     matches: int
@@ -1036,55 +1040,84 @@ class ReadEditInfo:
     deletions: int
     deleted_bases: int
     nm: int
+    md: str
 
 
-def calculate_edit_info(
-    rec: AlignedSegment, reference_sequence: str, reference_offset: Optional[int] = None
+def calculate_edit_info(  # noqa: C901 (11 > 10)
+    rec: AlignedSegment,
+    reference_sequence: str,
+    n_as_match: bool,
+    reference_offset: Optional[int] = None,
 ) -> ReadEditInfo:
     """
     Constructs a `ReadEditInfo` instance giving summary stats about how the read aligns to the
-    reference.  Computes the number of mismatches, indels, indel bases and the SAM NM tag.
+    reference.  Computes the number of mismatches, indels, indel bases as well as the
+    SAM NM and MD tags.
+
     The read must be aligned.
 
     Args:
         rec: the read/record for which to calculate values
-        reference_sequence: the reference sequence (or fragment thereof) that the read is
-            aligned to
+        reference_sequence: the reference sequence (or fragment thereof) to which the read is
+            aligned
+        n_as_match: if True, mirror htsjdk `calculateMdAndNmTags` and treat N->N as a match. If
+            False, N->N is treated as a mismatch.
         reference_offset: if provided, assume that reference_sequence[reference_offset] is the
             first base aligned to in reference_sequence, otherwise use r.reference_start
 
     Returns:
         a ReadEditInfo with information about how the read differs from the reference
     """
-    assert not rec.is_unmapped, f"Cannot calculate edit info for unmapped read: {rec}"
-
-    query_offset = 0
-    target_offset = reference_offset if reference_offset is not None else rec.reference_start
-    cigar = Cigar.from_cigartuples(rec.cigartuples)
+    assert (
+        not rec.is_unmapped and rec.query_sequence != NO_QUERY_BASES
+    ), f"Cannot calculate edit info for unmapped and/or empty read: {rec}"
+    query_offset: int = 0
+    target_offset: int = reference_offset if reference_offset is not None else rec.reference_start
+    cigar: Cigar = Cigar.from_cigartuples(rec.cigartuples)
 
     matches, mms, insertions, ins_bases, deletions, del_bases = 0, 0, 0, 0, 0, 0
-    ok_bases = {"A", "C", "G", "T"}
-
+    ok_bases: set[str] = {"A", "C", "G", "T"}
+    md_edits: list[str] = []
+    current_match_count: int = 0
     for elem in cigar.elements:
         op = elem.operator
-
-        if op == CigarOp.I:
-            insertions += 1
-            ins_bases += elem.length
-        elif op == CigarOp.D:
+        # TODO: use match-case statements after we drop Python 3.9 support
+        if op == CigarOp.N:  # skipped region from ref, consumes ref
+            target_offset += elem.length
+        elif op in [CigarOp.I, CigarOp.S]:  # consumes query
+            query_offset += elem.length
+            if op == CigarOp.I:
+                insertions += 1
+                ins_bases += elem.length
+        elif op == CigarOp.D:  # consumes ref
+            if target_offset + elem.length >= len(reference_sequence):
+                break  # out of bounds
             deletions += 1
             del_bases += elem.length
-        elif op == CigarOp.M or op == CigarOp.X or op == CigarOp.EQ:
-            for i in range(0, elem.length):
-                q = rec.query_sequence[query_offset + i].upper()
-                t = reference_sequence[target_offset + i].upper()
-                if q != t or q not in ok_bases:
-                    mms += 1
-                else:
-                    matches += 1
+            md_edits.append(str(current_match_count))  # append match count and reset
+            current_match_count = 0
+            md_edits.append(f"^{reference_sequence[target_offset:target_offset + elem.length]}")
+            target_offset += elem.length
+        elif op in [CigarOp.M, CigarOp.X, CigarOp.EQ]:
+            for in_block_offset in range(0, elem.length):
+                if (target_offset + in_block_offset) >= len(reference_sequence):
+                    break  # out of bounds
+                query_base: str = rec.query_sequence[query_offset + in_block_offset].upper()
+                ref_base: str = reference_sequence[target_offset + in_block_offset].upper()
 
-        query_offset += elem.length_on_query
-        target_offset += elem.length_on_target
+                if query_base != ref_base or (query_base not in ok_bases and not n_as_match):
+                    mms += 1
+                    md_edits.append(str(current_match_count))
+                    current_match_count = 0
+                    md_edits.append(ref_base)  # grab mismatched base from the reference
+                else:  # match
+                    matches += 1
+                    current_match_count += 1
+
+            query_offset += elem.length_on_query
+            target_offset += elem.length_on_target
+
+    md_edits.append(str(current_match_count))
 
     return ReadEditInfo(
         matches=matches,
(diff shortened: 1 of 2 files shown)


znorgaard (changes requested):

tests/fgpyo/sam/test_sam.py
```diff
@@ -651,18 +651,35 @@ def test_calc_edit_info_no_edits() -> None:
...
+    info = sam.calculate_edit_info(
+        rec=rec, reference_sequence=chrom[10:40], reference_offset=0, n_as_match=False
+    )
     assert info.mismatches == 0
     assert info.nm == 0
+    assert info.md == "30"
+
+
+def test_calc_edit_info_no_edits_with_nonzero_offset() -> None:
+    """Assert that a non-zero offset slides ref start position as expected."""
```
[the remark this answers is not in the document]
emmcauley: thank you for catching the oversight on the test name -- I've renamed as you suggested and included the ASCII art!

tests/fgpyo/sam/test_sam.py
```diff
@@ -688,30 +706,132 @@ def test_calc_edit_info_with_clipping_and_deletions() -> None:
...
+        chrom="chr2",
+        start=0,
+        cigar="16M",
+        attrs={"MD": "0T2A0T2A0t2a0t2a0", "NM": 8},
+    )
+
+    generated_result = sam.calculate_edit_info(
+        rec=rec, reference_sequence=chrom, n_as_match=n_as_match
+    )
+    expected_md: str = str(rec.get_tag("MD"))
```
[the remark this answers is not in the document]
emmcauley: makes sense, the file should converge on `info` now

tests/fgpyo/sam/test_sam.py
```diff
@@ -688,30 +714,122 @@ def test_calc_edit_info_with_clipping_and_deletions() -> None:
...
+    assert info.mismatches == 0
+    assert info.insertions == 0
+    assert info.inserted_bases == 0
+    assert info.deletions == 0
+    assert info.deleted_bases == 0
+    assert info.nm == 0
+    assert info.md == "6"
+
+
+def test_calc_edit_info_with_matches_out_of_bounds() -> None:
```
znorgaard, line 819: I only caught it because of codecov on the PR.
image

This test isn't covering what we think it is.

I did a quick confirmation by adding a `raise ValueError` before the break and it never triggered in the tests.

We currently have a query that completely overlaps the reference.

```
# Ref:   AGTCCGTTAGC
# Query: ------TTAG  6D4M
```

To get the matched bases to fall outside the range of the reference, we need to shorten the reference used in this test.

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
-    target_offset = reference_offset if reference_offset is not None else rec.reference_start
-    cigar = Cigar.from_cigartuples(rec.cigartuples)
+    query_offset: int = 0
+    target_offset: int = reference_offset if reference_offset is not None else rec.reference_start
+    cigar: Cigar = Cigar.from_cigartuples(rec.cigartuples)
 
     matches, mms, insertions, ins_bases, deletions, del_bases = 0, 0, 0, 0, 0, 0
-    ok_bases = {"A", "C", "G", "T"}
-
+    ok_bases: set[str] = {"A", "C", "G", "T"}
```
nh13, line 1090: issue:

To follow this sentence in the spec
> The special sequence base ‘=’ will always be considered to be a match, even if the reference is ambiguous at that point.

I think we need to add `=` to the ok bases so it will never be treated as a mismatch.

```suggestion
    ok_bases: set[str] = {"A", "C", "G", "T", "="}
```

We also need one guard added as a suggestion below.

Could you please add a unit test to test for `=` in the query bases?

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
+            md_edits.append(f"^{reference_sequence[target_offset:target_offset + elem.length]}")
+            target_offset += elem.length
+        elif op in (CigarOp.M, CigarOp.X, CigarOp.EQ):
+            for in_block_offset in range(0, elem.length):
+                if (target_offset + in_block_offset) >= len(reference_sequence):
+                    break  # out of bounds
+                query_base: str = rec.query_sequence[query_offset + in_block_offset].upper()
+                ref_base: str = reference_sequence[target_offset + in_block_offset].upper()
+
+                if query_base != ref_base or (query_base not in ok_bases and not n_as_match):
```
nh13, line 1119: issue:

To support that `=` is always a match, regardless of the reference.

```suggestion
                if (query_base != ref_base and query_base != "=") or (query_base not in ok_bases and not n_as_match):
```

tests/fgpyo/sam/test_sam.py
```diff
@@ -688,30 +715,136 @@ def test_calc_edit_info_with_clipping_and_deletions() -> None:
...
+
+@pytest.mark.parametrize("query_sequence", [None, "*"])
+def test_calc_edit_info_with_missing_query_bases(query_sequence: Optional[str]) -> None:
+    """Assert expected behavior for a read that has missing query bases."""
+    chrom = "AGTCCGTTA"
+    builder = SamBuilder(r1_len=10)
+    rec = builder.add_single(chrom="chr1", start=0)
+    rec.query_sequence = query_sequence  # set those bases to `None` or missing (`*`)
+    info = sam.calculate_edit_info(rec=rec, reference_sequence=chrom, n_as_match=False)
+    assert info is None
```
nh13, line 847: issue: can you please add a test cases for:
1. with `=` in the query bases?
2. with hard-clipping in the cigar (should not fail).
3. with padding in the cigar (should not fail)

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
+                    md_edits.append(str(current_match_count))
+                    current_match_count = 0
+                    md_edits.append(ref_base)  # grab mismatched base from the reference
+                else:  # match
                     matches += 1
+                    current_match_count += 1
+            query_offset += elem.length_on_query
+            target_offset += elem.length_on_target
+        elif op not in (CigarOp.H, CigarOp.P):  # if we come across an unexpected CIGAR operation
+            raise ValueError(f"Invalid CIGAR operation: {op}")
```
nh13, line 1130: Issue: I don't think these are errors.

Both hard clipping and padding operators consume neither reference bases nor read bases (see [the table on page 8 in the spec](https://samtools.github.io/hts-specs/SAMv1.pdf)), so they can be silently ignored.  This is what the `htsjdk`  implementation does, which we're following.

Can you add a test cases with cigars that have hard clips and padding operators (separately)?


```suggestion
```

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
+    current_match_count: int = 0
     for elem in cigar.elements:
         op = elem.operator
-
-        if op == CigarOp.I:
-            insertions += 1
-            ins_bases += elem.length
-        elif op == CigarOp.D:
+        # TODO: use match-case statements after we drop Python 3.9 support
+        if op == CigarOp.N:  # skipped region from ref, consumes ref
```
nh13, line 1096: issue: I think the order of the conditionals is important for performance.

We'll want to put the operators we expect to see most frequently first, as `htsjdk` does, so we have fewer conditional checks on average.  We rarely see an `N` in the cigar for example.  Could you please re-order the if/elif blocks to match `htsjdk`?

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
+            target_offset += elem.length
+        elif op in (CigarOp.I, CigarOp.S):  # consumes query
+            query_offset += elem.length
+            if op == CigarOp.I:
+                insertions += 1
+                ins_bases += elem.length
+        elif op == CigarOp.D:  # consumes ref
+            if target_offset + elem.length >= len(reference_sequence):
+                break  # out of bounds
```
nh13, line 1105: issue: this break occurs _earlier_ than it does in `htsjdk`, and that's important.

We need to add the current match count to the `md_edits` before we break.

nh13 (changes requested):

tests/fgpyo/sam/test_sam.py
```diff
@@ -688,30 +715,136 @@ def test_calc_edit_info_with_clipping_and_deletions() -> None:
...
+    assert info.nm == int(rec.get_tag("NM"))
+
+
+@pytest.mark.parametrize("n_as_match", [True, False])
+def test_calc_edit_info_with_soft_clip_at_end(n_as_match: bool) -> None:
+    """`htsjdk` testing data with soft clips and deletions. `n_as_match` should have no effect."""
+    chrom = "TCGATCGAtcgatcga"
+    builder = SamBuilder(r1_len=8)
+    rec = builder.add_single(
+        bases="TCGACGAA", chrom="chr2", start=0, cigar="4M1D2M2S", attrs={"MD": "4^T2", "NM": 1}
```
nh13, line 781: Can you add this one too: https://github.com/samtools/htsjdk/blob/7034b33636b4cb9fec300a2136588e7c12c7ccd5/src/test/resources/htsjdk/samtools/SequenceUtil/upper_and_lowercase_read.sam#L9C27-L9C35?

fgpyo/sam/__init__.py
nh13: I am so sorry, I read this as `elif op in (CigarOp.H, CigarOp.P)`. Consider `assert_never`.  Sorry again

fgpyo/sam/__init__.py
znorgaard: That's on me! I suggested we do the "short-circuit early" pattern we tend to prefer for ease of reading/reasoning. @emmcauley, when you change this back to the order you originally had, can you add a comment that the ordering is for performance?

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,96 @@ class ReadEditInfo:
...
+    reference sequence.
 
     Args:
         rec: the read/record for which to calculate values
-        reference_sequence: the reference sequence (or fragment thereof) that the read is
-            aligned to
+        reference_sequence: the reference sequence (or fragment thereof) to which the read is
+            aligned
+        n_as_match: if True, mirror htsjdk `calculateMdAndNmTags` and treat N->N as a match. If
+            False, N->N is treated as a mismatch. (default: True)
```
znorgaard, line 1075: I know I originally said we should match htsjdk, but I think we should match the SAM spec and set this to False by default. And update the doc string that `False` is the spec compliant option.

Side note, tt's still not clear to me why htsjdk deviates and why you might chose the htsjdk implementation over the spec compliant version beyond compatibility.

fgpyo/sam/__init__.py
```diff
@@ -1087,45 +1090,51 @@ def calculate_edit_info(  # noqa: C901 (11 > 10)
...
                     md_edits.append(ref_base)  # grab mismatched base from the reference
+                    mms += 1
                 else:  # match
                     matches += 1
                     current_match_count += 1
             query_offset += elem.length_on_query
             target_offset += elem.length_on_target
+        elif op == CigarOp.D:  # consumes ref
+            md_edits.append(str(current_match_count))  # append match count and reset
+            md_edits.append("^")
```
emmcauley, line 1120: we append before we break here

fgpyo/sam/__init__.py
emmcauley: added: `test_calc_edit_info_with_equals_in_query`

fgpyo/sam/__init__.py
emmcauley: fixed

tests/fgpyo/sam/test_sam.py
emmcauley: added, L790 in tests: `@pytest.mark.parametrize("cigar", ["4M1D2M2N", "4M1D2M2H", "4M1D2M2P"])`

fgpyo/sam/__init__.py
emmcauley: I could not get `assert_never()` to work here. I reviewed it with Clint and it sounded like a potential rabbit hole?
```
elif op not in (CigarOp.H, CigarOp.P):  # pragma: no cover
  assert_never(op)

error: Argument 1 to "assert_never" has incompatible type "Literal[CigarOp.M, CigarOp.I, CigarOp.S, CigarOp.H, CigarOp.P, CigarOp.EQ, CigarOp.X, CigarOp.code, CigarOp.character, CigarOp.consumes_query, CigarOp.consumes_reference]"; expected "Never"  [arg-type]
```

fgpyo/sam/__init__.py
emmcauley: I added a comment in the docstring about the ordering

fgpyo/sam/__init__.py
emmcauley: fixed -- we append then break

fgpyo/sam/__init__.py
emmcauley: understood, this stuck out to me too. I've set the `n_as_match` default to False to be concordant with the SAM spec.

tests/fgpyo/sam/test_sam.py
emmcauley: added: `test_calc_edit_info_all_matches`

emmcauley:
@nh13 can you please confirm this is the behavior you want? I changed the results of this test and I just want to be careful that this is the expected behavior:
```
def test_calc_edit_info_with_matches_out_of_bounds() -> None:
    """Assert expected behavior for matches that extends out of bounds of the read."""
    chrom = "AGTCCGTTA"
    builder = SamBuilder(r1_len=4)
    rec = builder.add_single(bases="TTAG", chrom="chr1", start=0, cigar="6D4M")

    info = sam.calculate_edit_info(rec=rec, reference_sequence=chrom, n_as_match=False)
    # Offset Ref: AGTCCGTTA
    #             xxxxxx|||.
    #      Query: ------TTAG
    assert info.mismatches == 0
    assert info.insertions == 0
    assert info.inserted_bases == 0
    assert info.deletions == 0
    assert info.deleted_bases == 0
    assert info.nm == 0
    assert info.md == "0^AGTCCG0"  # report AGTCCG deletion, 3 matches are out of bounds
```

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1040,105 @@ class ReadEditInfo:
...
+        # TODO: use match-case statements after we drop Python 3.9 support
+        if op in (CigarOp.M, CigarOp.X, CigarOp.EQ):
+            for in_block_offset in range(0, elem.length):
+                if (target_offset + in_block_offset) >= len(reference_sequence):
+                    break  # out of bounds
+                query_base: str = rec.query_sequence[query_offset + in_block_offset].upper()
+                ref_base: str = reference_sequence[target_offset + in_block_offset].upper()
+
+                if (query_base != ref_base and query_base != "=") or (
+                    query_base not in ok_bases and not n_as_match
```
tfenne, line 1107: This logic appears to deviate in an unexpected way from the HTSJDK implementation.  I understand adding `n_as_match` to allow for spec-compliant or HTSJDK style behavior, but with `n_as_match` set to true this now does something quite different from HTSJDK.

Reviewing the HTSJDK implementation I see:

```java
 if ((bases[readBase] == bases[refBase]) || readBase == 0) {
    // a match
    ++matchCount;
} else {
    mdString.append(matchCount);
    mdString.appendCodePoint(refBase);
    matchCount = 0;
    ++nmCount;
}
```

What this does is count a read base as matching _if it matches the reference base_, regardless of whether the bases are A/C/G/T or some other base.  However the python code here will allow any non-ACGT base in the read to match _any_ base in the reference if `n_as_match` is set.  Nowhere does it check a) the read base is an `N` (in theory SAM files can contain other ambiguity codes), and b) nowhere does it check that the read base and the reference base are the same.

So with `n_as_match` set, a read-base of `N` vs. a reference base of `C` will be treated as a match.  The result is that an MD string will be generated that is insufficient to reconstruct the reference sequence, which is not the case with the HTSJDK implementation.

emmcauley: as discussed, I've added some logic to handle the missing branch that you highlighted TIm. I'll mark this as resolved since we chatted on Slack.

fgpyo/sam/__init__.py
```diff
@@ -201,9 +201,15 @@
 NO_REF_POS: int = -1
 """The reference position to use to indicate no position in SAM/BAM."""
 
+NO_QUERY_BASES: str = "*"
+"""The string to use for a SAM record with missing query bases."""
+
 NO_QUERY_QUALITIES: array = qualitystring_to_array(STRING_PLACEHOLDER)
 """The quality array corresponding to an unavailable query quality string ("*")."""
 
+OK_BASES: set[str] = {"A", "C", "G", "T", "="}
```
nh13, line 210: suggestion: shall we make this private for now since it's only used here and not intended to be use by outside projects
```suggestion
_OK_BASES: set[str] = {"A", "C", "G", "T", "="}
```

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1043,119 @@ class ReadEditInfo:
...
+                    mms += 1
+            query_offset += elem.length_on_query
+            target_offset += elem.length_on_target
+        elif op == CigarOp.D:  # consumes ref
+            md_edits.append(str(current_match_count))  # append match count and reset
+            md_edits.append("^")
+            for in_block_offset in range(0, elem.length):
+                if (target_offset + in_block_offset) >= len(reference_sequence):
+                    break  # out of bounds
+            md_edits.append(f"{reference_sequence[target_offset:target_offset + elem.length]}")
```
nh13, line 1141: chore: (1) I think we need to upper case the reference and (2) I don't think we need an f-string since reference sequence is already a string
```suggestion
            md_edits.append(reference_sequence[target_offset:target_offset + elem.length].upper())
```

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1043,119 @@ class ReadEditInfo:
...
+        return None
 
-    query_offset = 0
-    target_offset = reference_offset if reference_offset is not None else rec.reference_start
-    cigar = Cigar.from_cigartuples(rec.cigartuples)
+    query_offset: int = 0
+    target_offset: int = reference_offset if reference_offset is not None else rec.reference_start
+    cigar: Cigar = Cigar.from_cigartuples(rec.cigartuples)
 
     matches, mms, insertions, ins_bases, deletions, del_bases = 0, 0, 0, 0, 0, 0
```
nh13, line 1109: chore: while we're here, can we update this variable to a longer form?
```suggestion
    matches, mismatches, insertions, ins_bases, deletions, del_bases = 0, 0, 0, 0, 0, 0
```

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1043,119 @@ class ReadEditInfo:
...
+        elif op == CigarOp.D:  # consumes ref
+            md_edits.append(str(current_match_count))  # append match count and reset
+            md_edits.append("^")
+            for in_block_offset in range(0, elem.length):
+                if (target_offset + in_block_offset) >= len(reference_sequence):
+                    break  # out of bounds
+            md_edits.append(f"{reference_sequence[target_offset:target_offset + elem.length]}")
+            current_match_count = 0
+            if target_offset < elem.length:
+                break
```
nh13, line 1144: chore:  I found another difference between htsjdk and samtools:

1. [htsjdk will break _before_ updating the `NM` value](https://github.com/samtools/htsjdk/blob/b16416c078e2df3e3a21486207d0c0c31b293cb9/src/main/java/htsjdk/samtools/util/SequenceUtil.java#L1014-L1016)
2. [samtools will break _after_ updating the `NM` value](https://github.com/samtools/samtools/blob/f2c61f7ac8df069426ef1b74bf3ac163a5331d5b/bam_md.c#L120-L121)

What do you think about updating `deletions` and `del_bases` if `match_htsjdk` is `false`?

fgpyo/sam/__init__.py
```diff
@@ -1036,55 +1043,119 @@ class ReadEditInfo:
...
+            query_offset += elem.length
+            if op == CigarOp.I:
+                insertions += 1
+                ins_bases += elem.length
+        elif op == CigarOp.N:  # skipped region from ref, consumes ref
+            target_offset += elem.length
+        elif op not in (CigarOp.H, CigarOp.P):  # pragma: no cover
+            raise ValueError(f"Invalid CIGAR operation: {op}")
+
+    md_edits.append(str(current_match_count))
```
nh13, line 1158: nit: f-strings are just a little faster
```suggestion
    md_edits.append(f"{current_match_count}")
```

nh13 (changes requested):

fgpyo/sam/__init__.py
```diff
@@ -1044,64 +1043,133 @@ class ReadEditInfo:
...
-        query_offset += elem.length_on_query
-        target_offset += elem.length_on_target
+        elif op in (CigarOp.I, CigarOp.S):  # consumes query
+            query_offset += elem.length
+            if op == CigarOp.I:
+                insertions += 1
+                ins_bases += elem.length
+        elif op == CigarOp.N:  # skipped region from ref, consumes ref
+            target_offset += elem.length
+        elif op not in (CigarOp.H, CigarOp.P):  # pragma: not covered
```
nh13:
```suggestion
        elif op not in (CigarOp.H, CigarOp.P):  # pragma: no cover
```

nh13 (approved):
One pragma issue, then LGTM.

znorgaard (approved):
Nice work!

PR #231, opened 2025-03-27
fix: typing_extensions needed for @deprecated up until python 3.12

nh13:
We need to use `typing_extensions` for python 3.12, as the [`deprecated` decorator wasn't added until python 3.13](https://docs.python.org/3/library/warnings.html#warnings.deprecated).  This causes import issues when using fgpyo on python 3.12 or above.

```console
Traceback (most recent call last):
  File "foo.py", line 4, in <module>
    from fgpyo.sam import Cigar
  File "/path/to/python3.12/site-packages/fgpyo/sam/__init__.py", line 183, in <module>
    from typing_extensions import Self
ModuleNotFoundError: No module named 'typing_extensions'
```


fgpyo/sam/__init__.py
```diff
@@ -180,8 +180,16 @@
 from pysam import AlignmentFile as SamFile
 from pysam import AlignmentHeader as SamHeader
 from pysam import qualitystring_to_array
-from typing_extensions import Self
-from typing_extensions import deprecated
+
+if sys.version_info >= (3, 10):
```
nh13, line 184:
```suggestion
if sys.version_info > (3, 10):
```

fgpyo/sam/__init__.py
```diff
@@ -180,8 +180,16 @@
...
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+if sys.version_info >= (3, 13):
```
nh13, line 189:
```suggestion
if sys.version_info > (3, 12):
```

fgpyo/sam/__init__.py
```diff
@@ -180,8 +180,16 @@
 from pysam import AlignmentFile as SamFile
 from pysam import AlignmentHeader as SamHeader
 from pysam import qualitystring_to_array
-from typing_extensions import Self
-from typing_extensions import deprecated
+
+if sys.version_info > (3, 10):
```
nh13, line 184: Added in 3.11: https://docs.python.org/3/library/typing.html#typing.Self

fgpyo/sam/__init__.py
```diff
@@ -180,8 +180,16 @@
...
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+if sys.version_info > (3, 12):
```
nh13, line 189: Added in 3.13: https://docs.python.org/3/library/warnings.html#warnings.deprecated

fgpyo/sam/__init__.py
clintval: Because `sys.version_info()` has a `micro` field it will be better to do:

```suggestion
if sys.version_info[:2] >= (3, 11):
```

For example, see the full tuple like this:

```python
>>> tuple(sys.version_info)
(3, 11, 7, 'final', 0)
```

fgpyo/sam/__init__.py
clintval: You should do this instead so you are not affected by the `micro` field of the version.

```suggestion
if sys.version_info[:2] >= (3, 13):
```

clintval (approved):
Approve after you fixup the handling of micro version numbers which fall outside your conditional check. Before my suggestion a statement like:

```python
sys.version_info > (3, 10)
```

Would return `True` for versions like `(3, 10, 1, ...)` which is not the intent of the check.

PR #234, opened 2025-04-16
fix: always restore stderr, even with an exception

nh13:
This was causing exceptions not to be printed to stderr.

The following would print nothing:

```console
>>> from fgpyo.io import suppress_stderr
>>> with suppress_stderr():
...    pysam.AlignmentFile("/does/not/exist.bam")
... 
```


yfarjoun (approved):
thanks! LGTM!

clintval:
Hey @coderabbitai how would you write a unit test for this PR that would fail with the previous implementation but succeed with the new implementation?

Issue #104, opened 2024-04-17
Metric wishlist


msto:
Hey @nh13 @tfenne ,

I'd like to consolidate a wishlist of features/improvements to the `Metric` class. 

Their absence has blocked the use of `Metric` in client contexts (of note, #103 being blocked required me to re-implement much of the `Metric` functionality from scratch on a client project), and I'd greatly appreciate their inclusion in the `fgpyo` roadmap.

In rough order of perceived priority:
- [ ] `Metric` should not be dependent on `pysam`.
    - It would be helpful if `Metric` (and other non-pysam-dependent modules) could be spun out into a separate package, or if  `pysam` could be made an optional dependency.
- [ ] https://github.com/fulcrumgenomics/fgpyo/issues/88
- [ ] https://github.com/fulcrumgenomics/fgpyo/issues/90
- [ ] https://github.com/fulcrumgenomics/fgpyo/pull/103
- [ ] https://github.com/fulcrumgenomics/fgpyo/issues/89
    - I think this would require dropping support for 3.8/3.9 and don't know if it's viable
- [ ] https://github.com/fulcrumgenomics/fgpyo/issues/70

I've written proof-of-concept solutions to these in https://github.com/msto/dataclass_io/, and am curious which might be considered for adoption by `fgpyo`. I would very much like to see `Metric` become the general solution for serializing/deserializing dataclasses in our client repos, but I've been hesitant to open additional PRs on `fgpyo` to address these issues due to lack of clarity around which of these features are compatible with the vision for `Metric` and `fgpyo`.

tfenne:
@msto I have no objections to any of the issues linked; I might have input on implementation of those but I'm 100% on board with the functionality being added.  With the possible exception of #89 if we really can't find a way to make it back-compat, or guarded by a check to see if the python version in use is high enough to support it.

Re pysam ... I am also on board with finding a way to use the rest of fgpyo without forcing the pysam dependency.  I don't know enough about how optional dependencies work in python/pypi (and by extension conda).  That said, I think I could easily :+1: a solution that looks something like:

- Move pysam to a `dev` dependency
- Clearly state in docs that `fgpyo.sam` requires a version of pysam >= x.y.z
- In the `__init__.py` for the sam module (or similar place) check to see if pysam is available and if it is not then raise an exception with a super-clear warning about what's going on

I think that would be preferable to trying to split fgpyo in two, but I'm open to being convinced of better solutions.

msto:
[...]

PR #238, opened 2025-06-10
chore: Make `pysam` an optional dependency

msto:
## Summary

This PR implements a proposal by Tim Fennell a number of months ago to make `pysam` an optional dependency[^1].

After this PR is merged, `pysam` _will not_ be installed by default when installing `fgpyo` - it will be an opt-in dependency via `pip install fgpyo[pysam]`. Features which depend on `pysam` will not be supported when `pysam` has not been installed.

This will make `fgpyo` a significantly lighter dependency when developers wish to use its pure Python features such as `Metric`. 

[^1]: https://github.com/fulcrumgenomics/fgpyo/issues/104#issuecomment-2061454887

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,11 +71,11 @@ jobs:
 
       - name: Verify pysam is not installed
         run: |
-          uv run python -c "import pysam" 2>&1 && exit 1 || echo "pysam correctly not installed"
+          uv run --no-sync python -c "import pysam" 2>&1 && exit 1 || echo "pysam correctly not installed"
 
       - name: Run pysam-independent tests
         run: |
-          uv run pytest \
+          uv run --no-sync pytest \
             tests/fgpyo/test_read_structure.py \
             tests/fgpyo/test_sequence.py \
             tests/fgpyo/test_requirements.py \
@@ -90,4 +90,4 @@ jobs:
 
       - name: Verify pysam-dependent modules are importable
         run: |
-          uv run python -c "from fgpyo.sam import Cigar; print('fgpyo.sam importable without pysam')"
+          uv run --no-sync python -c "from fgpyo.sam import Cigar; print('fgpyo.sam importable without pysam')"


msto:
@clintval @nh13 @tfenne I am opening this draft PR for review now because I would like to collect early feedback on the pattern being implemented.

fgpyo/fastx/__init__.py
```diff
@@ -28,18 +28,27 @@
...
+from fgpyo._optional_dependencies import HAS_PYSAM
+from fgpyo._optional_dependencies import require_pysam
 
+if TYPE_CHECKING:
+    import pysam
+else:
+    if HAS_PYSAM:
+        import pysam
+    else:
+        pysam = None
```
msto, line 48: The plan would be to just stamp this into all the modules that use `pysam`, and add a call to `require_pysam()` at the top of each function (or class init method) that requires it

msto: Yeah that worked a charm

fgpyo/fastx/__init__.py
```diff
@@ -35,11 +35,11 @@
 from typing import Type
 from typing import Union
 
-from pysam import FastxFile
-from pysam import FastxRecord
+from fgpyo._optional_dependencies import pysam
+from fgpyo._optional_dependencies import require_pysam
```
msto, line 39: @clintval @nh13 @tfenne If you're happy with this pattern, I can go through and update all the modules accordingly!

msto:
@clintval Do you have suggestions or preferences for how having an optional dependency should be tested?

msto:
Closing in favor of the module-level guards in #270. Otherwise we have to use from `__future__ import annotations` to protect forward references to `pysam.AlignedSegment` (e.g. on `Template attributes`), and that doesn't play nicely with the runtime type introspection in `fgpyo.util.inspect`, `pydantic`, or `attrs`.

PR #246, opened 2025-08-13
feat: `require()`

### LINKED ISSUE #132: Add `require`
`assert` should be avoided outside unit tests.

The Google style guide [requires this](https://google.github.io/styleguide/pyguide.html#244-decision):
> Do not use `assert` statements in place of conditionals or validating preconditions. They must not be critical to the application logic. A litmus test would be that the `assert` could be removed without breaking the code. `assert` conditionals are [not guaranteed](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) to be evaluated. For [pytest](https://pytest.org/) based tests, `assert` is okay and expected to verify expectations.

The primary reason to avoid `assert` statements is that [they are intended for debugging and can be disabled](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

Per slack discussion, it would be convenient to have a `require()` function in `fgpyo` that acts like `assert` but cannot be disabled.

```python
def require(
    condition: bool,
    message: str | None = None,
    error: Exception = RequirementError,
) -> None:
    """
    Require a condition to be met.

    Args:
        condition: The test condition.
        message: The error message to report.
        error: The exception class to raise.

    Raises:
        `RequirementError`: if `condition` is False. (The `Exception` class raised may be overridden by the `error` argument.
    """
    ...
```

msto:
## Summary

Closes #132 

I've been using this implementation in a client repo and found it helpful, so thought it would be good to transfer it over.

I do still think the utility presents a possible footgun for coverage checks, since both the happy and unhappy paths share a single line of code, and we may want to consider including a warning in the documentation.

### DIFF
--- a/fgpyo/__init__.py
+++ b/fgpyo/__init__.py
@@ -1,7 +1,15 @@
 from importlib.metadata import PackageNotFoundError
 from importlib.metadata import version
 
+from fgpyo._requirements import RequirementError
+from fgpyo._requirements import require
+
 try:
     __version__ = version("fgpyo")
 except PackageNotFoundError:
     __version__ = "dev"  # package is not installed
+
+__all__ = [
+    "require",
+    "RequirementError",
+]
--- a/fgpyo/_requirements.py
+++ b/fgpyo/_requirements.py
@@ -7,18 +7,18 @@ class RequirementError(ValueError):
     """Exception raised when a requirement is not satisfied."""
 
 
-def require(condition: bool, msg: Optional[str] = None) -> None:
+def require(condition: bool, message: Optional[str] = None) -> None:
     """Require a condition be satisfied.
 
     Args:
         condition: The condition to satisfy.
-        msg: An optional message to include with the error when the condition is false.
+        message: An optional message to include with the error when the condition is false.
 
     Raises:
         RequirementError: If the condition is false.
     """
     if not condition:
-        if msg is not None:
-            raise RequirementError(msg)
+        if message is not None:
+            raise RequirementError(message)
         else:
-            raise RequirementError
+            raise RequirementError()
--- a/fgpyo/_requirements.py
+++ b/fgpyo/_requirements.py
@@ -1,0 +1,24 @@
+"""Enforce requirements."""
+
+from typing import Optional
+
+
+class RequirementError(ValueError):
+    """Exception raised when a requirement is not satisfied."""
+
+
+def require(condition: bool, msg: Optional[str] = None) -> None:
+    """Require a condition be satisfied.
+
+    Args:
+        condition: The condition to satisfy.
+        msg: An optional message to include with the error when the condition is false.
+
+    Raises:
+        RequirementError: If the condition is false.
+    """
+    if not condition:
+        if msg is not None:
+            raise RequirementError(msg)
+        else:
+            raise RequirementError
--- a/fgpyo/_requirements.py
+++ b/fgpyo/_requirements.py
@@ -1,24 +1,29 @@
 """Enforce requirements."""
 
-from typing import Optional
+from typing import Callable
+from typing import Union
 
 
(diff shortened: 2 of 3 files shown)


fgpyo/__init__.py
```diff
@@ -0,0 +1,7 @@
+from fgpyo._requirements import RequirementError
+from fgpyo._requirements import require
+
+__all__ = [
+    "require",
+    "RequirementError",
+]
```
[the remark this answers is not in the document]
clintval: I think this works well for me, but by definining `__all__` here you make it so only `require` and `RequirementError` are imported with `from fgpyo import *` which might be surprising (not that anyone should be doing wildcard import anyway...).

It's long overdue but I'd like to tackle this type of issue and others with a reorganization of the actual code files:

- https://github.com/fulcrumgenomics/fgpyo/issues/136

My preferred pattern is to use `__init__.py` for defining `__all__` (as you've done) and having all code in protected files/modules like `fgpyo/io/_io.py`. Then, if anyone where to use a wildcard import, or use their auto-complete feature in their IDEA, they'd only see the true public API and not all importable things that we never meant for people to import.

fgpyo/_requirements.py
```diff
@@ -0,0 +1,24 @@
+"""Enforce requirements."""
+
+from typing import Optional
+
+
+class RequirementError(ValueError):
+    """Exception raised when a requirement is not satisfied."""
```
clintval, line 7: I can think of some requirements that wouldn't necessary relate to values.

Wondering if it'd be better to inherit from `AssertionError` instead?

But since you are not using `assert` below (probably to avoid someone turning off the behavior in Python's debug mode), maybe inheriting from the `Exception` class is ideal.

fgpyo/_requirements.py
```diff
@@ -0,0 +1,24 @@
...
+        msg: An optional message to include with the error when the condition is false.
+
+    Raises:
+        RequirementError: If the condition is false.
+    """
+    if not condition:
+        if msg is not None:
+            raise RequirementError(msg)
+        else:
+            raise RequirementError
```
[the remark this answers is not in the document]
msto: I've fixed both in 7f2e666df5fb480dc5aec33f65db94da73865f62

clintval (approved):
I'm OK with this, but request we think about which exception base class to inherit from a little more carefully.

fgpyo/_requirements.py
msto: And the standard library docs explicitly instructs developers to inherit from `Exception` when defining their own exceptions. (emphasis mine)

> All built-in, non-system-exiting exceptions are derived from this class. **All user-defined exceptions should also be derived from this class.**

https://docs.python.org/3/library/exceptions.html#Exception

TIL!

fgpyo/_requirements.py
```diff
@@ -0,0 +1,24 @@
+"""Enforce requirements."""
+
+from typing import Optional
+
+
+class RequirementError(Exception):
+    """Exception raised when a requirement is not satisfied."""
+
+
+def require(condition: bool, message: Optional[str] = None) -> None:
```
tfenne, line 10: Scala has a nice syntax for arguments where the type can be e.g. `message: => str` which means "any expression or function that yields a str".  It's used places like this to defer evaluation of the message to when it's used (i.e. only if the requirement fails).  I'm not sure there's a nice way to do this in python though.

nh13: `Callable[[], str]` would be the type from `collections.abc.Callable`

msto: Added in 6de405e9afb9c76b906fcf4a4c5871ff589acdf4.

Happy to keep or revert if you don't prefer the Python variation of the syntax, up to you!

nh13 (approved):

PR #252, opened 2025-09-10
feat: make fastx mismatch name error deterministic

theJasonFan:
### Summary:

Motivation: I was working in a project and wanted to test the error raised when read names are mismatched for input reads to `FastxZipped` and found that it was non-deterministic.

Changes:
- feat: updates error message to be deterministic and report all read names in a zipped record if said names are mismatched.

### DIFF
--- a/fgpyo/fastx/__init__.py
+++ b/fgpyo/fastx/__init__.py
@@ -29,8 +29,8 @@ from contextlib import AbstractContextManager
 from pathlib import Path
 from types import TracebackType
 from typing import Iterator
+from typing import List
 from typing import Optional
-from typing import Set
 from typing import Tuple
 from typing import Type
 from typing import Union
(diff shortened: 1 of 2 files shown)


fgpyo/fastx/__init__.py
```diff
@@ -76,9 +76,9 @@ def __next__(self) -> Tuple[FastxRecord, ...]:
                 )
             )
         else:
-            record_names: Set[str] = {self._name_minus_ordinal(record.name) for record in records}
-            if len(record_names) != 1:
-                raise ValueError(f"FASTX record names do not all match: {record_names}")
+            record_names: List[str] = [self._name_minus_ordinal(record.name) for record in records]
+            if len(set(record_names)) != 1:
+                raise ValueError(f"FASTX record names do not all match, found: {record_names}")
```
msto, line 81: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
The interpolation of `record_names` in the error message uses the original list, so the error message will include duplicate record names. 

I usually use `sorted(set(iter))` to deduplicate deterministically - would that be okay here?

```suggestion
            record_names: list[str] = sorted({self._name_minus_ordinal(record.name) for record in records})
            if len(record_names) != 1:
                raise ValueError(f"FASTX record names do not all match, found: {record_names}")
```

msto (approved):

fgpyo/fastx/__init__.py
theJasonFan: I did think about this. I thought that it was more useful to see all record names for the zipped record in the order provided to FastxZipped. I'd want to have a quick way to check which fastx file caused errors.

Issue #253, opened 2025-09-11
`sam.reader` fails silently if there are no SQ headers in a BAM file and `unmapped=False`


ameynert:
Input is an unsorted BAM file that does not have SQ headers, read content does not seem to matter.

Example anonymized header:
```
@HD     VN:1.6  SO:unsorted     GO:query
@RG     ID:A    SM:1  LB:1 PL:illumina
```

```
from pathlib import Path
from fgpyo.sam import reader

def get_query_names(bam: Path) -> set[str]:
    query_names: set[str] = set()
    with reader(bam) as bam_reader:
        for record in bam_reader:
            if record.query_name is None:
                raise ValueError(f"Record in {bam} is missing a query name.")
            query_names.add(record.query_name)
    return query_names
```

Adding `unmapped=True` when creating the reader works.

The same function implemented with `pysam.AlignmentFile` works, but only if you set `check_sq=False`. It throws an appropriate error if that is not set.

```
from pathlib import Path
from pysam import AlignmentFile

def get_read_names(bam: Path) -> set[str]:
    query_names: set[str] = set()
    samfile = AlignmentFile(str(bam), "rb", check_sq=False)
    for record in samfile:
        if record.query_name is None:
            raise ValueError(f"Record in {bam} is missing a query name.")
        query_names.add(record.query_name)
    samfile.close()
```

The tool crashes with no error message when using the `sam.reader` implementation of the function without `unmapped=True`. There should be an error thrown with a useful message to the user, along the lines of "No sequence dictionary found. Consider setting unmapped=True when creating the reader."

PR #256, opened 2025-09-25
Fix/issue 253

nh13:
Fixes: #253

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -306,19 +306,23 @@ def _pysam_open(  # noqa: C901
     if unmapped and open_for_reading:
         kwargs["check_sq"] = False
 
-    # Open it alignment file, suppressing stderr in case index files are older than SAM file
+    # Open the alignment file, suppressing stderr in case index files are older than the SAM/BAM.
     try:
         with fgpyo.io.suppress_stderr():
             alignment_file = pysam.AlignmentFile(path, **kwargs)
-    except ValueError as ex:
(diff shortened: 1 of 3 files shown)


fgpyo/sam/__init__.py
```diff
@@ -307,8 +307,19 @@ def _pysam_open(
...
+            alignment_file = pysam.AlignmentFile(path, **kwargs)
+    except ValueError as ex:
+        if "Consider opening with check_sq=False" in f"{ex}":
+            raise ValueError(
+                "No SQ (sequence) lines in the header."
+                " Is this SAM/BAM unmapped?"
+                " Consider opening with `unmapped=True`"
+            ) from ex 
+        else:
+            raise ex
```
nh13, line 321: This improves the error message (since in our API, we use `unmapped=True`)

tests/fgpyo/sam/test_sam.py
```diff
@@ -119,6 +119,11 @@ def test_unmapped_sam_file_open_reading_with_reader(unmapped_sam: Path) -> None:
         assert len([read for read in samfile.fetch() if read.is_unmapped]) == 1
 
 
+def test_unmapped_sam_open_reading_exception(unmapped_sam: Path) -> None:
+    with pytest.raises(ValueError):
+        sam.reader(path=unmapped_sam)
```
ameynert, line 124: [suggestion](https://pullpo.io/cc?l=suggestion) 
Check the error message is what you expect

PR #257, opened 2025-09-25
feat: SamBuilder can output SAM/BAM/CRAM

nh13:
See: https://github.com/myriad-opensource/samwell/pull/26

### DIFF
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -242,7 +242,7 @@ class SamFileType(enum.Enum):
     @property
     def indexable(self) -> bool:
         """True if the file type can be indexed, false otherwise."""
-        return self == SamFileType.BAM or self == SamFileType.CRAM
+        return self is SamFileType.BAM or self is SamFileType.CRAM
 
     @classmethod
     def from_path(cls, path: Union[Path, str]) -> "SamFileType":
--- a/fgpyo/sam/__init__.py
+++ b/fgpyo/sam/__init__.py
@@ -239,6 +239,11 @@ class SamFileType(enum.Enum):
     BAM = ("b", ".bam")
     CRAM = ("c", ".cram")
 
+    @property
+    def indexable(self) -> bool:
+        """True if the file type can be indexed, false otherwise."""
+        return self == SamFileType.BAM or self == SamFileType.CRAM
+
     @classmethod
     def from_path(cls, path: Union[Path, str]) -> "SamFileType":
         """Infers the file type based on the file extension.
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -519,7 +519,7 @@ class SamBuilder:
         Returns:
             Path: The path to the sorted (and possibly indexed) file.
         """
-        ext = ".sam" if file_type == sam.SamFileType.SAM else ".bam"
+        ext = file_type.extension
 
         if path is None:
             with NamedTemporaryFile(suffix=ext, delete=False) as fp:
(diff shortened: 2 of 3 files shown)


fgpyo/sam/builder.py
```diff
@@ -503,34 +503,36 @@ def to_path(
         path: Optional[Path] = None,
         index: bool = True,
         pred: Callable[[AlignedSegment], bool] = lambda r: True,
+        file_type: sam.SamFileType = sam.SamFileType.BAM,
```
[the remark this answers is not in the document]
clintval: I still think there is going to be an issue. If you accidentally define a path with a different suffix than what is requested in the `file_type` argument, then the latter pysam sort command will override your initial file format (say `sam.SamFileType.BAM`) and convert to the format as-defined by the suffix of your input path.

Perhaps the best way to deal with this is to let pysam do it's auto-format detection, but for temporary files, force the BAM format.

fgpyo/sam/__init__.py
```diff
@@ -239,6 +239,11 @@ def __init__(self, mode: str, ext: str) -> None:
     BAM = ("b", ".bam")
     CRAM = ("c", ".cram")
 
+    @property
+    def indexable(self) -> bool:
+        """True if the file type can be indexed, false otherwise."""
+        return self == SamFileType.BAM or self == SamFileType.CRAM
```
clintval, line 245: Normally best practice to use identity comparison for enums:


```suggestion
        return self is SamFileType.BAM or self is SamFileType.CRAM
```

fgpyo/sam/builder.py
```diff
@@ -503,34 +503,49 @@ def to_path(
...
+        is provided.  See `~pysam` for more details.
+
+        If `path` is not provided, the file type will default to BAM unless `tmp_file_type` is
+        provided.
+
         Args:
             path: a path at which to write the file, otherwise a temp file is used.
-            index: if True and `sort_order` is `Coordinate` index is generated, otherwise not.
+            index: if True and `sort_order` is `Coordinate` and output is a BAM file, then
+                   an index is generated, otherwise not.
```
clintval, line 522:
```suggestion
            index: if True and `sort_order` is `Coordinate` and output is a BAM/CRAM file,
                   then an index is generated, otherwise not.
```

fgpyo/sam/builder.py
```diff
@@ -503,34 +503,49 @@ def to_path(
...
+
+        ext = tmp_file_type.extension
 
         if path is None:
-            with NamedTemporaryFile(suffix=".bam", delete=False) as fp:
+            with NamedTemporaryFile(suffix=ext, delete=False) as fp:
                 path = Path(fp.name)
 
-        with NamedTemporaryFile(suffix=".bam", delete=True) as fp:
+        with NamedTemporaryFile(suffix=ext, delete=True) as fp:
```
clintval, line 541: All the if-elses and context managers are making this tricky for me to logic through. I'm still wondering if this works correctly for all cases. It looks like _**if**_ a path is defined, then the extension (`ext`) is always going to be BAM. However, if the sort order is **_neither_** queryname or coordinate, there is no pysam sort to the final file (by filename, not by open file handle).

So, for this combination of values, do you indeed get a named SAM file?

- Path suffixed with `.sam`
- Sort order _**neither**_ queryname or coordinate
- Leaving `tmp_file_type` as None

clintval: I see code rabbit spotted the same issue. I testify I saw it first despite being late to hit "submit review" ;)

PR #258, opened 2025-09-26
build: Modernize pyproject.toml

### LINKED ISSUE #224: Modernize pyproject.toml
See: https://github.com/fulcrumgenomics/fgpyo/pull/223/files#r1970206805

msto:
# Summary

Closes #224 

This PR modernizes the package's `pyproject.toml` to better align it with our [template](https://github.com/fulcrumgenomics/python-template).

# Changes

- Added `poethepoet` as a dependency, and defined our usual `poe` tasks for static analysis and unit tests.
- Updated `ci/check.sh` and our GitHub Actions to use these poe tasks instead of explicit commands.
- Bumped `ruff` and `mypy` to latest versions, and bumped `ruff` target from py38 to py39 (the earliest version this package currently supports).

### DIFF
--- a/ci/check.sh
+++ b/ci/check.sh
@@ -7,7 +7,7 @@ if [ ! -x "$(command -v uv)" ]; then
     exit 1
 fi
 
-set +euo pipefail
+set -euo pipefail
 
 uv run poe fix-and-check-all
 uv run poe build-docs


.github/workflows/tests.yml
```diff
@@ -19,58 +22,31 @@ jobs:
...
-        run: uv run pytest --cov=fgpyo --cov-report=xml --cov-branch 
+      - name: Test the library
+        run: |
+          uv run poe check-all
 
       - name: Run docs
         run: |
           set -euo pipefail
-          uv run mkdocs build --strict
+          uv run poe build-docs
```
nh13, line 47: [question(non-blocking)](https://pullpo.io/cc?l=question&d=non-blocking) 
why not include `build-docs` as part of `check-all`?

ci/check.sh
```diff
@@ -1,58 +1,11 @@
...
 fi
 
-run "Checking lockfile"    "uv lock --check"
-run "Updating environment" "uv sync --locked"
-run "Style Checking"       "uv run ruff format fgpyo tests"
-run "Linting"              "uv run ruff check --fix fgpyo tests"
-run "Type Checking"        "uv run mypy fgpyo tests --config pyproject.toml"
-run "Unit Tests"           "uv run pytest -vv -r sx tests"
-run "Make docs"            "uv run mkdocs build --strict"
+uv run poe fix-and-check-all
```
nh13, line 10: [issue(blocking)](https://pullpo.io/cc?l=issue&d=blocking) 
This skips building the docs.  Can you include that in `fix-and-check-all`?

pyproject.toml
```diff
@@ -111,3 +114,50 @@ exclude_lines = [
...
+fix-and-check-all.ignore_fail = "return_non_zero"
+fix-and-check-all.sequence    = [
+    "check-lock",
+    "fix-format",
+    "fix-lint",
+    "check-typing",
+    "check-tests"
+]
+
+build-docs = "mkdocs build --strict"
```
nh13, line 163: [suggestion](https://pullpo.io/cc?l=suggestion) 
can `build-docs` be part of `check-all` and `fix-and-check-all`?

nh13 (changes requested):

.github/workflows/tests.yml
msto: `build-docs` has side effects, and `check-all` is intended to not have side effects.

I deliberately omitted `build-docs` from `fix-and-check-all` because `build-docs` is slow, and running unit tests should be fast. With `build-docs`, the checks take 7.5 seconds, 7.1 of which are spent on `build-docs`.

```console
INFO    -  Building documentation to directory: /Users/msto/code/fulcrumgenomics/fgpyo/site
INFO    -  Documentation built in 7.07 seconds
bash ci/check.sh  7.50s user 4.22s system 115% cpu 10.152 total
```

I'm happy to add `build-docs` to `ci/check.sh`, but don't think it should be added to `fix-and-check-all` because it is neither fixing nor checking the code, and because it substantially slows developer iteration.

ci/check.sh
msto: No, see here: https://github.com/fulcrumgenomics/fgpyo/pull/258#discussion_r2384304271

(Since all the PR comments relate to this, I'd like to consolidate discussion in one thread)

ci/check.sh
```diff
@@ -1,58 +1,11 @@
...
-run "Make docs"            "uv run mkdocs build --strict"
-
-if [ -z "$failures" ]; then
-    banner "Checks Passed"
-else
-    banner "Checks Failed with failures in: $failures"
-    exit 1
-fi
+uv run poe fix-and-check-all
+uv run poe build-docs
```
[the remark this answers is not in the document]
nh13: Previously checking the exit code was handled by the `run` function, so I agree we need to have a `set` here

nh13 (approved):
One agreement with the rabbit, then LGTM

ci/check.sh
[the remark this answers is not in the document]
msto: Done!

ci/check.sh
```diff
@@ -1,58 +1,13 @@
...
+set +euo pipefail
 
-if [ -z "$failures" ]; then
-    banner "Checks Passed"
-else
-    banner "Checks Failed with failures in: $failures"
-    exit 1
-fi
+uv run poe fix-and-check-all
+uv run poe build-docs
```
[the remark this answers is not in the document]
msto: fixed, thanks

Issue #247, opened 2025-08-13
Drop support for Python 3.9


msto:
## Summary

This issue is forward-looking to the end of Python 3.9 support in [October 2025](https://devguide.python.org/versions/).

## Python 3.10 Features

Dropping support for Python 3.9 will give us access to the following Python 3.10 features:

1. Structural Pattern Matching (`match-case` statements)
2. Parenthesized context managers (e.g. `with (in_path.open() as fin, out_path.open() as fout)`)
3. Length-Checking in `zip()` (i.e. `zip(..., strict=True)`)
4. Writing Union Types as `X | Y`
5. `typing.TypeGuard` and `typing.TypeAlias` (both available to Python 3.9 through `typing_extensions`)

PR #259, opened 2025-09-27
chore: Modernize mypy config

msto:
# Summary

This PR adds the mypy configuration used by our [template](https://github.com/fulcrumgenomics/python-template) and addresses resulting type errors.

Most of the changes are a consequence of setting `strict_optional = true`, a setting [recommended](https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_optional) by the mypy developers.


fgpyo/util/logging.py
```diff
@@ -159,10 +159,7 @@ def record_alignment(
         Returns:
             true if a message was logged, false otherwise
         """
-        if rec.reference_start is None:
-            return self.record(None, None)
-        else:
-            return self.record(rec.reference_name, rec.reference_start + 1)
+        return self.record(rec.reference_name, rec.reference_start + 1)
```
msto, line 162: `AlignedSegment.reference_start` is of type `int`, so the `if` block was unreachable.

fgpyo/fastx/__init__.py
```diff
@@ -38,6 +38,8 @@
 from pysam import FastxFile
 from pysam import FastxRecord
 
+from fgpyo.util.types import all_not_none
```
msto, line 41: I tried to address the type errors with minimal changes to the code, but here adding a helper to narrow the type of a sequence of `Optional[T]` to `T` seemed like the least bad of the available options.

msto:
Mypy checks are failing on 3.9. I wonder if this is related to our [decision to not enforce a python version in the mypy config](https://github.com/fulcrumgenomics/python-template/pull/19).

TBH I'd prefer to drop support for 3.9 (https://github.com/fulcrumgenomics/fgpyo/issues/247) since it EOLs next month anyways, rather than spend time fixing.

fgpyo/util/types.py
```diff
@@ -216,3 +225,37 @@ def none_parser(value: str) -> Literal[None]:
...
+def all_not_none(values: Sequence[Optional[T]]) -> TypeGuard[Sequence[T]]: ...
+
+
+@overload
+def all_not_none(values: Collection[Optional[T]]) -> TypeGuard[Collection[T]]: ...
+
+
+def all_not_none(values: Iterable[Optional[T]]) -> bool:
+    """
+    Type guard that checks all Optional collection elements are non-null.
```
clintval, line 252:
```suggestion
    Type guard that checks all Optional collection elements are not None.
```

fgpyo/fastx/__init__.py
```diff
@@ -64,22 +66,37 @@ def _name_minus_ordinal(name: str) -> str:
...
+
+        elif not all_not_none(records):
+            non_null_names: List[Optional[str]] = [
+                record.name for record in records if record is not None
+            ]
+            assert all_not_none(non_null_names)  # type narrowing
+            # We know there is at least one non-null record because the previous conditional covers
+            # the case where all records are null, so it is safe to index into the first element of
+            # non_null_names.
+            sequence_name: str = non_null_names[0]
```
clintval, line 81: There isn't a null in Python, so I'd replace null with None in this section.

clintval (approved):
Thank you!

fgpyo/fastx/__init__.py
```diff
@@ -64,22 +66,37 @@ def _name_minus_ordinal(name: str) -> str:
...
-                raise ValueError(f"FASTX record names do not all match, found: {record_names}")
-            return records
+
+        names_with_ordinals: List[Optional[str]] = [record.name for record in records]
+
+        assert all_not_none(names_with_ordinals)  # type narrowing
+        record_names: List[str] = [self._name_minus_ordinal(name) for name in names_with_ordinals]
+
+        if len(set(record_names)) != 1:
+            raise ValueError(f"FASTX record names do not all match, found: {record_names}")
```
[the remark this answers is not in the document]
msto: https://github.com/fulcrumgenomics/fgpyo/issues/281

fgpyo/fastx/__init__.py
msto: Done!

fgpyo/util/logging.py
msto: Done!

fgpyo/fastx/__init__.py
```diff
@@ -63,22 +65,37 @@ def _name_minus_ordinal(name: str) -> str:
     def __next__(self) -> Tuple[FastxRecord, ...]:
         """Return the next set of FASTX records from the zipped FASTX files."""
         records = tuple(next(handle, None) for handle in self._fastx)
+
         if all(record is None for record in records):
             raise StopIteration
-        elif any(record is None for record in records):
-            sequence_name: str = [record.name for record in records if record is not None][0]
+
```
nh13: **nitpick:** 
```suggestion
```

fgpyo/fastx/__init__.py
```diff
@@ -63,22 +65,37 @@ def _name_minus_ordinal(name: str) -> str:
...
+        elif not all_not_none(records):
+            non_none_names: List[str | None] = [
+                record.name for record in records if record is not None
+            ]
+            assert all_not_none(non_none_names)  # type narrowing
+            # We know there is at least one non-None record because the previous conditional
+            # covers the case where all records are None, so it is safe to index into the first
+            # element of non_none_names.
+            sequence_name: str = non_none_names[0]
+
```
nh13: **nitpick:** 
```suggestion
```

fgpyo/fastx/__init__.py
```diff
@@ -63,22 +65,37 @@ def _name_minus_ordinal(name: str) -> str:
...
-            return records
+
+        names_with_ordinals: List[str | None] = [record.name for record in records]
+
+        assert all_not_none(names_with_ordinals)  # type narrowing
+        record_names: List[str] = [self._name_minus_ordinal(name) for name in names_with_ordinals]
+
+        if len(set(record_names)) != 1:
+            raise ValueError(f"FASTX record names do not all match, found: {record_names}")
+
```
nh13: **nitpick:** 
```suggestion
        assert all_not_none(names_with_ordinals)  # type narrowing
        record_names: List[str] = [self._name_minus_ordinal(name) for name in names_with_ordinals]
        if len(set(record_names)) != 1:
            raise ValueError(f"FASTX record names do not all match, found: {record_names}")

```

fgpyo/util/types.py
```diff
@@ -131,7 +138,7 @@ def _make_union_parser_worker(
     union: Type[UnionType],
     parsers: Iterable[Callable[[str], UnionType]],
     value: str,
-) -> UnionType:
+) -> Optional[UnionType]:
```
nh13, line 141: **question:** 
We could use `UnionType | None` given we support only Python 3.10+; shall we do this now, or in a separate PR (if so, lets create an issue)

fgpyo/read_structure.py
```diff
@@ -207,7 +210,7 @@ class ReadStructure(Iterable[ReadSegment]):
     @property
     def _min_length(self) -> int:
         """The minimum length read that this read structure can process"""
-        return sum(segment.length for segment in self.segments if segment.has_fixed_length)
+        return sum(segment.fixed_length for segment in self.segments if segment.has_fixed_length)
```
nh13, line 213: **question (non-blocking):** 
(1) why was `length` changed to `fixed_length`?  Probably "safer", but is this only for type narrowing (since `length` can be `None`)?

fgpyo/read_structure.py
```diff
@@ -292,7 +295,8 @@ def from_segments(
             segs = []
             for seg in segments:
                 seg = attr.evolve(seg, offset=off)
-                off += seg.length if seg.has_fixed_length else 0
+                off += seg.fixed_length if seg.has_fixed_length else 0
```
nh13, line 298: **nitpick:** ditto

tests/fgpyo/sam/test_clipping.py
```diff
@@ -8,7 +8,7 @@
 from fgpyo.sam.builder import SamBuilder
 
 
-def r(start: int | None, cigar: str | None, strand: str | None = "+") -> AlignedSegment:
+def r(start: int | None, cigar: str | None, strand: str = "+") -> AlignedSegment:
```
nh13, line 11: **question (non-blocking):** the type got changed!

fgpyo/read_structure.py
msto: Yes. `length` is `int | None`, which yields a type error inside the `sum()`. Since the comprehension is already filtering the segments to those with a `fixed_length`, we can just use that property instead.

tests/fgpyo/sam/test_clipping.py
msto: Yes! This is a helper in the test suite for constructing reads through `SamBuilder` and not part of the public API. `SamBuilder.add_pair`'s `strand1` parameter isn't Optional, and no caller within the test suite actually passes `None` for strand.

fgpyo/util/types.py
msto: Let's open a separate PR to migrate from `Optional` and `Union` to PEP 604 syntax globally.

I've opened #286 to track.

PR #262, opened 2025-11-04
Speed up reverse_complement by using str.translate

TedBrookings:
This is the same strategy as used by `biopython`, slightly re-engineered to meet `fgpyo`'s tests (e.g. to fail with `KeyError` if a bad nucleotide is passed). Testing with `%%timeit` shows this to be 10x faster for short sequences (around 100 bp), and 30x faster with long sequences (around 100K bp).

I made the _COMPLEMENTS_TABLE a `tuple` to keep it as `const` as possible.

### DIFF
--- a/fgpyo/sequence.py
+++ b/fgpyo/sequence.py
@@ -90,8 +90,10 @@ def reverse_complement(bases: str) -> str:
     """
     rev_comp = bases.translate(_COMPLEMENTS_TABLE)[::-1]
     if len(rev_comp) != len(bases):
-        # There were invalid characters that weren't translated. Find one and raise KeyError.
-        raise KeyError(next(base for base in bases if base not in _COMPLEMENTS))
+        # There were invalid characters that weren't translated.
+        # Raise KeyError with all the invalid bases.
+        bad_bases = "".join({base for base in bases if base not in _COMPLEMENTS})
+        raise KeyError(f"Invalid bases found: {bad_bases}")
     return rev_comp
 
 


fgpyo/sequence.py
```diff
@@ -56,6 +57,12 @@
 }
 
 
+_COMPLEMENTS_TABLE: tuple[Optional[int], ...] = tuple(
+    ord(_COMPLEMENTS.get(chr(key_idx))) if chr(key_idx) in _COMPLEMENTS else None
```
nh13: [suggestion(if-minor)](https://pullpo.io/cc?l=suggestion&d=if-minor) 
it might be useful to just use `str.maketrans` directly so that it is clearer that this is meant to be used with `str.translate`.  Then the `else None` is better understood from the `str` docs

nh13 (approved):
Really nice improvement!

fgpyo/sequence.py
```diff
@@ -56,6 +57,12 @@
 }
 
 
+_COMPLEMENTS_TABLE: tuple[Optional[int], ...] = tuple(
```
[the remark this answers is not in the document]
TedBrookings: Added. I added a few comments as well.

fgpyo/sequence.py
```diff
@@ -73,7 +88,11 @@ def reverse_complement(bases: str) -> str:
     Returns:
         the reverse complement of the provided base string
     """
-    return "".join([_COMPLEMENTS[b] for b in bases[::-1]])
+    rev_comp = bases.translate(_COMPLEMENTS_TABLE)[::-1]
+    if len(rev_comp) != len(bases):
+        # There were invalid characters that weren't translated. Find one and raise KeyError.
+        raise KeyError(next(base for base in bases if base not in _COMPLEMENTS))
```
clintval: It would be nice to list out all the failed bases, not just the first one found.

Something like:

```suggestion
        raise ValueError("Invalid bases found: " + ", ".join(sorted(set(base for base in bases if base not in _COMPLEMENTS))))
```

But I do understand you tried to keep backwards compatibility with the `KeyError`.

fgpyo/sequence.py
```diff
@@ -56,6 +58,19 @@
...
+    str.maketrans("".join(_COMPLEMENTS.keys()), "".join(_COMPLEMENTS.values()), _INVALID_BASES)
+)
+"""
+This table allows faster reverse-complement than directly using _COMPLEMENTS iterating over bases.
+Characters from _INVALID_BASES will be mapped to None and omitted from translation, which can be
+detected as an error by noting decreased string length.
+"""
+
+
 def complement(base: str) -> str:
```
clintval: Odd that we have a `complement()` that only takes 1bp sequences.

And a `reverse_complement()` that takes n-length sequences.

clintval (approved):
Good to squash merge! Or you could take the opportunity to upgrade the error message to be more useful to the user.

fgpyo/sequence.py
TedBrookings: I kept `KeyError` for backward compatibility, but gave a better description. Basically yours but I used a `set` instead of a `list`, so that you get all the _distinct_ errors but not potentially 1 Mbp of "there's no Z in the genome".

PR #264, opened 2025-11-06
Cover unicode corner case

TedBrookings:
In PR #262 (after the PR was merged) the rabbit AI flagged a potential bug if _unicode_ sequences were passed to `reverse_complement`. The suggested fix was bad but the bug was real. This PR covers this corner case by altering the strategy of what is passed to `str.translate`. We create a custom class derived from `Mapping` that throws `ValueError` instead of `KeyError` when a bad input key is passed. This bypasses `str.translate`'s default behavior of passing unrecognized keys untranslated.
Commit logic:
1.  Fix the bug by creating the ComplementsTable class.
2. Add a test that unicode errors are caught.

### DIFF
--- a/fgpyo/sequence.py
+++ b/fgpyo/sequence.py
@@ -9,7 +9,6 @@ If you are performing many distance calculations, using a C based method is pref
 ex. https://pypi.org/project/Distance/
 """
 
-from collections.abc import Iterator
 from collections.abc import Mapping
 from typing import Dict
 from typing import List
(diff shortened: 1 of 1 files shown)


fgpyo/sequence.py
```diff
@@ -58,17 +57,31 @@
...
-# Use str.maketrans to create a table matching ascii codes of bases to ascii codes of complements.
-_COMPLEMENTS_TABLE: MappingProxyType[int, Optional[int]] = MappingProxyType(
-    str.maketrans("".join(_COMPLEMENTS.keys()), "".join(_COMPLEMENTS.values()), _INVALID_BASES)
-)
-"""
-This table allows faster reverse-complement than directly using _COMPLEMENTS iterating over bases.
-Characters from _INVALID_BASES will be mapped to None and omitted from translation, which can be
-detected as an error by noting decreased string length.
-"""
+class ComplementsTable:
```
clintval: This can be protected and hidden from API:


```suggestion
class _ComplementsTable:
```

TedBrookings:
> @TedBrookings feel free to open a new PR, I'd like to close this one given my comment about wanting to go in a different direction.

Yeah, this is only a thing because I didn't consider unicode inputs a possibility and the bot pointed out they are possible. How about just adding a check for `bases.isascii()` to the front of `reverse_complement`? It's built-in and appears to be very fast (16 ns for a 100K sequence).

PR #270, opened 2025-12-17
feat: make pysam an optional dependency                                                                                                            

nh13:
Makes pysam an optional dependency so users who only need pure Python utilities can avoid the heavyweight pysam install.                          

- `pip install fgpyo` — core modules only (sequence, read_structure, collections, io, util)
- `pip install fgpyo[pysam]` — full functionality (sam, vcf, fastx, fasta)

Modules that require pysam will raise a clear error message:

ImportError: Missing optional dependency 'pysam'. The sam, vcf, fastx, and fasta modules require pysam. Install with: `pip install fgpyo[pysam]`

Packaging notes:*    
- PyPI: no changes needed
- Bioconda: recipe should update to `pip install .[pysam]` so users get full functionality by default

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,16 +34,15 @@ dependencies = [
     "attrs>=19.3.0",
     "typing_extensions>=4.12.2 ; python_version < '3.13'",
     "strenum>=0.4.15,<0.5",
+    "zlib-ng>=0.5.1",
 ]
 
 [project.optional-dependencies]
 pysam = [
     "pysam>=0.22.1",
-    "zlib-ng>=0.5.1",
 ]
 all = [
     "pysam>=0.22.1",
-    "zlib-ng>=0.5.1",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -243,16 +243,15 @@ dependencies = [
     { name = "attrs" },
     { name = "strenum" },
     { name = "typing-extensions", marker = "python_full_version < '3.13'" },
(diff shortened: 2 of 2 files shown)


clintval:
pysam is a lot easier to install these days and is also load-bearing for some important features of fgpyo. I'm 50/50 on if this is still a good idea.

msto:
Hey! Noticed this covers the same ground as #238. Were there things about the approach in #238 you wanted to go a different direction on? I'd love to understand your thinking here.

Also curious whether it makes sense to consolidate efforts on one PR so we're not duplicating work. Happy to fold anything useful from #238 into this one, or vice versa, just let me know how you'd like to handle it.

msto:
Going to close #238 in favor of this approach, pending Clint's question on whether this is still worth doing. If it does move forward, a few things I picked up while working through it on my end:

- The `Tests-without-pysam` job looks like `uv run` will re-install pysam on the subsequent calls. I hit this too, `uv run` re-syncs with default groups unless you pass `--no-sync`.
- The `requires_pysam` marker in conftest is a nice pattern. The pysam-dependent test files still import pysam at the top level though, so they'd error rather than skip without pysam. `pytest.importorskip("pysam")` at the top of those files might be a lighter way to get the skip behavior.
- For dev ergonomics, adding a pysam dependency group to `[tool.uv]`'s default-groups means `uv sync` still installs pysam locally without needing `--extra pysam` each time.

Happy to open small PRs against this branch with any of these if that's helpful.

PR #273, opened 2026-02-13
feat: implementation of cigar truncation methods ported from fgbio

ameynert:
Ported `truncate`, `truncateToQueryLength`, and `truncateToTargetLength` and related unit tests from [fgbio](https://github.com/fulcrumgenomics/fgbio/blob/7ee60c361f06c4a082a365f707608238383a2deb/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala#L249)

### DIFF
--- a/tests/fgpyo/sam/test_cigar.py
+++ b/tests/fgpyo/sam/test_cigar.py
@@ -277,3 +277,15 @@ def test_truncate_to_target_length_from_right(
     result = cigar.reversed().truncate_to_target_length(length).reversed()
     assert str(result) == expected
     assert result.length_on_target() == min(length, cigar.length_on_target())
+
+
+def test_truncate_to_query_length_negative_raises() -> None:
+    """truncate_to_query_length() should raise ValueError for negative length."""
+    with pytest.raises(ValueError, match="length must be >= 0"):
+        Cigar.from_cigarstring("10M").truncate_to_query_length(-1)
+
+
+def test_truncate_to_target_length_negative_raises() -> None:
+    """truncate_to_target_length() should raise ValueError for negative length."""
+    with pytest.raises(ValueError, match="length must be >= 0"):
+        Cigar.from_cigarstring("10M").truncate_to_target_length(-1)


tests/fgpyo/sam/test_cigar.py
```diff
@@ -80,3 +80,64 @@ def test_query_alignment_offsets_reversed(
...
+        # Actual truncation to length 50
+        ("60M", 50, "50M"),
+        ("10H50M", 50, "10H50M"),  # Hard clips preserved
+        ("25M10I25M", 50, "25M10I15M"),  # Insertions consume query
+        ("25M10D25M", 50, "25M10D25M"),  # Deletions don't consume query
+        ("50M10S", 50, "50M"),  # Trailing soft clips removed
+        # Additional edge cases
+        ("10M", 0, "*"),  # Truncate to zero
+        ("*", 5, "*"),  # Empty CIGAR
+    ],
```
nh13: The fgbio-ported cases are a good start. Consider expanding coverage with:

- **length=0 + leading non-consuming elements** — this catches the `_truncate` bug above:
  `("5H10M", 0, "*")`, `("10S10M", 0, "*")` (for target truncation)
- **Non-consuming ops interspersed** — `"10M5D10I5N10M"` at various lengths, verifying D/N pass through for query and I/S pass through for target
- **Element splitting** — `("100M", 1, "1M")`, `("100M", 50, "50M")`, `("100M", 99, "99M")`
- **All operator types** — `=`, `X`, `P` in addition to M/I/D/N/S/H
- **Reverse-compose truncation from the right** — `reversed().truncate_to_query_length(n).reversed()`
- **Length invariant sweep** — for each cigar, assert `truncated.length_on_query() == min(length, original)` for all lengths 0..N

fgbio's tests for reference:
- [Truncation + coalesce tests (lines 51-75)](https://github.com/fulcrumgenomics/fgbio/blob/0c75163/src/test/scala/com/fulcrumgenomics/alignment/AlignmentTest.scala#L51-L75)
- [Implementation (lines 192-293)](https://github.com/fulcrumgenomics/fgbio/blob/0c75163/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala#L192-L293)

Also consider adding `coalesce()` ([fgbio's `Cigar.coalesce`](https://github.com/fulcrumgenomics/fgbio/blob/0c75163/src/main/scala/com/fulcrumgenomics/alignment/Alignment.scala#L277-L293)) — merges adjacent same-operator elements, useful after truncation or other cigar manipulation.

ameynert:
@nh13 Should the truncation functions throw an error if the requested truncation length is longer than what is available for either query or target?

tests/fgpyo/sam/test_cigar.py
ameynert: #274 opened for porting `coalesce` - with the additional tests this PR is long enough!

nh13 (approved):
Looks great — all comments addressed, tests are thorough, and the countdown pattern is clean. To answer your question: no, truncation shouldn't throw when the requested length exceeds the cigar's length. fgbio doesn't either — its `while (pos <= len && iter.hasNext)` just returns the full cigar unchanged. It's a "truncate to *at most* N" operation, like Python's `string[:n]` which silently returns the whole string when `n > len(string)`. Your existing `("75M", 100, "75M")` cases already cover this.

One small addition below: a test for the negative-length `ValueError` path you added.

PR #287, opened 2026-04-16
refactor: use ReadSegment.length over .fixed_length in hot paths

msto:
## Summary

- Reverts two callsites in `fgpyo/read_structure.py` from `segment.fixed_length` back to `segment.length` with a narrow `# type: ignore`.
- `fixed_length` re-runs `has_fixed_length` and re-asserts non-None; at callsites already guarded by `has_fixed_length`, this is redundant work. The previous form (pre-#259) avoided that.
- Addresses Nils' post-merge feedback on #259.

## Test plan

- [x] `uv run poe fix-and-check-all` passes (format, lint, mypy, pytest, doctest)
- [x] `tests/fgpyo/test_read_structure.py` — 60 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fgpyo/read_structure.py
+++ b/fgpyo/read_structure.py
@@ -224,7 +224,7 @@ class ReadStructure(Iterable[ReadSegment]):
     @property
     def _min_length(self) -> int:
         """The minimum length read that this read structure can process."""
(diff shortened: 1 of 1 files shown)


nh13 (approved):
Thanks for doing this!

PR #290, opened 2026-04-23
fix: accept file handles and close VariantFile on exception

### LINKED ISSUE #288: `fgpyo.vcf.reader` and `writer` reject real file handles and leak on exception
## Summary

Two correctness bugs in `fgpyo/vcf/__init__.py`:

1. **`isinstance(path, typing.TextIO)` never matches a real file handle.** `typing.TextIO` is a generic alias intended for annotations, not runtime checks. Real file objects inherit from `io.TextIOBase`, not `typing.TextIO`, so the documented "open file handle" input to `reader()` and `writer()` is unreachable.

2. **`yield` is not wrapped in `try/finally`.** Both `reader` and `writer` are `@contextmanager`s. If the caller's `with` block raises, `close()` is skipped and the underlying `pysam.VariantFile` handle leaks.

Neither bug currently has a test.

## Reproduction

```python
import io
import typing
from pathlib import Path

# Bug 1: isinstance check is broken
f = open("/dev/null", "r")
print(isinstance(f, typing.TextIO))   # False
print(isinstance(f, io.IOBase))       # True

# Bug 2: close() skipped on exception
from fgpyo.vcf import reader
try:
    with reader(Path("some.vcf.gz")) as r:
        raise RuntimeError("oops")
except RuntimeError:
    pass
# underlying VariantFile was never closed
```

Passing an `open("some.vcf", "r")` file handle to `fgpyo.vcf.reader()` raises `TypeError: Cannot open '<class '_io.TextIOWrapper'>' for VCF reading.` — contrary to the docstring which advertises "path to a VCF, or an open file handle."

## Proposed change

In `fgpyo/vcf/__init__.py`:

1. Replace `isinstance(path, (str, Path, TextIO))` with `isinstance(path, (str, Path, io.IOBase))` (or mirror the `_IOClasses` tuple pattern already used in `fgpyo/sam/__init__.py:236`). Drop the `TextIO` import.
2. Wrap `yield` in `try/finally` so `_reader.close()` / `_writer.close()` always runs:

```python
@contextmanager
def reader(path: VcfPath) -> Generator[VcfReader, None, None]:
    if not isinstance(path, (str, Path, io.IOBase)):
        raise TypeError(f"Cannot open '{type(path)}' for VCF reading.")
    with fgpyo.io.suppress_stderr():
        _reader = VariantFile(path, mode="r")  # type: ignore[arg-type]
    try:
        yield _reader
    finally:
        _reader.close()
```

Apply the same `try/finally` shape to `writer()`.

## Acceptance criteria

- [ ] Passing a real file handle to `fgpyo.vcf.reader` / `writer` no longer raises `TypeError`.
- [ ] New test: raising inside `with fgpyo.vcf.reader(...) as r:` closes the underlying `VariantFile` (assert `r.is_open is False` in the `except` branch, or use a file-descriptor count check).
- [ ] New test: same for `writer`.
- [ ] `VcfPath` alias no longer references `typing.TextIO`.

msto:
## Summary

Fixes #288

`fgpyo.vcf.reader` / `writer` had two correctness bugs. The runtime `isinstance(path, typing.TextIO)` check never matched real file objects — `typing.TextIO` is an annotation alias, so the documented "open file handle" input was unreachable and raised `TypeError`. And `yield` was not wrapped in `try/finally`, so if the caller's `with` block raised, the underlying `pysam.VariantFile` leaked.

Swap `typing.TextIO` for `io.IOBase` in the isinstance check and the `VcfPath` alias, and wrap `yield` in `try/finally` in both `reader()` and `writer()`.

## Test plan

- New `tests/fgpyo/vcf/test_io.py` covers: raising inside `with reader(...)` / `with writer(...)` closes the `VariantFile`; passing a real `open(path, "rb")` handle to `reader()` now yields records; passing a non-path non-handle still raises `TypeError`.
- Full `tests/fgpyo/vcf/` suite, mypy, and ruff all clean.

### DIFF
--- a/fgpyo/vcf/__init__.py
+++ b/fgpyo/vcf/__init__.py
@@ -105,6 +105,8 @@ def writer(path: VcfPath, header: VariantHeader) -> Generator[VcfWriter, None, N
         header: the source for the output VCF header. If you are modifying a VCF file that you are
                 reading from, you can pass reader.header
     """
+    if not isinstance(path, (str, Path, io.IOBase)):
+        raise TypeError(f"Cannot open '{type(path)}' for VCF writing.")
     # Convert Path to str such that pysam will autodetect to write as a gzipped file if provided
     # with a .vcf.gz suffix.
     if isinstance(path, Path):
--- a/fgpyo/vcf/__init__.py
+++ b/fgpyo/vcf/__init__.py
@@ -58,10 +58,10 @@ in coordinate sorted order via the
 
 """
 
+import io
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Generator
-from typing import TextIO
 
 from pysam import VariantFile
 from pysam import VariantFile as VcfReader
@@ -71,7 +71,7 @@ from pysam import VariantHeader
 import fgpyo.io
 
 """The valid base classes for opening a VCF file."""
-VcfPath = Path | str | TextIO
+VcfPath = Path | str | io.IOBase
 
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


fgpyo/vcf/__init__.py
```diff
@@ -109,5 +110,7 @@ def writer(path: VcfPath, header: VariantHeader) -> Generator[VcfWriter, None, N
     if isinstance(path, Path):
         path = str(path)
     _writer = VariantFile(path, header=header, mode="w")
-    yield _writer
```
clintval, line 112: Given we type `path` as `VcfPath`, what happens to `io.IoBase` in this code block? Looks like we convert `Path` to `str` for format auto-detection. Does the same happen for `io.IoBase` types too? Should we guard for other types like we do in `def reader` with a `not isinstance: raise ...`?

clintval (approved):
Definitely an improvement!

nh13 (approved):

fgpyo/vcf/__init__.py
msto: Yep - also opened #297 to expose the write mode

PR #291, opened 2026-04-23
fix: `copy_umi_from_read_name` delimiter for multi-UMI read names

### LINKED ISSUE #289: `copy_umi_from_read_name` passes wrong delimiter, breaking multi-UMI read names
## Summary

`fgpyo.platform.illumina.copy_umi_from_read_name` passes the wrong delimiter when calling `extract_umis_from_read_name`, which prevents it from handling read names containing multiple UMIs (e.g. `inst:1:2:3:4:5:6:AAAA+TTTT`).

At `fgpyo/platform/illumina.py:111–115`:

```python
umi = extract_umis_from_read_name(
    read_name=rec.query_name,
    strict=strict,
    umi_delimiter=_ILLUMINA_READ_NAME_DELIMITER,   # ← should be _ILLUMINA_UMI_DELIMITER
)
```

`_ILLUMINA_READ_NAME_DELIMITER` is `":"`; `_ILLUMINA_UMI_DELIMITER` is `"+"`. Inside `extract_umis_from_read_name`, the read-name delimiter has already been consumed by `raw_umi = read_name.split(read_name_delimiter)[-1]` before the UMI delimiter is applied. Splitting the raw UMI by `:` a second time always produces a single-element list, so:

- The combined `AAAA+TTTT` string is passed whole to `_is_valid_umi`.
- `_is_valid_umi("AAAA+TTTT")` returns `False` (contains `+`, which is not in `_VALID_UMI_CHARACTERS`).
- `copy_umi_from_read_name` returns `False` (or raises in strict mode).

Single-UMI read names happen to work because the inner split is a no-op in that case — which is why this has not been caught.

## Reproduction

```python
import pysam
from fgpyo.platform.illumina import copy_umi_from_read_name

rec = pysam.AlignedSegment()
rec.query_name = "inst:1:2:3:4:5:6:AAAA+TTTT"
# Expected: True, RX tag set to "AAAA-TTTT"
# Actual:   False, RX tag unset
assert copy_umi_from_read_name(rec) is True  # fails
```

## Proposed change

Change the call site at `fgpyo/platform/illumina.py:114` to pass the UMI delimiter:

```python
umi = extract_umis_from_read_name(
    read_name=rec.query_name,
    strict=strict,
    umi_delimiter=_ILLUMINA_UMI_DELIMITER,
)
```

## Acceptance criteria

- [ ] `copy_umi_from_read_name` sets the `RX` tag to `AAAA-TTTT` (SAM-standard hyphen separator) when given a read name ending in `:AAAA+TTTT`.
- [ ] Regression test added to `tests/fgpyo/platform/test_illumina.py` covering: single UMI (unchanged behavior), multi-UMI (new passing case), invalid multi-UMI (`:AAAA+NNNZ`) in both strict and non-strict modes.
- [ ] `extract_umis_from_read_name` itself is unchanged — the bug is entirely in the caller.

msto:
## Summary

Fixes #289.

`copy_umi_from_read_name` was passing `:` (the read-name delimiter) as the UMI delimiter to `extract_umis_from_read_name`. By that point the raw UMI segment has already been split off, so splitting `"AAAA+TTTT"` by `:` yielded a single-element list, the whole `AAAA+TTTT` string was checked against the allowed UMI characters, and the `+` made it fail validation. Single-UMI read names happened to work because the inner split was a no-op, which is why this went unnoticed.

The fix drops the incorrect kwarg so the function's default (`_ILLUMINA_UMI_DELIMITER = "+"`) is used. `extract_umis_from_read_name` itself is unchanged.

Per review, the signature of `copy_umi_from_read_name` is also extended to expose `read_name_delimiter` and `umi_delimiter` (forwarded to `extract_umis_from_read_name`), so non-Illumina callers can use the higher-level helper without falling back to the underlying extractor. **Defaults match the existing Illumina constants, so all existing callers are unaffected.** The `remove_umi` branch now uses the passed-in `read_name_delimiter` rather than the hardcoded constant.

## Test plan

- [x] Multi-UMI happy path (`AAAA+TTTT` → `AAAA-TTTT`) now passes across all `(remove_umi, strict)` combinations.
- [x] Invalid multi-UMI (`AAAA+NNNZ`) raises in strict mode and returns `False` in non-strict mode, without setting `RX`.
- [x] Custom `read_name_delimiter` and `umi_delimiter` are exercised end-to-end, including the `remove_umi` path.
- [x] Incidentally fixes a pre-existing parametrize bug in `test_copy_valid_umi_from_read_name` that ran the same `(remove_umi, strict)` case twice.
- [ ] CI passes

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/fgpyo/platform/illumina.py
+++ b/fgpyo/platform/illumina.py
@@ -87,17 +87,28 @@ def extract_umis_from_read_name(
 
 
 def copy_umi_from_read_name(
-    rec: AlignedSegment, strict: bool = False, remove_umi: bool = False
+    rec: AlignedSegment,
+    strict: bool = False,
+    remove_umi: bool = False,
+    read_name_delimiter: str = _ILLUMINA_READ_NAME_DELIMITER,
+    umi_delimiter: str = _ILLUMINA_UMI_DELIMITER,
 ) -> bool:
     """
     Copy a UMI from an alignment's read name to its `RX` SAM tag.
 
     The UMI will not be copied to RX tag if it is invalid.
 
+    `strict`, `read_name_delimiter`, and `umi_delimiter` are forwarded to
+    [`extract_umis_from_read_name`][fgpyo.platform.illumina.extract_umis_from_read_name] — see
+    that function for their semantics.
+
     Args:
         rec: The alignment record to update.
-        strict: If `True` and UMI invalid, will throw an exception
+        strict: If `True` and UMI invalid, will throw an exception.
         remove_umi: If `True`, the UMI will be removed from the read name after copying.
+        read_name_delimiter: The delimiter separating the components of the read name.
+            Also used to strip the UMI segment when `remove_umi` is `True`.
+        umi_delimiter: The delimiter separating multiple UMIs.
 
     Returns:
         `True` if the UMI was successfully extracted, False if otherwise.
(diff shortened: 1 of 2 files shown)


fgpyo/platform/illumina.py
```diff
@@ -111,7 +111,6 @@ def copy_umi_from_read_name(
     umi = extract_umis_from_read_name(
         read_name=rec.query_name,
         strict=strict,
-        umi_delimiter=_ILLUMINA_READ_NAME_DELIMITER,
```
msto, line 114: @nh13 

Claude went with the simplest fix of calling `extract_umis_from_read_name()` with default values within `copy_umi_from_read_name()`. 

Should we instead add the various parameters from the former to the latter? `strict` already appears, but the latter doesn't support customizing the read name or UMI delimiters.

nh13 (approved):

Issue #286, opened 2026-04-16
chore: migrate type annotations to PEP 604 syntax (`T | None`, `A | B`)


msto:
## Summary

Now that fgpyo supports Python 3.10+ (see #261), we can migrate type annotations
from `typing.Optional[T]` and `typing.Union[A, B]` to PEP 604 syntax
(`T | None` and `A | B`).

This is purely a style/modernization change — no runtime behavior changes.

## Motivation

- PEP 604 syntax is the recommended style on Python 3.10+ and is easier to read.
- Reduces noise in imports (`from typing import Optional, Union` can often be dropped).
- Aligns annotations with the style already used in newer code (e.g. `fgpyo/fastx/__init__.py`
  uses `str | None`).

Raised in [review feedback on PR #259](https://github.com/fulcrumgenomics/fgpyo/pull/259#discussion_r3095164833).

## Scope

Replace annotation-site uses of:
- `Optional[T]` → `T | None`
- `Union[A, B]` → `A | B`
- `Union[A, B, None]` / `Optional[Union[A, B]]` → `A | B | None`

Current occurrences (~29 across ~7 files):

**`fgpyo/`** (14 total)
- `fgpyo/util/types.py` (11 — see note below)
- `fgpyo/sam/__init__.py` (2)
- `fgpyo/sequence.py` (1)

**`tests/`** (15 total)
- `tests/fgpyo/util/test_metric.py` (7)
- `tests/fgpyo/util/test_types.py` (4)
- `tests/fgpyo/util/test_inspect.py` (3)
- `tests/fgpyo/sam/test_sam.py` (1)

## Out of scope / keep as-is

- **Docstring and comment references** in `fgpyo/util/types.py` that document
  the three accepted syntaxes (`Optional[T]`, `Union[T, None]`, `T | None`) —
  those are user-facing documentation of valid input forms.
- **Runtime uses** of `typing.Union` / `types.UnionType` as objects (e.g.
  `typing.get_origin(...) is Union`) — only annotation-site uses should change.

## Acceptance criteria

- [ ] All annotation-site `Optional[...]` / `Union[...]` in `fgpyo/` and `tests/` migrated to PEP 604 syntax
- [ ] Unused `from typing import Optional, Union` imports removed
- [ ] `uv run poe fix-and-check-all` passes
- [ ] Docstring/comment references to the legacy syntaxes preserved where they document accepted input forms

PR #294, opened 2026-04-23
fix: allow SamBuilder to produce records with no sequence or qualities

### LINKED ISSUE #211: `SamBuilder.add_single()` cannot create a record without sequence or base qualities
## Summary

When attempting to construct a test record with no query sequence or qualities, I found that `SamBuilder.add_single()` constructs a record with the default base/quality values instead of a record with `None` in these attributes.

```py
In [1]: from fgpyo.sam.builder import SamBuilder

In [2]: builder = SamBuilder()

In [3]: record = builder.add_single(bases=None, quals=None)

In [4]: record.query_sequence
Out[4]: 'AGGGCGACCTTCGATTCGGATGTGACATTTCATTACATTACGCTCAGGACTGCGAACGAAAGATTAAGAATGCTTAACCCGGTACCTAACCCATCTGATT'

In [5]: record.query_qualities
Out[5]: array('B', [30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30])
```

## Suggested solution

I think this originates from using `None` as a sentinel value for "use the defaults" in this method. 

Instead, we could:

- Change the default value for these parameters to the actual default query sequence and qualities, rather than `None`
- When `None` is specified as a value for these parameters, return a record where `query_sequence` and `query_qualities` are None

msto:
## Summary

Fixes #211.

`SamBuilder.add_single(bases=None, quals=None)` could not produce a record with `query_sequence`/`query_qualities` set to `None` — it synthesized random bases and default qualities instead. `None` was overloaded as both "caller omitted this argument" and "caller wants no value," and the fill-in logic collapsed both to the synthesized-default branch.

Introduces a module-level `_UNSET` sentinel in `fgpyo/sam/builder.py` so the two intents can be distinguished. **Omitting `bases`/`quals` behaves exactly as before** — defaults are still synthesized, so all existing callers are unaffected. Only passing `None` explicitly changes: it now produces a record with the attribute unset. Since SAM records can't carry qualities without a sequence, `bases=None` implicitly clears qualities when the caller didn't pass `quals`, and passing concrete `quals` with `bases=None` raises `ValueError`.

| Call | Before | After |
|---|---|---|
| `add_single()` | random bases + default quals | unchanged |
| `add_single(bases="ACGT")` | uses "ACGT" | unchanged |
| `add_single(bases=None)` | random bases (bug) | seq + quals both `None` |
| `add_single(quals=None)` | default quals (bug) | quals `None`, sequence synthesized |
| `add_single(bases=None, quals=None)` | defaults (bug) | both `None` |

## Test plan

- New tests cover each combination of explicit `None` on `add_single` and per-mate on `add_pair`, on both unmapped and mapped reads.
- BAM round-trip verifies sequence-less records survive serialization.
- Full test suite passes; mypy and ruff are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -293,9 +293,11 @@ class SamBuilder:
         if quals is None:
             return None
         if quals is not _UNSET:
+            # Qualities without a sequence is explicitly disallowed by the SAM spec.
+            if bases is None:
+                raise ValueError("Cannot provide qualities when bases is None.")
             return array("B", quals)
-        # Argument was omitted: synthesize unless the sequence was explicitly cleared, since
-        # qualities without a sequence is not a valid SAM record.
+        # Argument was omitted: synthesize unless the sequence was explicitly cleared.
         if bases is None:
             return None
         return array("B", [self.base_quality] * length)
--- a/fgpyo/sam/builder.py
+++ b/fgpyo/sam/builder.py
@@ -24,6 +24,10 @@ from pysam import AlignmentHeader
 from fgpyo import sam
 from fgpyo.sam import SamOrder
 
+# Sentinel used to distinguish "caller did not supply this argument, synthesize a default"
+# from "caller explicitly passed None, produce a record with no value for this attribute."
+_UNSET: Any = object()
+
 
 class SamBuilder:
     """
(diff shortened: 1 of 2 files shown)


fgpyo/sam/builder.py
```diff
@@ -230,48 +234,72 @@ def _set_length_dependent_fields(
...
         if not rec.is_unmapped:
             rec.cigarstring = cigar if cigar else f"{length}M"
 
+    def _resolve_quals(
+        self,
+        quals: List[int] | None,
+        bases: str | None,
+        length: int,
+    ) -> "array[int] | None":
+        """Resolves the value to assign to `query_qualities` from the caller's inputs."""
```
nh13, line 292: **issue:**  providing `quals` while `bases` is `None` is explicitly disallowed in the SAM spec

tests/fgpyo/sam/test_builder.py
```diff
@@ -271,6 +271,87 @@ def test_add_single() -> None:
...
+
+def test_add_single_bases_none_also_clears_quals() -> None:
+    """Setting `bases=None` implies `quals=None`; qualities without a sequence is invalid SAM."""
+    builder = SamBuilder()
+    r = builder.add_single(bases=None)
+    assert r.query_sequence is None
+    assert r.query_qualities is None
+
+
+def test_add_single_quals_none_preserves_bases() -> None:
```
nh13, line 290: **nitpick:** add the same `quals_none` for the paired end case?

fgpyo/sam/builder.py
```diff
@@ -230,48 +234,72 @@ def _set_length_dependent_fields(
...
+        # Assign sequence first: pysam resets query_qualities when query_sequence is assigned.
+        rec.query_sequence = resolved_bases
+        rec.query_qualities = resolved_quals
+
         if not rec.is_unmapped:
             rec.cigarstring = cigar if cigar else f"{length}M"
 
+    def _resolve_quals(
+        self,
+        quals: List[int] | None,
```
nh13, line 288: **question:** is it time to use `list[int]`?

nh13 (approved):
One spec issue, I think, other than that ship it

fgpyo/sam/builder.py
msto: Yeah! I'll open a similar issue + PR to #286 + #292 - should be quick to fix globally.

PR #295, opened 2026-04-23
fix: recognize PEP 604 union syntax and multi-arg unions as optional

### LINKED ISSUE #286: chore: migrate type annotations to PEP 604 syntax (`T | None`, `A | B`)

### LINKED ISSUE #293: fix: `_attribute_is_optional` does not recognize PEP 604 union syntax
## Summary

`fgpyo/util/inspect.py::_attribute_is_optional` does not recognize PEP 604 union syntax (`T | None`), so `Metric` and `attr_from` classes defined with PEP 604 annotations treat optional fields as required.

## Reproducer

```python
import attr
from fgpyo.util.inspect import _attribute_is_optional

@attr.s(auto_attribs=True, frozen=True)
class Foo:
    x: str | None = None

_attribute_is_optional(attr.fields_dict(Foo)["x"])  # -> False  (expected: True)
```

The same class with `Optional[str]` returns `True`. Downstream, this means `Metric.read` raises `ValueError: fields in file missing from class` for optional fields declared with PEP 604, and `attr_from` fails the "No value given and no default" assertion.

## Root cause

```python
# fgpyo/util/inspect.py:492
def _attribute_is_optional(attribute: FieldType) -> bool:
    return typing.get_origin(attribute.type) is Union and isinstance(
        None, typing.get_args(attribute.type)
    )
```

`Optional[T]` and `Union[T, None]` have `typing.Union` as their origin, but PEP 604 `T | None` has `types.UnionType`. The function only checks the former.

The same bug does **not** exist in `fgpyo/util/types.py::_is_optional`, which correctly checks both (`origin is Union or origin is UnionType`).

## Proposed fix

Mirror the dual-origin check from `_is_optional`:

```python
def _attribute_is_optional(attribute: FieldType) -> bool:
    origin = typing.get_origin(attribute.type)
    return (origin is Union or origin is types.UnionType) and isinstance(
        None, typing.get_args(attribute.type)
    )
```

## Scope

Discovered while working on #286 (PR #292). The PEP 604 migrations of `tests/fgpyo/util/test_metric.py` (7 sites) and `tests/fgpyo/util/test_inspect.py` (3 sites) were deferred from #292 because they exercise this code path — they should be folded into the PR that fixes this.

## Acceptance criteria

- [ ] `_attribute_is_optional` recognizes `T | None` fields as optional
- [ ] Direct unit test coverage for PEP 604 annotations in `test_inspect.py::test_attribute_is_optional` / `test_attribute_has_default`
- [ ] Deferred migrations from #286 applied in `test_metric.py` and `test_inspect.py`
- [ ] `uv run poe check-all` passes

### LINKED ISSUE #296: Permit multi-arg unions containing `None` (e.g. `str | int | None`) as optional
## Summary

`fgpyo/util/types.py::_is_optional` currently requires exactly two type
arguments — it returns `True` for `str | None` but `False` for
`str | int | None`. We should loosen the check so that any union
containing `type(None)` is treated as optional.

## Motivation

After #295, `fgpyo/util/inspect.py::_attribute_is_optional` delegates to
`_is_optional`, so the behavior is now consistent across the two call
sites. Both share the same constraint: a field or type declared as
`Union[A, B, None]` / `A | B | None` is **not** recognized as optional.

This is a user-facing gap:

- `Metric` subclasses with `field: str | int | None` will raise
  `ValueError: fields in file missing from class` when the column is
  absent, even though the annotation plainly permits `None`.
- `_make_union_parser_worker` in `fgpyo/util/types.py` won't short-circuit
  to the `none_parser` for such fields, so the empty string `""` is
  passed through the other parsers instead of resolving to `None`.

There is no semantic reason to special-case 2-arg unions. "Contains
`None` in its args" is the natural definition of optional.

## Proposed change

```python
def _is_optional(dtype: TypeAnnotation) -> bool:
    ...
    origin = typing.get_origin(dtype)
    args = typing.get_args(dtype)
    return (
        origin is not None
        and (origin is Union or origin is UnionType)
        and type(None) in args
    )
```

i.e. drop the `len(args) == 2` constraint.

## Scope

- Update `_is_optional` in `fgpyo/util/types.py`.
- Update parametrize cases in `tests/fgpyo/util/test_types.py` — the
  existing cases `(Union[str, int, None], False)` and `(str | int | None,
  False)` should flip to `True`.
- Add Metric / attr coverage for multi-arg optional fields in
  `tests/fgpyo/util/test_metric.py` and `tests/fgpyo/util/test_inspect.py`.

## Acceptance criteria

- [ ] `_is_optional(str | int | None)` → `True`
- [ ] `_is_optional(Union[str, int, None])` → `True`
- [ ] `Metric` subclasses with multi-arg optional fields round-trip missing
      columns as `None` without raising
- [ ] `_make_union_parser_worker` returns `None` for empty string input
      on multi-arg optional unions
- [ ] `uv run poe check-all` passes

## Context

Surfaced during review of #295 (see
https://github.com/fulcrumgenomics/fgpyo/pull/295#discussion_r3135653578)
after consolidating `_attribute_is_optional` to delegate to `_is_optional`.

msto:
## Summary

Fixes #293. Fixes #296. Completes #286.

`fgpyo/util/inspect.py::_attribute_is_optional` only checked `typing.get_origin(attr.type) is Union`, so `Metric` and `attr_from` classes defined with PEP 604 annotations (`T | None`) treated optional fields as required — `Metric.read` raised `ValueError: fields in file missing from class` and `attr_from` failed the "No value given and no default" assertion. After consolidating the check to delegate to `types._is_optional` (which already handled PEP 604), a second gap surfaced: `_is_optional` required exactly 2 args, so multi-arg unions like `str | int | None` were not recognized as optional either. Both gaps are fixed.

Also picks up the PEP 604 migrations in `tests/fgpyo/util/test_metric.py` (7 sites) and `tests/fgpyo/util/test_inspect.py` (3 sites) that were deferred from #292 because they exercise these code paths.

## Test plan

- [x] `_attribute_is_optional` / `_attribute_has_default` recognize both `Optional[T]` / `Union[T, None]` (legacy) and `T | None` (PEP 604) fields — covered by new `LegacyOptionalName`-based tests.
- [x] `_attribute_is_optional` / `_attribute_has_default` recognize multi-arg optional fields (`str | int | None`, `Union[str, int, None]`) — covered by new `MultiArgOptionalName`-based tests.
- [x] `Metric` round-trip test confirms a multi-arg optional column may be missing from the file and is read as `None`.
- [x] Existing `test_is_optional` parametrize updated: `Union[str, int, None]` and `str | int | None` now expected to be optional.
- [x] Lint, type-check, unit (831), and doctest (17) suite passes.

### DIFF
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -24,7 +24,6 @@ from typing import Type
 from typing import TypeAlias
 from typing import TypeGuard
 from typing import TypeVar
-from typing import Union
 
 import fgpyo.util.types as types
 
@@ -491,10 +490,7 @@ def attr_from(
 
 def _attribute_is_optional(attribute: FieldType) -> bool:
     """Returns True if the attribute is optional, False otherwise."""
-    origin = typing.get_origin(attribute.type)
-    return (origin is Union or origin is python_types.UnionType) and isinstance(
-        None, typing.get_args(attribute.type)
-    )
+    return types._is_optional(attribute.type)
 
 
 def _attribute_has_default(attribute: FieldType) -> bool:
--- a/fgpyo/util/inspect.py
+++ b/fgpyo/util/inspect.py
@@ -491,7 +491,8 @@ def attr_from(
 
 def _attribute_is_optional(attribute: FieldType) -> bool:
     """Returns True if the attribute is optional, False otherwise."""
-    return typing.get_origin(attribute.type) is Union and isinstance(
+    origin = typing.get_origin(attribute.type)
+    return (origin is Union or origin is python_types.UnionType) and isinstance(
         None, typing.get_args(attribute.type)
     )
 
--- a/fgpyo/util/types.py
+++ b/fgpyo/util/types.py
@@ -116,8 +116,7 @@ def _is_optional(dtype: TypeAnnotation) -> bool:
         dtype: A type.
 
     Returns:
-        True if the type is a union type with exactly two elements, one of which is `None`.
-        False otherwise.
+        True if the type is a union type that includes `None` in its arguments. False otherwise.
 
     Raises:
         TypeError: If the input is not a valid `TypeAnnotation` type.
@@ -130,12 +129,7 @@ def _is_optional(dtype: TypeAnnotation) -> bool:
 
     # Optional[T] has `typing.Union` as its origin, but PEP604 syntax (e.g. `int | None`) has
     # `types.UnionType` as its origin.
-    return (
-        origin is not None
-        and (origin is Union or origin is UnionType)
-        and len(args) == 2
-        and type(None) in args
-    )
+    return origin is not None and (origin is Union or origin is UnionType) and type(None) in args
 
 
 def _make_union_parser_worker(
--- a/tests/fgpyo/util/test_inspect.py
+++ b/tests/fgpyo/util/test_inspect.py
@@ -48,6 +48,15 @@ class LegacyOptionalName:
     union_with_none: Union[str, None] = None
 
 
+@attr.s(auto_attribs=True, frozen=True)
+class MultiArgOptionalName:
+    """Test attr class with multi-arg union fields containing None."""
+
+    required: str
+    multi_arg_optional: str | int | None
+    legacy_multi_arg_optional: Union[str, int, None] = None
+
+
 def test_attr_from() -> None:
     name = attr_from(
         cls=Name,
@@ -94,6 +103,20 @@ def test_attribute_has_default_legacy_syntax() -> None:
     assert not _attribute_has_default(fields_dict["required"])
     assert _attribute_has_default(fields_dict["optional_no_default"])
     assert _attribute_has_default(fields_dict["union_with_none"])
+
+
+def test_attribute_is_optional_multi_arg_union() -> None:
+    fields_dict = attr.fields_dict(MultiArgOptionalName)
+    assert not _attribute_is_optional(fields_dict["required"])
+    assert _attribute_is_optional(fields_dict["multi_arg_optional"])
+    assert _attribute_is_optional(fields_dict["legacy_multi_arg_optional"])
+
+
+def test_attribute_has_default_multi_arg_union() -> None:
+    fields_dict = attr.fields_dict(MultiArgOptionalName)
+    assert not _attribute_has_default(fields_dict["required"])
+    assert _attribute_has_default(fields_dict["multi_arg_optional"])
+    assert _attribute_has_default(fields_dict["legacy_multi_arg_optional"])
 
 
 class Foo:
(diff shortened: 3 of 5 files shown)


fgpyo/util/inspect.py
```diff
@@ -491,7 +491,8 @@ def attr_from(
 
 def _attribute_is_optional(attribute: FieldType) -> bool:
     """Returns True if the attribute is optional, False otherwise."""
-    return typing.get_origin(attribute.type) is Union and isinstance(
```
nh13, line 494: **nitpick:** can this just be `types._is_optional(attribute.type)`?

nh13 (approved):
one nit

fgpyo/util/inspect.py
msto: I've done this, but it had a couple ramifications - mind taking a look at the latest two commits on this branch before I merge?

nh13 (approved):
The last commit looks like a nice bonus!

clintval (approved):

PR #299, opened 2026-04-24
feat: add mode kwarg to fgpyo.vcf.writer

### LINKED ISSUE #297: `fgpyo.vcf.writer` hardcodes `mode="w"`, can't produce bgzipped output via file handle
## Summary

`writer()` passes `mode="w"` to `pysam.VariantFile` unconditionally. For `str` / `Path` inputs pysam auto-detects the format from the extension (e.g. `.vcf.gz` → bgzipped). For `io.IOBase` inputs — now reachable after #290 — there is no filename to sniff, so output is always uncompressed VCF regardless of how the caller opened the handle:

```python
# Uncompressed VCF bytes written to a .vcf.gz filename.
with open("out.vcf.gz", "wb") as f, fgpyo.vcf.writer(f, header=h) as w: ...
```

Read-side is fine — pysam sniffs magic bytes off the stream, so `reader()` handles bgzipped handles transparently.

## Proposed change

Add a `mode` keyword argument to `writer()` that is forwarded to `pysam.VariantFile`, defaulting to `"w"` to preserve existing behavior:

```python
def writer(
    path: VcfPath,
    header: VariantHeader,
    mode: str = "w",
) -> Generator[VcfWriter, None, None]:
    ...
    _writer = VariantFile(path, header=header, mode=mode)
```

Callers writing bgzipped output via a handle would pass `mode="wz"`; BCF callers would pass `mode="wb"`.

Alternative (rejected): sniff the handle's `.name` attribute for an extension. Fragile — not all IO objects have a usable `.name`, and it silently breaks when they don't.

## Acceptance criteria

- [ ] `writer()` accepts a `mode` kwarg, forwarded to `pysam.VariantFile`.
- [ ] Existing `str` / `Path` callers unaffected.
- [ ] Test: `open(path, "wb")` handle + `mode="wz"` produces a bgzipped VCF that `reader()` round-trips.

Surfaced during review of #290.

msto:
## Summary

Closes #297

`fgpyo.vcf.writer` passed `mode="w"` to `pysam.VariantFile` unconditionally. For `str` / `Path` inputs pysam auto-detects format from the filename extension, but for `io.IOBase` inputs (reachable since #290) there is no filename to sniff, so output is always uncompressed VCF regardless of how the handle was opened.

Forward an explicit `mode` kwarg to `VariantFile`, defaulting to `"w"`. File-handle callers can now request bgzipped VCF with `mode="wz"` or BCF with `mode="wb"`.

**Existing `str` / `Path` callers are unaffected** — the default preserves the filename-based auto-detection they rely on, and a new regression test locks that in.

Typed as `str` rather than a `Literal[...]` because pysam's own stub omits `"wz"` even though it works at runtime; narrowing would reject valid modes. Suppressed on the `VariantFile` call the same way the reader side already does.

## Test plan

- New tests verify: writing via `open(path, "wb") + mode="wz"` produces a bgzipped file (gzip magic bytes) that round-trips through `reader()`; `writer(path.vcf.gz)` with default mode still produces bgzipped output.
- Existing `tests/fgpyo/vcf/` suite, mypy, and ruff all clean.

### DIFF
--- a/tests/fgpyo/vcf/test_io.py
+++ b/tests/fgpyo/vcf/test_io.py
@@ -48,3 +48,28 @@ def test_writer_rejects_non_path_non_handle() -> None:
     with pytest.raises(TypeError, match="Cannot open"):
         with vcf_writer(123, header=header):  # type: ignore[arg-type]
             pass
+
+
+_BGZF_MAGIC = b"\x1f\x8b"
+
+
+def test_writer_file_handle_bgzipped_round_trip(tmp_path: Path) -> None:
+    builder = VariantBuilder()
+    record = builder.add()
+    out = tmp_path / "out.vcf.gz"
+    with open(out, "wb") as handle:
+        with vcf_writer(handle, header=builder.header, mode="wz") as w:
+            w.write(record)
+    assert out.read_bytes()[:2] == _BGZF_MAGIC
+    with vcf_reader(out) as r:
+        records = list(r)
+    assert len(records) == 1
+
+
+def test_writer_default_mode_autodetects_bgzip(tmp_path: Path) -> None:
+    builder = VariantBuilder()
+    record = builder.add()
+    out = tmp_path / "out.vcf.gz"
+    with vcf_writer(out, header=builder.header) as w:
+        w.write(record)
+    assert out.read_bytes()[:2] == _BGZF_MAGIC
--- a/tests/fgpyo/vcf/test_io.py
+++ b/tests/fgpyo/vcf/test_io.py
@@ -50,7 +50,7 @@ def test_writer_rejects_non_path_non_handle() -> None:
             pass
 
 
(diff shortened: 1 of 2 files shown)


tests/fgpyo/vcf/test_io.py
```diff
@@ -48,3 +48,28 @@ def test_writer_rejects_non_path_non_handle() -> None:
     with pytest.raises(TypeError, match="Cannot open"):
         with vcf_writer(123, header=header):  # type: ignore[arg-type]
             pass
+
+
+_BGZF_MAGIC = b"\x1f\x8b"
```
nh13: **nitpick:** 
type?

nh13 (approved):

clintval (approved):

